PR #25, opened 2024-04-24
Add ci

MauroToscano:
Add CI with build and lint


.github/workflows/build.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          version: v1.54
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Build aggregator
+        run: go build aggregator/cmd/main.go
+      - name: Build operator
+        run: go build operator/cmd/main.go
+      - name: Build task sender
+        run: go build operator/cmd/main.go
```
entropidelic:
```suggestion
        run: go build task_sender/cmd/main.go
```

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,55 @@
...
+          # skip-cache: true
+
+          # Optional: if set to true, then the action won't cache or restore ~/go/pkg.
+          # skip-pkg-cache: true
+
+          # Optional: if set to true, then the action won't cache or restore ~/.cache/go-build.
+          # skip-build-cache: true
+
+          # Optional: The mode to install golangci-lint. It can be 'binary' or 'goinstall'.
+          # install-mode: "goinstall"
```
entropidelic, line 55: are these comments intended to be left just in case?

entropidelic (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+  build:
+    runs-on: ubuntu-latest
+    
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+          cache: false
+      - name: Checkout code
```
NicolasRampoldi, line 20: The first "uses" step checksout, do we need to do it again?

NicolasRampoldi (approved):

.github/workflows/build.yml
MauroToscano: Probably not

PR #27, opened 2024-04-24
refactor: make config more modular

### LINKED ISSUE #17: feat (aggregator): Allow different configuration files or environment variabels
`aggregator-start` make target currently runs:

```bash
go run aggregator/cmd/aggregator.go --config aggregator/config/config.yaml \
		--aligned-layer-deployment contracts/script/output/devnet/alignedlayer_deployment_output.json \
		--ecdsa-private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
```

We should figure out a better way to pass in configuration files, since make targets don't do well with parameters, a good way to do this might be a .env file + a bash script

### LINKED ISSUE #22: refactor: make config more modular
- Remove logger and eth clients creation from config.
- Create logger and eth clients outside config.
- Inject the logger and the eth clients to the config to make the code more modular and testable.

NicolasRampoldi:
- Create Logger.
- Make config.go more modular, created BaseConfig, OperatorConfig and AggregatorConfig.
- Validate all parameters before performing any action.
- Delete `config_test.go` as it is no longer necessary because there is no validate method.
- The cli now has 3 flags:
  - `--base-config-file`: for the base config file (ex: `config-files/config.yaml`).
  - `--aggregator-config-file`: for the specific aggregator config file (ex: `aggregator/config/config.yaml`).
  - `--operator-config-file`: for the specific operator config file (ex: `operator/config/config.yaml`).
- Adapt aggregator code to this new config format.


MauroToscano (approved):

core/config/config.go
```diff
@@ -20,201 +19,358 @@ import (
...
+		Name:     "aggregator-config-file",
+		Required: true,
+		Usage:    "Load aggregator configurations from `FILE`",
+	}
+	OperatorConfigFileFlag = cli.StringFlag{
+		Name:     "operator-config-file",
+		Required: true,
+		Usage:    "Load operator configurations from `FILE`",
+	}
+	// Optional Flags
```
taturosati: Remove comment
```suggestion
```

core/config/config.go
```diff
@@ -20,201 +19,358 @@ import (
...
-	err = sdkutils.ReadJsonConfig(alignedLayerDeploymentFilePath, &alignedLayerDeploymentRaw)
+
+	return &AggregatorConfig{
+		BaseConfig: baseConfig,
+		Aggregator: aggregatorConfigFromYaml.Aggregator,
+	}, nil
+
+}
+
+func NewOperatorConfig(baseConfigFilePath, operatorConfigFilePath string) (*OperatorConfig, error) {
```
taturosati: If we are using log.Fatal instead of returning error, remove error from return type. Two options here:
1. Go back to return error, and log error instead of fatal (which kills process)
2. Return just *OperatorConfig

core/config/config.go
```diff
@@ -20,201 +19,358 @@ import (
...
+	}
+	OperatorConfigFileFlag = cli.StringFlag{
+		Name:     "operator-config-file",
+		Required: true,
+		Usage:    "Load operator configurations from `FILE`",
+	}
+	// Optional Flags
+)
+
+func NewAggregatorConfig(baseConfigFilePath, aggregatorConfigFilePath string) (*AggregatorConfig, error) {
```
taturosati: If we are using log.Fatal instead of returning error, remove error from return type. Two options here:

1. Go back to return err, and log error instead of fatal (which kills process)
2. Return just *AggregatorConfig

core/config/config.go
```diff
@@ -20,201 +19,358 @@ import (
...
+	}
+
+	return &OperatorConfig{
+		BaseConfig: baseConfig,
+		Operator:   operatorConfigFromYaml.Operator,
+	}, nil
+
+}
+
+func newBaseConfig(baseConfigFilePath string) (*BaseConfig, error) {
```
taturosati: If we are using log.Fatal instead of returning error, remove error from return type. Two options here:

Go back to return error, and log error instead of fatal (which kills process)
Return just *BaseConfig

taturosati (changes requested):
👍🚧

aggregator/cmd/main.go
```diff
@@ -34,11 +34,18 @@ func main() {
...
-		log.Fatal("Error reading aggregator config: ", err)
+
+	baseConfigFilePath := context.String("base-config-file")
+	aggregatorConfigFilePath := context.String("aggregator-config-file")
+
+	aggregatorConfig := config.NewAggregatorConfig(baseConfigFilePath, aggregatorConfigFilePath)
+
+	if aggregatorConfig == nil {
+		log.Fatal("Error creating aggregator config: ")
 	}
```
taturosati: Aggregator config can not be nil

```suggestion
```

taturosati (approved):
lgtm

PR #36, opened 2024-04-25
feat: add util to wait for transaction receipt

### LINKED ISSUE #24: Send task waits 2 seconds instead of waiting for ETH answer
Send task should wait for ethereum receipt

EthClient does not have  a method to wait for receipt.

EigenSDK has a method called WaitForTransactionReceipt in InstrumentedEthClient
But is needs telemetry to work

https://github.com/Layr-Labs/eigensdk-go/blob/master/chainio/clients/eth/instrumented_client.goreceipt := avsWriter.Client.WaitForTransactionReceipt(context.Background(), tx.Hash())

NicolasRampoldi:
- Added util in `core/utils/eth_client_utils.go` to wait for transaction receipt.


MauroToscano:
We need to add this to the operator too

NicolasRampoldi:
> We need to add this to the operator too

Perfect. When the operator is pushed to main I will add it there too 👍

JuArce (approved):

MauroToscano:
> > We need to add this to the operator too
> 
> Perfect. When the operator is pushed to main I will add it there too 👍

I meant in the deposit into strategy

core/utils/eth_client_utils.go
```diff
@@ -0,0 +1,20 @@
+package utils
+
+import (
+	"context"
+	"github.com/Layr-Labs/eigensdk-go/chainio/clients/eth"
+	common2 "github.com/ethereum/go-ethereum/common"
```
JuArce: Import this as `gethcommon` for better readibiity

JuArce (changes requested):

JuArce (approved):

taturosati (approved):
Create an issue to add max retries

PR #58, opened 2024-04-29
chore: add retry when listening for tasks

### LINKED ISSUE #44: chore (aggregator): retry listening for tasks
When aggregator subscription to new tasks on ServiceManager gets disconnected, retry connection up to N times.

NicolasRampoldi:
# Description
- Added retry when listening for tasks. 
- There is a limit of 5 max retries and each retry is done every 10 seconds.
- When the subscription fails, `WatchNewTaskCreated` is called to create the task subscriber again.


taturosati (approved):
lgtm

JuArce (approved):

PR #62, opened 2024-04-29
feat(operator): add gnark plonk verifier

entropidelic:
This PR adds consensys gnark PLONK verifier in the operator


operator/pkg/operator.go
```diff
@@ -73,14 +79,90 @@ func (o *Operator) Start(ctx context.Context) error {
...
+	if err != nil {
+		panic("Could not read PLONK verifying key from bytes")
+	}
+
+	err = plonk.Verify(proof, verificationKey, pubInput)
+	if err != nil {
+		return false
+	} else {
+		return true
+	}
```
taturosati:
```suggestion
        return err == nil
```

simplify

taturosati (approved):
lgtm. tried invalid proof as well & got false

operator/pkg/operator.go
entropidelic: thanks, done!

MauroToscano (approved):
It may be nicer to keep printing the task hash. Signature data is not very useful

PR #70, opened 2024-04-30
chore: add max retries for waitForTransactionReceipt

### LINKED ISSUE #54: feat: add max retries for WaitForTransactionReceipt
# Description
- When doing WaitForTransactionReceipt, instead of making an infinite loop, set a number of max retries.

JuArce:
# Changes
Add a max retry times to waitForTransactionReceipt.
So, if it exceeds max retries to get a tx receipt, the function will return an error
Now the max retries is set to 10. After every try, the function waits 2 second to check again


core/utils/eth_client_utils.go
```diff
@@ -2,19 +2,23 @@ package utils
...
+	"fmt"
 	"github.com/Layr-Labs/eigensdk-go/chainio/clients/eth"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"time"
 )
 
-func WaitForTransactionReceipt(client eth.Client, ctx context.Context, txHash gethcommon.Hash) *types.Receipt {
-	for {
+var maxRetries = 10
```
NicolasRampoldi: Could change to `const MaxRetries = 10`

MauroToscano (approved):

MauroToscano (approved):

NicolasRampoldi (approved):

PR #81, opened 2024-04-30
feat: query if task exists if missing on SubmitTaskResponse

### LINKED ISSUE #50: feat (aggregator): query if task exists if missing
When receiving task response from operator, and the aggregator does not know about task, it should check with rpc if it might be missing task with task index.

On aggregator/internal/pkg/server.go SubmitTaskResponse method

JuArce:
# Description
If aggregator misses tasks from websocket suscription, on SubmitTaskResponse the task will not exist, so before returning an error, will check if the task exists
If the task exists, it will be added to the aggregator storage

# How to test
It will be better to wait until #94 is completed (done) 

1. Run anvil
2. Send a proof without running the aggregator -> The aggregator will lose the event
3. Start the aggregator
4. Send summy response


core/chainio/avs_reader.go
```diff
@@ -54,3 +60,31 @@ func (r *AvsReader) GetErc20Mock(tokenAddr gethcommon.Address) (*contractERC20Mo
...
+	if err != nil {
+		return nil, err
+	}
+
+	for itr.Next() {
+		event := itr.Event
+		if event.TaskIndex == taskIndex {
+			return event, nil
+		}
+	}
```
taturosati: Could be replaced with one call to itr.Next, since we only query about one task index

taturosati (approved):

core/chainio/avs_reader.go
JuArce: Done

NicolasRampoldi (approved):

PR #87, opened 2024-05-01
feat(aggregator): integrate with eigensdk's BLS signature aggregation service

### LINKED ISSUE #72: feat: Aggregate signature and send to ServiceManager

entropidelic:
This PR adds the integration with the BLS aggregation service from the eigensdk.
This service is responsible for aggregating the task responses and signatures from operators, and returning the signature when the threshold is reached.

**NOTE**: For now, threshold is set to 100% in all tasks, meaning that all operators have to verify the sent proofs, and we have only one quorum.

To test the PR: 
* start anvil
* register operator with: `make operator-full-registration`
* start aggregator: `make aggregator-start`
* start operator: `make operator-start`
* send a proof to verify: `make send-plonk-proof`

You should see logs from the operator and aggregator, in particular a `TASK RESPONDED EVENT` dummy log will appear with some data, meaning that the aggregator sent the aggregated signature to the contract and it emitted an event.

**NOTE**: In some cases, specially some seconds after the aggregator is started, if a proof is sent to be verified, the aggregator service replies with a response that is empty (but is not an error). This makes the aggregator fail. The issue for this is in #91.
If this happens, restart the aggregator and then the operator. Wait for a few seconds and send the proof to verify.


taturosati (approved):
lgtm

JuArce (approved):
lgtm

MauroToscano (approved):

PR #108, opened 2024-05-03
feat (aggregator): use different config files

### LINKED ISSUE #106: feat (aggregator): custom config file
On `make aggregator-start` config file is fixed to `config-files/config.yaml`

Should be able to use any path by running like operator:
`make aggregator-start CONFIG_FILE=<path_to_config_file>`

taturosati:
Give user option to use a different config file than default when starting operator.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ anvil-start:
 # TODO: Allow enviroment variables / different configuration files
 aggregator-start:
 	@echo "Starting Aggregator..."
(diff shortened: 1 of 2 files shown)


NicolasRampoldi (approved):

JuArce (changes requested):
There is missing instructions in README

entropidelic (approved):

JuArce (approved):

PR #115, opened 2024-05-03
refactor(operator): refactor scripts into commands of the same CLI

entropidelic:
Refactored the different operator scripts into CLI commands:

* `register`
* `deposit-into-strategy`
* `start`

Also added a makefile target to build every component: aggregator, operator and task sender.
The binaries can be found in the `build` folder of each component, for example `operator/build/aligned-operator`.


taturosati (commented):
looks good, but can we have different files for subcommands? just for clarity.

e.g ` depositIntoStrategyMain` could go in `operator/pkg/deposit_into_strategy.go`

entropidelic:
> looks good, but can we have different files for subcommands? just for clarity.
> 
> e.g ` depositIntoStrategyMain` could go in `operator/pkg/deposit_into_strategy.go`

totally agree, now every action has been refactored into its own file. Thanks!

uri-99 (approved):
passes integration test 👍

JuArce (approved):

PR #116, opened 2024-05-03
refactor(operator): remove hardcoded aggregator server ip port address from operator

### LINKED ISSUE #102: (fix): Stop hardcoding aggregators IP address and port.
In NewOperatorFromConfig function, from ./operator/pkg/operator.go, it would be good if we can avoid hardcoding IP address of the aggregator.

entropidelic:
Closes #102

### DIFF
--- a/config-files/config.yaml
+++ b/config-files/config.yaml
@@ -26,6 +26,7 @@ aggregator:
 
 ## Operator Configurations
 operator:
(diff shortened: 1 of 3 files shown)


taturosati (approved):
lgtm

NicolasRampoldi (approved):

PR #127, opened 2024-05-07
Fix Typo in README.md

lachenlama:
Hi Alignedlayer, was following the steps in the README.md. Found a small typo in there, rectified it. Do check on it. Thanks!


entropidelic:
thank you for the correction. Unfortunately we don't accept typo fix PRs, but we will fix it

PR #130, opened 2024-05-07
feat: different DA solutions

### LINKED ISSUE #124: feat: implement EigenDA
Eigen DA implementation:

- [x] Need to change Task format on ServiceManager to take EigenDA
- [x] Task sender should post proof data to EigenDA, then hash instead of proof on chain
- [x] Operators should retrieve proof from EigenDA based on hash

### LINKED ISSUE #128: feat: implement Celestia
Celestia implementation:

- [x] Add necessary configuration and setup to run Celestia light node.
- [x] Task sender should post proof data to Celestia.
- [x] Operators should retrieve proof from Celestia based on id.

taturosati:
# Changelog

- Add the option for task sender to post proof using either Calldata, EigenDA or Celestia
- Smart contract changed to accept a DA struct that contains which DA solution is being used + the necessary data for the operator to retrieve the proof from the DA system.
- Operator then retrieves proof and verifies it.
- Added config for EigenDA and Celestia connections

# To test

- Run a Celestia node as explained in the README file.
- `make anvil-start`.
- `make aggregator-start`.
- `make operator-full-registration`.
- `make operator-start`
- Access [this](https://faucet.celestia-arabica-11.com/) faucet and enter your account_id to get celestia funds.

Then try sending a task using the different da solutions. Calldata is the default.
1. `make send-plonk_bls12_381-proof`
2. `make send-plonk_bls12_381-proof DA_SOLUTION=eigenda`
3. `make send-plonk_bls12_381-proof DA_SOLUTION=celestia`

Closes https://github.com/yetanotherco/aligned_layer/issues/128
Closes https://github.com/yetanotherco/aligned_layer/issues/124


contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -26,9 +26,21 @@ contract AlignedLayerServiceManager is ServiceManagerBase, BLSSignatureChecker {
...
     // STRUCTS
+    enum DASolution {
+        Calldata,
+        EigenDA,
+        Celestia
+    }
+
+    struct TaskDA {
+        DASolution solution;
+        bytes proof; // Proof bytes for calldata -  BatchHeaderHash for EigenDA - Commitment for Celestia
```
entropidelic: small nit, I think it would be better if this field could have a more generic name like `DAPayload` or something similar. What do you think?

taturosati: Agree, changed on 89788f258aae54a058ef12ed83bb4cb75bbb8b55

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -26,9 +26,21 @@ contract AlignedLayerServiceManager is ServiceManagerBase, BLSSignatureChecker {
...
     // STRUCTS
+    enum DASolution {
+        Calldata,
+        EigenDA,
+        Celestia
+    }
+
+    struct DAPayload {
+        DASolution solution;
+        bytes proof; // Proof bytes for calldata -  BatchHeaderHash for EigenDA - Commitment for Celestia
```
MauroToscano:
```suggestion
        bytes proof_associated_data; // Proof bytes for calldata -  BatchHeaderHash for EigenDA - Commitment for Celestia
```

MauroToscano (approved):
We should post a commitment to the proof and check it matches on the AVS, so we don't trust the DA not changing the proof in the future, and this should work no matter where the Data is stored. But we can do it in another PR

contracts/src/core/AlignedLayerServiceManager.sol
taturosati: changed on d0c81ec0dfe8e3d755b64aeebc4a1b430f2c9434

entropidelic (approved):

PR #164, opened 2024-05-14
fix: check if task exists in addNewTask to avoid race conditions

### LINKED ISSUE #162: fix (aggregator): race condition on addNewTask
# Description
- In the `ProcessOperatorSignedTaskResponse` method, sometimes the task gets added more than once.
- Then, the second time the `blsAggregationService` is called to process a new signature, the following error happens: `"msg":"BLS aggregation service error when initializing new task: task <task_id> already initialized"`

How to test
- It is easier to test in the server due to the RPC latency
- In devnet you can test it with 3 operators running

JuArce:
# Description
In the `ProcessOperatorSignedTaskResponse` method, sometimes the task gets added more than once.
Then, the second time the `blsAggregationService` is called to process a new signature, the following error happens: `"msg":"BLS aggregation service error when initializing new task: task <task_id> already initialized"`

# How to test
- It is easier to test in the server due to the RPC latency
- In devnet you can test it with 3 operators running


NicolasRampoldi (approved):
lgtm, tested it with 3 operators and works properly.

taturosati (approved):
Such a clever solution!

PR #174, opened 2024-05-17
feat: check if batch is already responded

### LINKED ISSUE #172: feat: check if task is already responded in ServiceManager
# Description
In respondToTask method check if the task has already been responded

JuArce:
respondToTask method now checks if batch was already responded

To test you can duplicate lines 200-203 from `aggregator/internal/pkg/aggregator.go`
``` go
	_, err := agg.avsWriter.SendAggregatedResponse(context.Background(), batchMerkleRoot, nonSignerStakesAndSignature)
	if err != nil {
		agg.logger.Error("Aggregator failed to respond to task", "err", err)
	}
```
So the second time it tries to respond the task, the service manager will revert the transaction


taturosati (approved):
lgtm

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -124,6 +128,8 @@ contract AlignedLayerServiceManager is ServiceManagerBase, BLSSignatureChecker {
             "Signatories do not own at least threshold percentage of a quorum"
         );
 
+        batchesState[batchMerkleRoot].responded = true;
```
uri-99: This assignment should be right after the `require` statement it is involved in (line 104), to follow CEI pattern and thus avoid possible re-entrancy attacks.
Even though this function has not much value to be exploited with a reentrancy, it is always a good idea to follow precaution and safety patterns 👍

uri-99 (changes requested):
👀

taturosati (approved):

uri-99 (approved):
👍

PR #176, opened 2024-05-17
docs: update aligned documentation

cliraa:
- [x] An additional section was included describing the AlignedLayerServiceManager contract.
- [x] The video of the talk at ZK Accelerate was added.
- [x] Minor punctuational changes were made.


MauroToscano:
We shouldn't have a Docs folder. All the documentation goes to the README. It's our usual policy. Though it is already there ... we should migrate this later

PR #177, opened 2024-05-17
173 feat task sender submit infinite different proofs

uri-99:
This pr develops the batch task sender, as well as creates an infinite task sender for both GO task sender and batch task sender.

more specific changelog:

- make send_infinite_groth16_bn254_proof
  - sends a different valid proof every 3 seconds, using GO task sender

- make batcher_send_infinite_tasks
  - sends a different valid proof every n seconds, using batcher task sender 
 
- fixed a small bug on task_sender/test_examples/gnark_groth16_bn254_script/main.go
  - output files were called plonk instead of groth16
- fixed a bug when unmarshalling

PR #195, opened 2024-05-22
feat improve batch task sender throughput

### LINKED ISSUE #193: feat: improve batch task sender throughput
to make this feature, the task sender will generate one proof, and send a whole batch to the batch sender with only this proof. It will then generate another proof and send the whole batch with only this proof

1000/10000 per block (12 seconds)

set in anvil the --time flag to set how much time per block.

this would be around 100 per second

this new function should get a number param that is how many proofs does it want to send in a burst of proofs

this should be done with plonk proving system

uri-99:
this pr must have [this pr](https://github.com/yetanotherco/aligned_layer/pull/177) merged before it

Added ability to send bursts of N tasks every second

# To test
1. Run `make make anvil_start_with_block_time`
2. Start aggregator, operator and batcher
3. Run `make batcher_send_burst_groth16`

You can specify burst size by running `make batcher_send_burst_groth16 BURST_SIZE=x`


MauroToscano:
Can we add a log of the amount of proofs verified in a batch ? So we can see how many is the operator receiving ?

MauroToscano (approved):

taturosati (approved):
lgtm

PR #197, opened 2024-05-23
chore(ffi): Change ffi bindings to explicitly be uint32_t

PatStiles:
Based on @Oppen advice this changes the length parameters from usize to u64 removing ambiguity usize has between different architectures.


MauroToscano:
An u32 maybe better in the future so it works in 32 bit architectures

Oppen:
As long as it's consistent. Note most 32 bits architectures do support 64 bits integers, it's just not as efficient as they need to use two registers instead of one, and possibly handle carry in software. Another option is to use `size_t` in all platforms, so it uses the native size type.

Oppen:
The point being, `int` types in C are crap. If at all possible, assume C99 so you can have better integers.

MauroToscano:
> As long as it's consistent. Note most 32 bits architectures do support 64 bits integers, it's just not as efficient as they need to use two registers instead of one, and possibly handle carry in software. Another option is to use `size_t` in all platforms, so it uses the native size type.

But can you then use them as pointers ? Because those are in fact representing pointers.

Oppen:
> But can you then use them as pointers ? Because those are in fact representing pointers.

In neither case can you use non-native integers as pointers. For example, pointer tagging by certain CPUs can break if you use a smaller integer, but also you would alias any region to the low 4GiB, which means at the very least you would need to explicitly map memory in that region rather than take whatever `malloc` or taking the address of a variable gives you. That said, `u32` is a much more suitable offset and is serializable at a constant size. It's also better for compatibility with wasm, which is most often 32 bits regardless of what the native platform is.

(Just added the comment as documentation, since we talked about this privately)

MauroToscano (approved):

entropidelic (approved):

PR #204, opened 2024-05-24
feat (batcher): verify gnark proofs

### LINKED ISSUE #200: feat (batcher): verify gnark proofs
Batcher should verify gnark proofs before it receives before adding them to the batch (to avoid failed batches).

For this to work we will need to add ffi wrapper for go verifiers.

taturosati:
Batcher now verifiers different gnark proofs

# To test:

1. Start anvil, batcher, aggregator & operator
2. Turn on debug logs with `RUST_LOG=debug` environment variable
3. From a different terminal, run:
    - `make batcher_send_plonk_bn254_task`
    - `make batcher_send_plonk_bls12_381_task`
    - `make batcher_send_groth16_bn254_task`
4. See for each of those verification result is true on batcher
5. Batcher will not send batch to s3 & Ethereum unless the blockchain advances on block. There is two ways to do this:
- Start anvil with specified block time
- Send a transaction from another terminal


jotabulacios (approved):
lgtm

MauroToscano:
We should test a bad proof too

taturosati:
@MauroToscano I have tested bad proofs by removing / adding bytes manually from good proofs, I can add a make target if needed

MauroToscano:
Proof is not reaching the blockchain for me

MauroToscano (changes requested):
Not sending proofs to Eth now

batcher/src/gnark/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+        ProvingSystemId::GnarkPlonkBn254 => unsafe {
+            VerifyPlonkProofBN254(&proof, &public_input, &verification_key)
+        },
+        ProvingSystemId::GnarkPlonkBls12_381 => unsafe {
+            VerifyPlonkProofBLS12_381(&proof, &public_input, &verification_key)
+        },
+        ProvingSystemId::Groth16Bn254 => unsafe {
+            VerifyGroth16ProofBN254(&proof, &public_input, &verification_key)
+        },
+        _ => panic!("Unsupported proving system"),
```
MauroToscano, line 51: We shouldn't panic. Return an error and skip

```suggestion
        _ => panic!("Unsupported proving system"),
```

taturosati: Changed on ba5e56841737d800794f6c45d599bcbaea8f169f

uri-99 (approved):

MauroToscano (approved):

PR #205, opened 2024-05-24
fix: add indexed event params + make storage public

### LINKED ISSUE #201: fix: add indexed event params + make storage public
AlignedLayerServiceManager has 2 events:
    event NewBatch(
        bytes32 batchMerkleRoot,
        uint32 taskCreatedBlock,
        string batchDataPointer
    );

    event BatchVerified(bytes32 batchMerkleRoot);
   
neither of them has any [indexed parameters](https://ethereum.stackexchange.com/questions/8658/what-does-the-indexed-keyword-do)

also, the mapping(bytes32 => BatchState) batchesState should be public so this data can be read by the explorer.

uri-99:
In this PR, an indexed event param is added on each event.
Also, the batchStatus storage is made public.

To adapt the rest of the code, the make go bindings were rebuilt, the abi files were replaced, and the aggregator event watcher had to take into account this new indexed param, adding 'nil' to watch for all indexed values.

To test this change, you should be able to run the setup (anvil, aggregator, operator, batcher) and everything should work as before


taturosati (approved):
lgtm

MauroToscano (approved):

PR #206, opened 2024-05-24
feat(read): add verification data commitment

entropidelic:
This PR adds the new verification data commitment that will be used for the read part of Aligned.

The main change is in the way the batch merkle root is calculated: 
Instead of calculating the merkle tree directly from the `VerificationData` batch, a new data structure is created from each `VerificationData`, which is made up of the commitments of all of its internal values (`proof_commitment`, `pub_input_commitment`, ...). The merkle root is calculated from the batch of these new structures, `VerificationDataCommitment`, instead of the original

There is nothing in particular to test, just send proofs using the batcher client like always and check that everything is ok


batcher/src/types/mod.rs
```diff
@@ -15,22 +16,91 @@ pub enum ProvingSystemId {
...
+            pub_input_commitment = hasher.finalize_reset().into();
+        }
+
+        // compute proving system auxiliary data commitment
+        let mut proving_system_aux_data_commitment = [0u8; 32];
+        // FIXME(marian): This should probably be reworked, for the moment when the proving
+        // system is SP1, `proving_system_aux_data` stands for the compiled ELF, while in the case
+        // of Groth16 and PLONK, stands for the verification key.
+        if let Some(vm_program_code) = &verification_data.vm_program_code {
+            debug_assert_eq!(verification_data.proving_system, ProvingSystemId::SP1);
```
taturosati, line 57: Debug assert should be removed

Makefile
```diff
@@ -126,20 +126,20 @@ build_batcher_client:
...
 	@echo "Sending SP1 fibonacci task to Batcher..."
-	@cd batcher/client/target/release && ./batcher-client \
+	@cd batcher/client/ && cargo run --release -- \
 		--proving_system SP1 \
-		--proof ../../test_files/sp1/sp1_fibonacci.proof \
-		--vm_program ../../test_files/sp1/sp1_fibonacci-elf
+		--proof test_files/sp1/sp1_fibonacci.proof \
+		--vm_program test_files/sp1/sp1_fibonacci-elf
 
 batcher_send_groth16_task: batcher/client/target/release/batcher-client
```
taturosati: remove requirement `batcher/client/target/release/batcher-client` if using cargo run
```suggestion
batcher_send_groth16_task:
```

batcher/client/src/main.rs
```diff
@@ -75,27 +89,37 @@ async fn main() {
...
+    let proof_generator_addr: Address = Address::parse_checksummed(addr_str, None).unwrap();
+
+    let verification_data = VerificationData {
         proving_system,
         proof,
-        public_input,
+        pub_input,
         verification_key,
         vm_program_code,
+        proof_generator_addr,
```
taturosati, line 103: Im not sure what the proof_generator_addr represents. Is it to handle payments to the batcher? im not sure why its needed in the verification data

MauroToscano: It's needed for the full read flow. It's to avoid someone watching the blockchain frontrunning the submission of your proofs.

batcher/src/types/mod.rs
entropidelic: I agree, I added it just as an extra check to make the development process easier. That's why I left the fixme but it could be removed

MauroToscano: It can stay

MauroToscano (approved):

uri-99 (changes requested):
loving the cargo run refactor on make targets.
I think this change should also be present in batcher/send_infinite_tasks.sh

NicolasRampoldi (approved):
lgtm, tested it and it works

uri-99 (approved):
✅

PR #216, opened 2024-05-27
refactor batcher and add batch merkle root response to client

entropidelic:
~~Depends on #206~~

This PR refactors the batcher to make it a little simpler while at the same time adding some functionality for the connected clients to receive the merkle root of the batch where their sent proof was included.

One of the main changes of the refactor is that now, the thread that listens for new blocks is the one in charge of checking if the conditions for submitting a batch are met. Once they are, the merkle root of the finalized batch is sent to all the connected clients who sent proofs to the batcher. In conclusion, new blocks are the ones that trigger batch submission.

To test the PR:
1. start anvil, operator and aggregator
2. start the batcher
3. in another terminal, run `make batcher_send_groth16_burst_5`. This will send 5 groth16 proofs, which will mean that the batch is ready to be sent, but first we need to receive a new block.
4. In another terminal, run `make operator_get_eth` so that some new blocks are produced. This will finalize de batch and calculate the merkle root. This merkle root will be sent as a response to the client and the connection will be closed.


batcher/src/lib.rs
```diff
@@ -31,275 +28,266 @@ pub mod types;
...
 
-pub struct App {
+pub struct Batcher {
     s3_client: S3Client,
-    eth_rpc_provider: Provider<Http>,
+    eth_ws_provider: Provider<Ws>,
     service_manager: AlignedLayerServiceManager,
-    sp1_prover_client: ProverClient,
-    eth_ws_url: String,
+    // eth_ws_url: String,
```
taturosati: Remove comment

```suggestion
```

batcher/src/lib.rs
```diff
@@ -31,275 +28,266 @@ pub mod types;
...
+            //             verify_gnark(&verification_data.proving_system, proof, public_inputs, vk);
+
+            //         debug!("Proof is valid: {}", is_valid);
+
+            //         if is_valid {
+            //             Ok(())
+            //         } else {
+            //             Err(anyhow::anyhow!("Failed to verify proof"))
+            //         }
+            //     }
```
taturosati: Remove comment

```suggestion
```

uri-99 (changes requested):
since new blocks will be the trigger to submit batches, i propose changing in this pr the following make targets:

- anvil_start should be removed
- anvil_start_with_block_time should be renamed to anvil_start

batcher/src/lib.rs
```diff
@@ -31,275 +28,266 @@ pub mod types;
...
+            return;
+        }
+        let (batch_bytes, batch_merkle_root) =
+            self.process_batch_and_update_state(block_number).await;
 
-            info!("Batch sent to S3 with name: {}", file_name);
-            info!("Uploading batch to contract");
+        let s3_client = self.s3_client.clone();
+        let service_manager = self.service_manager.clone();
+        // tokio::spawn(async move {
```
taturosati: Remove comment
```suggestion
```

batcher/src/lib.rs
```diff
@@ -31,275 +28,229 @@ pub mod types;
...
-                let hash = hasher.finalize().to_vec();
-
-                Ok(hash)
-            }
-            Err(e) => Err(e.to_string()),
+        if verification_data.verify() {
+            self.add_to_batch(verification_data).await;
+        // FIXME(marian): This condition should be removed once we have FFI bindings with Gnark.
+        } else if verification_data.proving_system == ProvingSystemId::Groth16Bn254 {
+            self.add_to_batch(verification_data).await;
```
taturosati: This condition can already be removed right

```suggestion
```

entropidelic: good catch, fixed 13e9c56dc993d19b17b088fd1d3b4b2d1920b1a1

taturosati (approved):
lgtm

jotabulacios (approved):

uri-99 (approved):

PR #219, opened 2024-05-28
security: data size checks

### LINKED ISSUE #213: Security: Data Size Checks
- [x] Operator checks the batch size is less than 256 MiB before downloading (This takes 2s with 1Gbps) 
- [x] Batcher sends task if the size of the batch is 256 MiB (Add a config file)
- [x] Batcher limits individual proof data size to 64 MiB

taturosati:
Added 3 data size checks for security:
- Operator checks full batch size is less than 256MiB
- Batcher checks each individual proof is less than 64MiB
- Batcher checks full batch is less than 256MiB
    - if it isn't its not ignored, but proofs that do not fit in this batch will be sent on the next one

All this sizes are configurable in config.yaml

# To test

First change config.yaml to a lower max size per proof and for batch.
I found that:
- max_batch_size: 16000
- max_proof_size: 4000

Work well to check the limits are working. Make sure to change operator.max_batch_size as well.

Then, you will need to start anvil, aggregator, operator and batcher.

Then:
1. Run `make batcher_send_groth16_burst_5` twice to send 10 groth_16 proofs
2. Manually advance chain by sending dummy txs. (e.g make operator_get_eth)
3. You should see the batcher send a batch of 6 proofs first and then a second batch of 3 proofs
4. Run `make batcher_send_sp1_task` to send one sp1 task
5. It should fail because proof size is too large
6. Restart anvil (to avoid Merkle root being repeated)
7. Change batcher max_batch_size to a bigger value (doubling it works fine) & restart batcher.  Another option is to change operator max_batch_size to a much smaller value (e.g setting it to 10) & restarting operator.
8. Run `make batcher_send_groth16_burst_5` twice to send 10 groth_16 proofs
9. Manually advance chain by sending dummy txs. (e.g make operator_get_eth)
10. Operator should not verify batch because batch size is too large


NicolasRampoldi (approved):
lgtm, tested it and it works

jotabulacios (approved):
lgtm

PR #224, opened 2024-05-28
Update readme

MauroToscano:
- add ToC
- change cds to pushd and popd, so commands don't move (cargo -C doesn't work with different toolchains)
- in the batcher client cli, removed --release to improve compilation time, since it doesn't need the optimizations


taturosati (approved):
lgtm

README.md
```diff
@@ -114,133 +149,109 @@ Install eigenlayer-cli:
...
 
-In another terminal window:
-
-```bash
-make operator_full_registration
-```
-
 ```bash
-make operator_start
+make operator_register_and_start
```
entropidelic, line 175: should we add some directives too for the case where the operator is already registered and has to be restarted? It could be something like:

---

The first time booting a new operator, it needs to be registered with EigenLayer and Aligned before starting it. To do this all in one shot run

```bash
```
**Optional**: If for any reason you want to restart the operator and the local anvil node is still running, it doesn't need to be re-registered. If you do so, an error will be prompted telling that the operator is already registered. 
If this is the case, simply run 

```bash
make operator_start
```

README.md
```diff
@@ -280,9 +291,7 @@ To register an operator in EigenLayer and Aligned and deposit strategy tokens in
 make operator_full_registration CONFIG_FILE=<path_to_config_file>
 ```
 
-### Deposit Strategy Tokens
-
-#### Anvil
+##### Deposit Strategy Tokens in Anvil
```
entropidelic: small nit, just to be super clear that anvil is the local devnet
```suggestion
##### Deposit Strategy Tokens in Anvil local devnet
```

README.md
MauroToscano: This is smart, I added this

entropidelic (approved):

PR #233, opened 2024-05-29
refactor: batcher create a shared crate between the cli and the batcher

### LINKED ISSUE #223: (batcher): Create a shared crate betweenthe CLI and the Batcher
Create a shared crate with common library between the CLI and the Batcher, as to avoid compiling the whole batcher for the CLI. CLI should only use this.
 
 This should be done with workspaces
 
 https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html
 
 Main Cargo should have this
 
```
 members = [
    "aligned-batcher-client",
    "aligned-batcher",
    "aligned-batcher-lib"
]
```

NicolasRampoldi:
# Description

- Batcher now has 3 different crates:
  - `aligned-batcher` (what was called **batcher** before).
  - `aligned` (what was called **client** before).
  - `aligned-batcher-lib` (**commons** between batcher and client).

# Results

- Now `aligned` builds in ~15s without `--release` since it doesn't depend anymore on building the batcher.

# How to test

- First run `mv batcher/.env batcher/aligned-batcher/.env` to move the .env file to the corresponding directory.
- To test make sure every batcher command present in the makefile works as expected.


MauroToscano (approved):

batcher/aligned-batcher-client/generate_proof_and_send.sh
```diff
@@ -16,7 +16,7 @@ fi
 echo "Generating proof $x != 0"
 go run task_sender/test_examples/gnark_groth16_bn254_infinite_script/cmd/main.go $x
 
-./batcher/client/target/release/batcher-client \
+./batcher/aligned-batcher-client/target/release/batcher-client \
```
uri-99: run with cargo run plase :)

uri-99 (changes requested):

batcher/aligned-batcher-client/generate_proof_and_send.sh
NicolasRampoldi: We should change that in another PR

uri-99 (approved):
worked for me

PR #244, opened 2024-05-29
fix(batcher): send correct responses to each client

entropidelic:
Up to now, the batcher responded with the most recent processed batch merkle root to all the connected clients. This was not a correct behaviour because clients should only receive responses from batches where their verification data was included.

To test it:
1. start anvil with `make anvil_start_with_block_time` so that a new block is submitted every 3 seconds.
2. start aggregator and operator.
3. start batcher
4. send a burst of 15 proofs with one of `make batcher_send_groth16_burst`, `make batcher_send_plonk_bls12_381_burst`, `make batcher_send_plonk_bn254_burst` or `make batcher_send_sp1_burst`. You can try this with one or more terminals at the same time. For every message sent by the client, there should also be one response. The responses will be logged as the merkle root of the batch the proof was included in and the hash of the proof. Once all the responses are received, the client will close the connection. If you find that some connection hangs and doesn't close, it may be that the batch was left with only 1 proof. If this happens, try sending an individual proof from another terminal using `make batcher_send_plonk_bn254_task`. The batch should be finalized and a response will arrive for both clients, closing the connection

**NOTE**: For the moment the responses are sent once the batch is processed by the batcher. In a future PR this should be handled in such a way so that it is responded once the full flow on aligned has been done (proofs confirmed in Ethereum)


batcher/src/lib.rs
```diff
@@ -186,41 +184,59 @@ impl Batcher {
     ///     * Has the received block number surpassed the maximum interval with respect to the last posted batch block?
     /// An extra sanity check is made to check if the batch size is 0, since it does not make sense to post
     /// an empty batch, even if the block interval has been reached.
-    async fn batch_ready(&self, block_number: u64) -> bool {
-        let current_batch_size = self.current_batch.lock().await.len();
+    /// Once the batch is ready to be finalized, a copy of it is made to be passed to the `finalize_batch` function
+    /// and the current batch queue is cleared so that new data can be added to the batch while the old one is being
+    /// processed.
+    async fn is_batch_ready(&self, block_number: u64) -> Option<BatchQueue> {
```
taturosati, line 193: I would change the name of this method to something more descriptive

batcher/src/lib.rs
```diff
@@ -229,63 +245,88 @@ impl Batcher {
...
                 current_batch_end
             );
-            batch_to_send = current_batch.drain(..current_batch_end).collect::<Vec<_>>();
+            batch_to_send = batch_verification_data.drain(..current_batch_end).collect();
 
             info!(
-                "# of Elements remaining for next batch: {}",
-                current_batch.len()
+                "Number of elements remaining for next batch: {}",
+                finalized_batch.len()
```
taturosati: should be batch_verification_data.len()

batcher/src/lib.rs
```diff
@@ -229,63 +245,88 @@ impl Batcher {
...
             );
             batch_bytes = serde_json::to_vec(&batch_to_send).expect("Failed to serialize batch");
-        } else {
-            batch_to_send = current_batch.clone();
-            current_batch.clear();
         }
 
-        let batch_commitment = VerificationCommitmentBatch::from(&batch_to_send);
+        info!("Finalizing batch. Size: {}", finalized_batch.len());
```
taturosati, line 260: should be batch_to_send

taturosati (changes requested):
I think max batch size check is not working properly, it is deleting elements from queue but if they are not sent on a batch it does not re add them. Maybe logic for max size should be moved to is_batch_ready function

batcher/src/lib.rs
MauroToscano: +1

taturosati:
> I think max batch size check is not working properly, it is deleting elements from queue but if they are not sent on a batch it does not re add them. Maybe logic for max size should be moved to is_batch_ready function

Actually it sends the whole batch, but does not respect max batch size

MauroToscano:
Logic to handle batches should be on is_batch_ready, not in finalize

jotabulacios (approved):
Everything worked as intended , there was no need to run the extra task

batcher/src/lib.rs
```diff
@@ -173,11 +164,18 @@ impl Batcher {
         Ok(())
     }
 
-    async fn add_to_batch(self: Arc<Self>, verification_data: VerificationData) {
+    /// Adds verification data to the current batch.
```
MauroToscano:
```suggestion
    /// Adds verification data to the current batch queue.
```

MauroToscano (approved):

taturosati (approved):
lgtm

PR #245, opened 2024-05-29
243 aggregator signature aggregator mixes batches making the agg response fail

### LINKED ISSUE #243: ( Aggregator ): Signature aggregator mixes batches, making the agg response fail

MauroToscano:
Fixed the bug by having maps with Roots and Indexes, and not assuming the  Signature is always working over the last element. 

Additionally I removed some data that wasn't being used or signed from the SignedResponse of the operator, changed the contract pragma so it can compile, and renamed a variable from task to batch for consistency


aggregator/internal/pkg/aggregator.go
```diff
@@ -193,44 +210,65 @@ func (agg *Aggregator) sendAggregatedResponseToContract(blsAggServiceResp blsagg
...
+	batchMerkleRoot := agg.batchesRootByIdx[blsAggServiceResp.TaskIndex]
+	agg.batchesRootByIdxMutex.Unlock()
 
 	_, err := agg.avsWriter.SendAggregatedResponse(context.Background(), batchMerkleRoot, nonSignerStakesAndSignature)
 	if err != nil {
 		agg.logger.Error("Aggregator failed to respond to task", "err", err)
 	}
 }
 
+// BatchIndex
```
entropidelic: small nit, but I think this is not needed

contracts/src/core/ERC20Mock.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts (last updated v4.8.0) (token/ERC20/ERC20.sol)
 
-pragma solidity ^0.8.9;
+pragma solidity =0.8.12;
```
entropidelic, line 4: is this change ok? seems unrelated to the PR

aggregator/internal/pkg/aggregator.go
MauroToscano: I missed it, thanks

entropidelic (approved):
lgtm, just left some small nits

contracts/src/core/ERC20Mock.sol
MauroToscano: It is, but make bindings is not working on main, I solved some small issues

entropidelic: great, thanks

NicolasRampoldi (approved):
lgtm, tested it and it works as expected

PR #271, opened 2024-05-31
fix (batcher): gnark verifier bindings on linux

taturosati:
Gnark verifier bindings on linux were not working. 

This PR fixes those issues. It has been tested on linux.

# To test
1. Get a linux machine (old code worked on docker so docker does not count)
2. Start anvil, batcher, aggregator and operator
3. Run `make batcher_send_groth16_bn254_task`, `make batcher_send_plonk_bls12_381_task` and `make batcher_send_plonk_bn254_task`
4. Batcher should verify them correctly


batcher/gnark/verifier.go
```diff
@@ -1,34 +1,55 @@
...
+#include <stdint.h>
+
+typedef struct ListRef {
+  const uint8_t *ptr;
+  uintptr_t len;
+} ListRef;
+
+
+*/
+import "C"
```
JuArce, line 14: Is it posible to move the import with all the imports?

JuArce (changes requested):

batcher/gnark/verifier.go
taturosati, line 14: unfortunately no, because of how CGo works, the comment has to ve directly above import.

https://pkg.go.dev/cmd/cgo#hdr-Using_cgo_with_the_go_command

JuArce (approved):

MauroToscano (approved):

PR #295, opened 2024-06-04
fix (aggregator): estimate respondToTask gas limit correctly

### LINKED ISSUE #238: fix: estimate RespondToTask gas limit correctly
Goethereum is underestimating gas limit for RespondToTask
So, we fixed to a hardcoded value

taturosati:
We had to hardcode gas estimation to 500k beacause it was reverting because estimate was not sufficient. Now we add 10% to the estimate instead.

# To test
Tested in prod and seems fine


JuArce (approved):

NicolasRampoldi (approved):
lgtm

PR #307, opened 2024-06-05
fix: batcher nonce are not increasing if receipt is not received so transaction fails

NicolasRampoldi:


NicolasRampoldi:
Couldn't replicate the error.

PR #317, opened 2024-06-05
fix (aggregator): race condition with task index plus bug

### LINKED ISSUE #304: ( Aggregator): Race Condition with Task Index plus bug
There are some possible race conditions and problems with the code. We should:

- Use one lock for all the maps and the task indexes, and update them all together. Since the 3 values should always be consistent 
- In ProcessOperatorSignedTaskResponse, don't assume the task is already loaded. Wait if it isn't

taturosati:
- Changed aggregator mutex implementation (used to have 4 different ones, now has only one)
- Removed unused struct OperatorTaskResponses

# To test

Nothing in particular, test normal flow


MauroToscano (approved):

MauroToscano:
We merged a PR that solved this. This is redundant

PR #337, opened 2024-06-07
fix (cli): error handling

### LINKED ISSUE #333: fix (cli): error handling
Fix error handling on batcher client, should remove expects and provide more descriptive errors

taturosati:
- Error messages in batcher client are more descriptive
- Removed warnings from missing files when not necessary for that proving system
    - Batcher client prints error that parameter is required for proving system

# To test

1. Test normal flow (some batcher_send make targets)
2. Test sending proof with batcher offline
3. Test sending proof with malformed url (--conn asdf)
4. Test sending proof with missing file (--proof ./asdf)
5. Test sending proof with missing param (e.g --proving_system Groth16Bn254 and don't specify --public_input)


MauroToscano (approved):

batcher/client/src/main.rs
```diff
@@ -157,8 +122,59 @@ async fn receive(
...
+        proof_generator_addr,
+    })
+}
+
+fn read_file(file_name: PathBuf) -> Result<Vec<u8>, BatcherClientError> {
+    std::fs::read(&file_name)
+        .map_err(|e| BatcherClientError::IoError(file_name, e))
+}
+
+fn read_file_option(param_name: &str, file_name: Option<PathBuf>) -> Result<Vec<u8>, BatcherClientError> {
```
entropidelic, line 177: it might be better to have an Enum for the param names but it is ok for now

entropidelic (approved):

PR #338, opened 2024-06-07
Start burst from given number

jotabulacios:
to test run:

```bash
make batcher_send_burst_groth16 START_COUNTER=15
```
In this case, the burst will start from the counter value 15, generating and processing proofs sequentially from this starting point.

If no START_COUNTER is provided, it will default to starting from 1:

```bash
```

### DIFF
--- a/batcher/client/send_burst_tasks.sh
+++ b/batcher/client/send_burst_tasks.sh
@@ -1,10 +1,8 @@
 #!/bin/bash
 
-# Valores por defecto
 counter=1
 burst=8
 
-# Procesar el argumento para el valor de burst
 if [ -z "$1" ]; then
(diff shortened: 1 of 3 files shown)


batcher/client/send_burst_tasks.sh
```diff
@@ -1,24 +1,36 @@
 #!/bin/bash
 
+# Valores por defecto
```
taturosati: comments should be in English

jotabulacios: My bad. 
Removed

taturosati (approved):
lgtm

entropidelic (approved):

JuArce (approved):

PR #344, opened 2024-06-07
feat(aggregator): add check for Tx inclusion for Aggregated Responses in case of Re-orgs

PatStiles:
After a successful tx is sent by the aggregator to the contract check the aggregator waits 30 sec before confirming the Tx receipt is included in case of a re-org.


MauroToscano:
Don't error, go back at retrying

PatStiles:
Done!

MauroToscano (commented):
Code seems alright. Give it two tests before merging

taturosati (commented):
Code looks good

MauroToscano:
Stale

PR #351, opened 2024-06-10
Fix infinite loop in batcher burst

### LINKED ISSUE #347: (task sender): Burst loop should be infinite again
Burst loop changed it's behaviour. It was good for it to start from a given public input and continue building as to not repeat proofs, but it shouldn't send N burst of size N. The burst size and the amount of iterations is unrelated. It should be an infinite loop

jotabulacios:
to test run:

```bash
make batcher_send_burst_groth16 START_COUNTER=15
```
In this case, the burst will start from the counter value 15, generating and processing proofs sequentially from this starting point.

If no START_COUNTER is provided, it will default to starting from 1:

```bash
```

closes #347

### DIFF
--- a/batcher/client/send_burst_tasks.sh
+++ b/batcher/client/send_burst_tasks.sh
@@ -3,8 +3,9 @@
 counter=1
 burst=8
 
+# Establecer el valor de 'burst' desde el primer argumento
 if [ -z "$1" ]; then
-    echo "Using default burst value: 10"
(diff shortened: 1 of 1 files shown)


MauroToscano (changes requested):
- Remove comments in spanish
- Loop should be infinite

MauroToscano (approved):

JuArce (approved):

PR #365, opened 2024-06-11
fix(batcher) add timeout and retries logic to submit batch function

### LINKED ISSUE #349: fix(batcher): add timeout and retries logic to submit_batch function

JuArce:
> [!NOTE]
> This PR is missing Testing and Merge Conflicts before Ready for Review

I added this timeout logic to the wait for tx receipt
```rust
match timeout(Duration::from_secs(60), pending_tx).await {
        Ok(Ok(Some(receipt))) => Ok(receipt),
        Ok(Ok(None)) => Err(anyhow::anyhow!("Transaction was mined but no receipt was returned")),
        Ok(Err(e)) => Err(anyhow::anyhow!("Transaction failed: {:?}", e)),
        Err(_) => Err(anyhow::anyhow!("Transaction was not mined within 60 seconds")),
    }
```
I think `Err(_)` case should return `Ok(None)` instead of `Err(...)` because this case does not mean that the tx failed, but the receipt was not received @MauroToscano @entropidelic 

To test the normal behavior, run anvil and the aggregator and send proofs

### DIFF
--- a/batcher/src/eth/mod.rs
+++ b/batcher/src/eth/mod.rs
@@ -1,9 +1,11 @@
 use std::str::FromStr;
 use std::sync::Arc;
+use std::time::Duration;
 
 use ethers::prelude::k256::ecdsa::SigningKey;
 use ethers::prelude::*;
 use stream::EventStream;
+use tokio::time::timeout;
 
 use crate::config::ECDSAConfig;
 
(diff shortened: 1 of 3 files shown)


MauroToscano:
Stale

PR #378, opened 2024-06-13
feat (operator): Dockerfile

MauroToscano:
- Added dockerfile
- Added guide on how to use dockerfile
- Operator now registers on startup if not registered

# To test

Follow instructions in Readme on how to register operator (using docker version)


README.md
```diff
@@ -179,38 +175,45 @@ The eigen guide can be found [here](https://docs.eigenlayer.xyz/eigenlayer/resta
...
 - [Holesky PoW Faucet](https://holesky-faucet.pk910.de/)
 
-### Register as an operator with Aligned
+### Start the operator
 
-To register the operator with Aligned, run:
+#### Using Docker
 
 ```bash
-./operator/build/aligned-operator register --config ./config-files/config-operator.yaml
```
entropidelic: shouldn't we preserve this? if the docker is doing it internally al least it should be in the `From source` section

entropidelic: nvm, just seen the added `register` function

entropidelic (approved):

jotabulacios (approved):

PR #381, opened 2024-06-13
feat(batcher): add bool configuration option for enabling pre-verification of proofs

entropidelic:
Add `enable_pre_verification` to batcher config so that we can configure if batch should pre-verify received proofs

The important change is in the `batcher/aligned-batcher/src/lib.rs` file, the others are because of formatting


NicolasRampoldi (approved):

config-files/config.yaml
```diff
@@ -24,6 +24,7 @@ batcher:
   max_proof_size: 67108864 # 64 MiB
   max_batch_size: 268435456 # 256 MiB
   eth_ws_reconnects: 99999999999999
+  enable_pre_verification: false
```
MauroToscano: True by default

```suggestion
  enable_pre_verification: true
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -50,6 +51,7 @@ pub struct Batcher {
     max_proof_size: usize,
     max_batch_size: usize,
     last_uploaded_batch_block: Mutex<u64>,
+    enable_pre_verification: bool,
```
```suggestion
   pre_verification_is_enabled: bool,
```

MauroToscano (changes requested):

MauroToscano (approved):

PR #382, opened 2024-06-13
feat(explorer): extend database to reduce amount of rpc calls

### LINKED ISSUE #322: feat(explorer): extend database to relieve rpc calls
Currently, every explorer request is forwarded to the RPC.
This had very low performance on prod.

We started using our own RPC, which increased the performance, but will still be unsustaniable with many users.

We should store some of the basic blockchain info to our postgres-ecto db to make it more performant, relieve the pressure on RPC, and make it less dependant of it.

uri-99:
# Changes

## Backend

* Added columns to db
* made fetch_old_batches
* Improve performance eliminating unnecessary RPC calls by replacing them with the db

## Frontend

* Update twitter background
* Update home background
* Add more detail for verified tasks in batch view
* Add timestamp to batches table, tweak detail timestamp
* Update color palette

#### Before

image

#### After

image
#### Before

image

#### After

image

#### Before

image

#### After

image


MauroToscano (approved):

explorer/.env.example
```diff
@@ -11,3 +11,6 @@ DB_NAME=<db_name>
 DB_USER=<db_user>
 DB_PASS=<db_pass>
 DB_HOST=<db_host>
+
+# Config file
+ALIGNED_CONFIG_FILE="<aligned_config_file>"
```
NicolasRampoldi: Missing new line

explorer/lib/explorer/aligned_layer_service_manager.ex
```diff
@@ -28,7 +31,7 @@ defmodule AlignedLayerServiceManager do
   @first_block (
     case @environment do
       "devnet" -> 0
-      "holesky" -> 1600000
+      "holesky" -> 1728056
       "mainnet" -> 20020000
```
NicolasRampoldi, line 35: I think we should add holesky stage here. Also, it would be nice if we could set this in environment variables

explorer/lib/scripts/fetch_old_batches.ex
```diff
@@ -10,7 +10,7 @@ defmodule Scripts.FetchOldBatches do
 
   def fetch_old_events(fromBlock, toBlock) do
     "fetching old events, from #{fromBlock} to #{toBlock}" |> IO.inspect()
-    chunk_size = 5000 #do in smaller chunks as to not request 100k events to rpc in one call
+    chunk_size = 3 #do in smaller chunks, if there are too many blocks to process
```
NicolasRampoldi: Maybe the chunk size could be parametrized

NicolasRampoldi (changes requested):

explorer/lib/explorer/aligned_layer_service_manager.ex
uri-99: This separation was develpoed but not really useful, both use holesky so separating holesky stage and holesky prod initial blocks is not currently necesarry, both use the same values

NicolasRampoldi (approved):

PR #428, opened 2024-06-17
fix (aggregator): dont wait for tx before listening new bls aggregation responses

### LINKED ISSUE #421: fix (aggregator): don't wait for tx before listening new bls aggregation responses
Aggregator behavior is linear at the moment, so it's not listening for new bls aggregation service responses until it gets its receipt from its previous tx. We need to fix this. To do this the sending of the response will be moved to a separate go routine but a lock needs to be added to protect the ethereum wallet.

taturosati:
- Aggregator now spawns go routine for handling of bls aggregation services responses
- Added lock for sending tx to ethereum
- Fixed issue where lock was unlocked twice when batch was not found
- Improved logs

# To test:

- Locally just check everything is working normally
- Being tested in stage deployment on holesky, can check explorer at http://stage.alignedlayer.com/batches


aggregator/internal/pkg/aggregator.go
```diff
@@ -228,9 +233,35 @@ func (agg *Aggregator) sendAggregatedResponseToContract(blsAggServiceResp blsagg
 		"merkleRoot", hex.EncodeToString(batchMerkleRoot[:]))
 }
 
+func (agg *Aggregator) sendAggregatedResponse(batchMerkleRoot [32]byte, nonSignerStakesAndSignature servicemanager.IBLSSignatureCheckerNonSignerStakesAndSignature) (*gethtypes.Receipt, error) {
```
MauroToscano: Document that it waits for the receipt

MauroToscano (approved):

JuArce (approved):

PR #449, opened 2024-06-19
feat: first version of payment system

entropidelic:
This PR implements the first version of payment system. The service manager contract now has a mapping `batchersBalances`, with balances for each batcher. 
`batcherAddress` is added to the BatchState struct of the service manager.

When the function call to respondToTask is called by the aggregator, the cost of the transaction is discounted from the batcher balance in the service manager and sent to the aggregator.

### Testing
The happy path can be tested checking the full flow with the usual commands.

For testing the new details from this PR
* Start anvil, aggregator and the operator as usual.
* Start batcher with
```
cargo +nightly-2024-04-17 run --manifest-path ./batcher/aligned-batcher/Cargo.toml --release -- --config ./config-files/config.yaml --env-file ./batcher/aligned-batcher/.env
```
* Send some proofs and check that the batch verification task transaction is reverted because batcher has no funds.
* Send some funds to the batcher with
```
make batcher_fund_service_manager_balance
```
* Send proofs and check that everything works as expected.

The commands to fund the batcher is the one that is run in `make batcher_start`.


taturosati (approved):
lgtm

MauroToscano (changes requested):
Hold until confirmation

uri-99 (approved):

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -149,4 +178,22 @@ contract AlignedLayerServiceManager is
...
+
+    function withdraw(uint256 value) external {
+        require(
+            batchersBalances[msg.sender] >= value,
+            "The value to withdraw is greater than deposited funds"
+        );
+
+        batchersBalances[msg.sender] -= value;
+        payable(msg.sender).transfer(value);
+    }
```
uri-99: remove withdraw function
```suggestion
```

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -56,16 +56,19 @@ contract AlignedLayerServiceManager is
     function createNewTask(
         bytes32 batchMerkleRoot,
         string calldata batchDataPointer
-    ) external payable {
+    ) external {
```
uri-99: Make payable to accept payment of response from here as well

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -149,4 +182,12 @@ contract AlignedLayerServiceManager is
...
             );
     }
+
+    function balanceOf(address account) public view returns (uint256) {
+        return batchersBalances[account];
+    }
+
+    receive() external payable {
+        batchersBalances[msg.sender] += msg.value;
+    }
```
uri-99, line 192: This is already being done in `createNewTask()`. We should either remove this function, or add a function in the batcher contract to only send funds to this contract.
This is because now the funds mapping is to the batchers' contract, and not his wallet.
I think removing this function is the simpler solution

contracts/src/core/AlignedLayerServiceManagerStorage.sol
```diff
@@ -1,14 +1,20 @@
...
         uint32 taskCreatedBlock;
         bool responded;
+        address batcherAddress;
     }
 
     /* STORAGE */
     mapping(bytes32 => BatchState) public batchesState;
 
+    // Storage for batchers balances. Used by aggregator to pay for respondToTask
+    mapping(address => uint256) public batchersBalances;
```
uri-99:
```suggestion
    mapping(address => uint256) internal batchersBalances;
```
there is already a prettier `balanceOf(address) returns (uint256)` in AlignedLayerServiceManager.sol

MauroToscano (approved):

MauroToscano (approved):

PR #502, opened 2024-06-26
Add Linux dependencies & systemd service

OsamaMomani:
- Add make and build-essential dependencies to operator guide
- Add systems service file


MauroToscano:
If you want to discuss an improvement please create an issue before creating the PR. The additional requirements seems fine, we will add it.

PR #503, opened 2024-06-26
feat: add version check to batcher and client

### LINKED ISSUE #500: feat: add version to batcher and client
# Description

- Batcher and client should have a shared version. If the client isn't in the required version, when it connects to the batcher it should tell that it needs to update.

NicolasRampoldi:
# Description

- This PR makes the batcher sent the current protocol version to the client. If the client is in an older version, it suggests to update it.

# To Test

- In line 44 of `aligned-batcher/src/lib.rs`, change the `PROTOCOL_VERSION` constant to 1 and then start the batcher as usual.
- Then, run `make install_aligned_compiling` and run `which aligned` to make sure you are using the `/.cargo/bin/aligned` binary.
- Then, run `make batcher_send_sp1_task`, you should see a message like:

```
[2024-06-26T17:19:59Z INFO  aligned] You are running an old version of the client, update it by installing it again running:
    curl -L https://raw.githubusercontent.com/yetanotherco/aligned_layer/main/batcher/aligned/install_aligned.sh | bash
    Client version: 0, Expected version: 1
```


taturosati (approved):
lgtm

MauroToscano (approved):

PR #510, opened 2024-06-27
feat(verifier): Add verifier for Axiom's Halo2 fork

PatStiles:
Adds verifier for Axiom's fork of Halo2.


uri-99:
“Found Unneeded after exploring exploration with zk-email. Zk-emails implementation was using an out of date version of halo2-axiom” - @PatStiles

PR #519, opened 2024-06-28
feat: integrate batcher with payment system

entropidelic:
Integrates batcher with the new batcher payment service.
When a user sends a proof, the message is signed and the batcher verifies the signature and recovers the user address. Then, it checks that the address has funds in the batcher payment service.

### Testing:
 For fine grained testing:
 * Start anvil with block time, aggregator and operator
 * Start batcher with
 ```bash 
cargo +nightly-2024-04-17 run --manifest-path ./batcher/aligned-batcher/Cargo.toml --release -- --config ./config-files/config-batcher.yaml --env-file ./batcher/aligned-batcher/.env
```
This will start the batcher without sending funds.
* Send some proofs with
```bash
aligned submit --proving_system GnarkPlonkBls12_381 \
--proof batcher/aligned/test_files/plonk_bls12_381/plonk.proof \
--public_input batcher/aligned/test_files/plonk_bls12_381/plonk_pub_input.pub \
--vk batcher/aligned/test_files/plonk_bls12_381/plonk.vk --repetitions 2 --keystore_path config-files/anvil.ecdsa.key.json
```
An error will be logged meaning that the user has no funds in the batcher payment system.
* Send some funds with
```bash
cast send 0x7969c5eD335650692Bc04293B07F5BF2e7A673C0 --value 1ether --rpc-url http://localhost:8545 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
```
* Send proofs again, everything should work as usual.

## Happy path:
* Start anvil again so that the previous state is reseted, and do the usual flow. Everything should work as usual. The only difference now is that when `make batcher_start` is called, a transaction is made from the user to the payment system to add funds in devnet.


Makefile
```diff
@@ -181,7 +181,7 @@ batcher_fund_service_manager_balance:
 ./batcher/aligned-batcher/.env:
 	@echo "To start the Batcher ./batcher/aligned-batcher/.env needs to be manually set"; false;
 
-batcher_start: ./batcher/aligned-batcher/.env batcher_fund_service_manager_balance
+batcher_start: ./batcher/aligned-batcher/.env
```
uri-99: I think we should also delete the whole `make batcher_fund_service_manager_balance` target

uri-99 (changes requested):
my review is still wip

batcher/aligned-batcher-lib/Cargo.toml
```diff
@@ -4,7 +4,8 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-ethers = { version = "2.0", features = ["ws", "rustls"] }
+# ethers = { version = "2.0", features = ["ws", "rustls"] }
```
taturosati: Remove comment
```suggestion
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -189,12 +200,28 @@ impl Batcher {
         // FIXME: We are not doing anything for the moment with the address from the
         // sender, this logic should be added for the payment system.
```
taturosati, line 201:
```suggestion
```

Remove Old comment

batcher/aligned-batcher/src/lib.rs
```diff
@@ -189,12 +200,28 @@ impl Batcher {
...
+                .call()
+                .await
+                .unwrap_or_default();
+
+            if user_balance == U256::from(0) {
+                error!("Insufficient funds for address {:?}", addr);
+                return Err(tokio_tungstenite::tungstenite::Error::Protocol(
+                    ProtocolError::HandshakeIncomplete,
+                ));
+            }
```
taturosati, line 217: Should probably check against a fixed value > 0

I think in contract it is 0.0004 eth or something like that

batcher/aligned-batcher/src/lib.rs
```diff
@@ -381,9 +425,19 @@ impl Batcher {
         info!("Batch sent to S3 with name: {}", file_name);
 
         info!("Uploading batch to contract");
-        let service_manager = &self.service_manager;
+        // let service_manager = &self.service_manager;
```
taturosati: Remove comment
```suggestion
```

batcher/aligned-batcher/src/lib.rs
uri-99: +1

uri-99: this was added in issue #521 , to solve all of these cases together

uri-99:
is this error supposed to happen? I think only the first one should appear:

image

MauroToscano (approved):

PR #536, opened 2024-07-01
perf: reorder batchstate struct

### LINKED ISSUE #517: perf: reorder + delete AlignedLayerServiceManagerStorage BatchState struct
Currently 
```
struct BatchState {
    uint32 taskCreatedBlock;
    bool responded;
    address batcherAddress;
}
```

should be 
```
struct BatchState {
    address batcherAddress;
    uint32 taskCreatedBlock;
    bool responded;
}
```

Also, `batcherAddress` should be deleted after its use, at the end of respondToTask.

This reduces gas consumption but breaks the past storage of this contract. This is not a problem since this is why we use testnet. Also, the explorer will be unaffected by this.

uri-99:
> [!Warning]
> This PR will make previous batches unreadeable because it restructures the contract variables.

# To test:
Everything should work normally, so start:
- anvil, aggregator, operator, batcher, explorer
- send proofs
- they should


uri-99:
No gas savings after this refactor

PR #609, opened 2024-07-08
feat: show proof hashes in each batch

uri-99:
## Notes

* ~~Merge after #580~~

## Changes

* add new table ecto migrate for proof data
* calculate keccak of proof in explorer to get proof data

### Before

image

### After

https://github.com/user-attachments/assets/0e6dbedb-d1a1-48a2-b4b2-88101affed50

PR #617, opened 2024-07-10
feat(explorer): add cost per proof of batch

### LINKED ISSUE #598: feat(explorer): add cost per proof of batch

glpecile:
# Changes

* ~~merge after #609~~
* dynamically grab gas per proof `batcher-payment-service` config json
* display cost = gas * gas price per proof in:
  * ETH ✅ 
  * USD ✅ 
* USD prices are now tied to the periodical process that runs and updates views, the value is then sent via the pubsub connection and updated in the batch view
* A global cache has been added that holds for 5 minutes the latest ETH to USD value from the coingecko API

## Before

image

## After

image


explorer/lib/explorer_web/live/eth_converter.ex
NicolasRampoldi, line 1: Maybe we can cache the price with the following library: https://hexdocs.pm/cachex/Cachex.html if it is easy to implement

NicolasRampoldi (approved):

explorer/lib/explorer/periodically.ex
```diff
@@ -12,36 +12,38 @@ defmodule Explorer.Periodically do
...
-      run_every_n_iterations = 8
-      new_count = rem(count + 1, run_every_n_iterations)
-      if new_count == 0 do
-        Task.start(&process_unverified_batches/0)
-      end
+    run_every_n_iterations = 8
+    new_count = rem(count + 1, run_every_n_iterations)
+    # if new_count == 0 do
+    Task.start(&process_unverified_batches/0)
+    # end
```
entropidelic: should this condition be checked? if not, let's remove the commented code

uri-99: yes, good catch

entropidelic (approved):

PR #626, opened 2024-07-12
fix: avs subscriber error not being handled correctly

### LINKED ISSUE #625: bug: avs subscriber error not being handled correctly
```
func (s *AvsSubscriber) SubscribeToNewTasks(newTaskCreatedChan chan *servicemanager.ContractAlignedLayerServiceManagerNewBatch) event.Subscription {
	sub, err := s.AvsContractBindings.ServiceManager.WatchNewBatch(
		&bind.WatchOpts{}, newTaskCreatedChan, nil,
	)
	if err != nil {
		s.logger.Error("Failed to subscribe to new AlignedLayer tasks", "err", err)
	}
	s.logger.Infof("Subscribed to new AlignedLayer tasks")
	return sub
}
```

When the subscription can't be established, an error is logged but the function continues and returns nil. This ends up crashing the operator or the aggregator.
We would need to retry the connection a few times and ultimately make the operator shut down gracefully

taturosati:
Added retires on subscribe.

Aggregator still crashes because of eigen sdk in memory subscription to pub keys.

# To test

First test normal flow, then try to kill anvil and operator should stay online and retrying (aggregator will still crash as mentioned above). Restart anvil and operator should resubscribe.

### DIFF
--- a/core/chainio/avs_subscriber.go
+++ b/core/chainio/avs_subscriber.go
@@ -12,8 +12,10 @@ import (
 	sdklogging "github.com/Layr-Labs/eigensdk-go/logging"
 )
 
-const MAX_RETRIES = 5
-const RETRY_INTERVAL = 5
+const (
+	MaxRetries    = 100
+	RetryInterval = 1 * time.Second
+)
 
 // NOTE(marian): Leaving this commented code here as it may be useful in the short term.
 // type AvsSubscriberer interface {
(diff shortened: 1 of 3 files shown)


core/chainio/avs_subscriber.go
```diff
@@ -42,15 +50,22 @@ func NewAvsSubscriberFromConfig(baseConfig *config.BaseConfig) (*AvsSubscriber,
...
-	)
-	if err != nil {
-		s.logger.Error("Failed to subscribe to new AlignedLayer tasks", "err", err)
+func (s *AvsSubscriber) SubscribeToNewTasks(newTaskCreatedChan chan *servicemanager.ContractAlignedLayerServiceManagerNewBatch) (event.Subscription, error) {
+	for i := 0; i < MaxRetries; i++ {
+		sub, err := s.AvsContractBindings.ServiceManager.WatchNewBatch(
+			&bind.WatchOpts{}, newTaskCreatedChan, nil,
+		)
+		if err != nil {
+			s.logger.Info("Failed to subscribe to new AlignedLayer tasks", "err", err)
```
entropidelic: lets make it a warn instead of info

entropidelic (approved):

NicolasRampoldi (approved):

PR #628, opened 2024-07-12
feat(deprecated): add requirement of user signature in batcherpaymentservice

### LINKED ISSUE #535: feat: add requirement of user signature in BatcherPaymentService
User must sign merkle root + how many proofs he submitted. Also Batch must have an id or nonce to avoid users signing for 2 valid batches at once.
PaymentService smart contract must verify this signature, to stop batcher to invent user payments

uri-99:
This is a first exploration of batcherPaymentService implementation. This PR should probably not be merged, instead, #629 should be merged.

This PR is temporarily left as draft as a helper.


uri-99:
#629 implemented

PR #650, opened 2024-07-18
chore: bump `eigensdk-go` version

entropidelic:
Bumps eigensdk-go from `v0.1.6` to `v0.1.9`

The main changes were some APIs interfaces changes and structs names.

1. For instantiating a new BLS aggregation service, an additional parameter has to be passed, `hashFunction`, and the BLS aggregation service method `ProcessNewSignature` now receives `TaskResponse` (which is just a go `interface{}`, so it can be anything) as an argument instead of `TaskResponseDigest`, which was done before. The v0.1.9 version can be seen [here](https://github.com/Layr-Labs/eigensdk-go/blob/4b1bb2a86e2acc0de02952b5a20de1aa5eaaf746/services/bls_aggregation/blsagg.go#L191) and the v0.1.6 [here](https://github.com/Layr-Labs/eigensdk-go/blob/cff810715271da986a7e594f7967a86fc4299834/services/bls_aggregation/blsagg.go#L174). What we were sending in `v0.1.6`as the `TaskResponseDigest` was really just the batch merkle root, so to fullfil the API requirements, a dummy hash function that outputs its input is being passed as the `hashFunction` argument. This should not affect in anything and will just replicate the behavior we were having in v0.1.6. 
2. The `NewOperatorsInfoServiceInMemory` function from the `operatorsinfo` module now receives a parameter `logFilterQueryBlockRange`, and if set to `nil`, a default value for that parameter is set. Check [this](https://github.com/Layr-Labs/eigensdk-go/blob/4b1bb2a86e2acc0de02952b5a20de1aa5eaaf746/services/operatorsinfo/operatorsinfo_inmemory.go#L68) lines and [this](https://github.com/Layr-Labs/eigensdk-go/blob/4b1bb2a86e2acc0de02952b5a20de1aa5eaaf746/chainio/clients/avsregistry/reader.go#L24) comment.
3. Some structs such as `AvsRegistryReader` and `AvsRegistryWriter`changed their names to `ChainReader` and `ChainWriter`.

### Testing
To test the PR, just test normal flow, ideally with more than one operator


NicolasRampoldi (approved):
Tested it with 2 operators and worked as expected also code looks good

uri-99 (approved):
worked smoothly for me with 2 operators 👍

PR #655, opened 2024-07-22
Update 0_about_aligned.md

cdesaintguilhem:
Add missing word on line 28.


MauroToscano:
We don't accept PRs that change the readme from non contributors. You can create an issue if you see a problem

PR #669, opened 2024-07-25
fix: nonce handling

### LINKED ISSUE #667: fix: nonces
Fix nonce error handling. Batcher never recovers from reverted tx. Proposed solution is this:

image

taturosati:
# Implementation

Implement nonce cache clearing on fail, both on client and on batcher

image

# To test

Test normal flow, make some txs revert by sending same burst twice


batcher/aligned-batcher/src/lib.rs
```diff
@@ -496,6 +496,20 @@ impl Batcher {
         Ok(())
     }
 
+    async fn reset_state(&self) {
```
```suggestion
    async fn flush_queue_and_clear_nonce_cache(&self) {
```

taturosati: changed on 7700d8197cd52aacb695bb885f428f741532665a

uri-99 (changes requested):
If a tx reverts because of invalid nonce, batcher will try to send the same tx to blockchain ~50 times before it stops trying. It should try only once or twice, because no variable is changing between retries.

The following error appeared every times in the batcher when sending concurrent proofs
`ERROR aligned_batcher] Error while sending batch inclusion data response: IO error: Broken pipe (os error 32)`

Batcher gets the following error when trying to send second message to client
`Unexpected error: IO error: Connection reset by peer (os error 54)`

MauroToscano (approved):

uri-99 (approved):
I understand my previous comments should be addressed in another PR

PR #679, opened 2024-07-29
refactor (batcher): put both the nonces and the queue together in a lock

### LINKED ISSUE #678: refactor (batcher): put both the nonces and the queue together in a lock
# Description

- We should put the `batch_queue`, the `user_nonces` and the `user_proof_count_in_batch` together in the same lock.

NicolasRampoldi:
# Description

- This PR creates a BatchState struct that holds the `batch_queue`, the `user_nonces` and the `user_proof_count_in_batch` and unifies the lock.

# To Test

- Run everything as usual and make sure it is working correctly.


batcher/aligned-batcher/src/lib.rs
```diff
@@ -285,25 +295,30 @@ impl Batcher {
...
+        let mut batch_state = self.batch_state.lock().await;
 
-        let mut user_proof_counts = self.user_proof_count_in_batch.lock().await;
-        let user_proofs_in_batch = *user_proof_counts.get(addr).unwrap_or(&0) + 1;
-
+        let user_proofs_in_batch = batch_state
+            .user_proof_count_in_batch
+            .get(addr)
+            .unwrap_or(&0)
+            + 1;
```
entropidelic: let's define a method on `BatchState` with this functionality

batcher/aligned-batcher/src/lib.rs
```diff
@@ -285,25 +295,30 @@ impl Batcher {
...
 
         let min_balance = U256::from(user_proofs_in_batch) * U256::from(MIN_BALANCE_PER_PROOF);
         if user_balance < min_balance {
             return false;
         }
 
-        user_proof_counts.insert(*addr, user_proofs_in_batch);
+        batch_state
+            .user_proof_count_in_batch
+            .insert(*addr, user_proofs_in_batch);
```
entropidelic: let's define a method on `BatchState` with this functionality too. Also, let's do this update outside `check_user_balance`, after ir returns `true`

batcher/aligned-batcher/src/lib.rs
NicolasRampoldi: Solved in [2f7ea9d](https://github.com/yetanotherco/aligned_layer/pull/679/commits/2f7ea9dd1cffb4a63d3286eb83b7c96f204daca0)

batcher/aligned-batcher/src/lib.rs
```diff
@@ -51,21 +51,43 @@ const ADDITIONAL_SUBMISSION_COST_PER_PROOF: u128 = 13_000;
...
+        Self {
+            batch_queue: BatchQueue::new(),
+            user_nonces: HashMap::new(),
+            user_proof_count_in_batch: HashMap::new(),
+        }
+    }
+
+    fn increment_user_proof_count(&mut self, addr: &Address) -> u64 {
+        self.user_proof_count_in_batch.get(addr).unwrap_or(&0) + 1
+    }
```
entropidelic: the `increment` in the name could lead to confusion since this does not update state. Let's call it something like `get_user_proof_count`, and then sum one where needed

NicolasRampoldi: Solved in [913a930](https://github.com/yetanotherco/aligned_layer/pull/679/commits/913a930bff8caa6aa635af1c054ffb66fc672209)

batcher/aligned-batcher/src/lib.rs
```diff
@@ -51,21 +51,43 @@ const ADDITIONAL_SUBMISSION_COST_PER_PROOF: u128 = 13_000;
...
+
+    fn update_user_proof_count(&mut self, addr: &Address, count: u64) {
+        self.user_proof_count_in_batch.insert(*addr, count);
+    }
+}
```
entropidelic: we always update by 1, right? if so, this function could actually be `increment_user_proof_count`. Check the `entry()` [API for hashmaps](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry).
This method should be called after `check_user_balance` returns true, or the function should be renamed to `check_user_balance_and_increment_proof_count`

NicolasRampoldi: Solved in [5e72f95](https://github.com/yetanotherco/aligned_layer/pull/679/commits/5e72f951b9886bef437e6f333b5c8db2b8613ec8)

batcher/aligned-batcher/src/lib.rs
NicolasRampoldi: Solved in [5e72f95](https://github.com/yetanotherco/aligned_layer/pull/679/commits/5e72f951b9886bef437e6f333b5c8db2b8613ec8)

entropidelic (approved):

taturosati (approved):

PR #687, opened 2024-07-30
hotfix: handle nil prices per proofs

### LINKED ISSUE #686: hotfix(explorer): empty blocks without price

uri-99:


glpecile (approved):
We'll test this in stage but it works locally 👍 

edit: it works!

JuArce (approved):

PR #699, opened 2024-08-01
fix: dont upload nonce to bucket

### LINKED ISSUE #698: fix: don't upload nonce to storage
Revert changes. Upload verification data instead of nonced verification data

taturosati:
Revert change. Bucket now holds VerificationData[] instead of NoncedVerificationData[]

# To test

Test normal flow, try different proof types


uri-99 (changes requested):
This PR changes the s3 object structure, so a change in how the explorer reads this object is missing

uri-99 (approved):

MauroToscano (approved):

PR #703, opened 2024-08-01
feat: add polling to the operator

### LINKED ISSUE #700: feat: add polling to the operator
# Description

- The operator should check every 5s which is the last task in the blockchain.

NicolasRampoldi:
> [!NOTE]  
> This was already tested on stage.

# Description

- This PR makes the operator get the latest task from the AlignedLayerServiceManager contract every X seconds, if the task was not yet responded, it starts processing it.

# To Test

- Test normal flow, make sure everything runs smoothly.
- Test just the polling, for this, modify `core/chainio/avs_subscriber.go` adding the following after line 97:
```go
          case newBatch := <-internalChannel:
                  continue
	          s.processNewBatch(newBatch, batchesSet, newBatchMutex, newTaskCreatedChan)
```
This will ignore the new batches received from the channel with the binding and just receive the ones from the polling.


operator/pkg/operator.go
```diff
@@ -164,8 +183,19 @@ func (o *Operator) Start(ctx context.Context) error {
 				OperatorId:      o.OperatorId,
 			}
 
+			o.processedBatches[newBatchLog.BatchMerkleRoot] = struct{}{}
+			o.latestProcessedBatch = newBatchLog
+
+			o.processedBatchesMutex.Unlock()
```
taturosati: this should be on top, no need to lock for whole verification process

operator/pkg/operator.go
```diff
@@ -210,6 +240,117 @@ func (o *Operator) ProcessNewBatchLog(newBatchLog *servicemanager.ContractAligne
...
+	latestBlock, err := o.Config.BaseConfig.EthRpcClient.BlockNumber(context.Background())
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to get latest block number: %w", err)
+	}
+
+	var fromBlock uint64
+
+	if latestBlock < blockInterval {
+		fromBlock = 0
```
taturosati: I don't think this is correct, why check current block number < 50_000

operator/pkg/operator.go
NicolasRampoldi: What if the o.ProcessNewBatchLog(newBatchLog) throws an error? Wouldn't you be updating the latestProcessedBatch even though it failed?

taturosati (approved):

core/chainio/avs_subscriber.go
```diff
@@ -147,6 +154,90 @@ func subscribeToNewTasks(
...
+			delete(batchesSet, batch.BatchMerkleRoot)
+			newBatchMutex.Unlock()
+		}()
+	}
+}
+
+// getLatestTaskFromBlockchain queries the blockchain for the latest task using the FilterLogs method.
+// The alternative to this is using the FilterNewBatch method from the contract's filterer, but it requires
+// to iterate over all the logs, which is not efficient and not needed since we only need the latest task.
+func (s *AvsSubscriber) getLatestTaskFromBlockchain() (*servicemanager.ContractAlignedLayerServiceManagerNewBatch, error) {
```
entropidelic:
```suggestion
func (s *AvsSubscriber) getLatestTaskFromEthereum() (*servicemanager.ContractAlignedLayerServiceManagerNewBatch, error) {
```

entropidelic:
lgtm, just suggested a different name for one of the new functions

core/chainio/avs_subscriber.go
NicolasRampoldi: Solved in [2312232](https://github.com/yetanotherco/aligned_layer/pull/703/commits/2312232befa6ac0f87ff49b614d1cebd3fb0009d)

entropidelic (approved):

PR #721, opened 2024-08-05
fix: operator dockerfile

### LINKED ISSUE #720: fix: operator dockerfile
# Description

- The operator dockerfile is outdated and needs some fixes.

NicolasRampoldi:
# Description

- Merges install build dependencies commands for better readability.
- Copies the `aligned-sdk` into docker for the ffis to compile correctly.
- Removes the download of go dependencies because it throws an error of conflicts. The other option was to modify the `go.mod` file, but this seems better. The only reason the `go mod download` was being done was to improve docker run times but  just running `go build` works correctly.
- Adds `OPERATOR_VERSION` to the docker build.

# To Test

- Set the `ECDSA_KEY_FILE_HOST`, `BLS_KEY_FILE_HOST` and `OPERATOR_CONFIG_FILE_HOST` variables in the `operator/docker/.env` file.
- Run `make operator_start_docker`.


operator/docker/operator.Dockerfile
```diff
@@ -1,45 +1,41 @@
...
 COPY common /usr/src/app/common
 
-# Download dependencies
-RUN go mod download && go mod tidy && go mod verify
+# Define the OPERATOR_VERSION variable
+ARG OPERATOR_VERSION=v0.4.0
 
 # Build the operator
-RUN go build -v -o /usr/local/bin/operator /usr/src/app/operator/cmd/main.go
+RUN go build -ldflags "-X main.Version=${OPERATOR_VERSION}" -v -o /usr/local/bin/operator /usr/src/app/operator/cmd/main.go
```
entropidelic: could we just run `make build_operator` so that the `OPERATOR_VERSION` variable doesn't have to be defined here too?

NicolasRampoldi: You are right. I removed the `ARG OPERATOR_VERSION=v0.4.0` in this commit: [ca3060f](https://github.com/yetanotherco/aligned_layer/pull/721/commits/ca3060fc78dca50bf04b2d615688dffa0b83eebb) because it didn't make sense. Just declared the argument.

entropidelic (approved):

NicolasRampoldi:
Docker will not be supported for the time being.

PR #761, opened 2024-08-09
feat (batcher): fallback rpc

### LINKED ISSUE #757: feat (batcher): fallback rpc
Add fallback rpc for batcher

taturosati:
Add fallback rpc for batcher calls.

# to test

Test normal flow

> Works on stage


batcher/aligned-batcher/src/eth/mod.rs
```diff
@@ -61,3 +63,36 @@ pub async fn get_batcher_payment_service(
...
+        .send()
+        .await
+        .map_err(|e| BatcherError::TaskCreationError(e.to_string()))?;
+
+    let receipt = pending_tx
+        .await
+        .map_err(|_| BatcherError::TransactionSendError)?
+        .ok_or(BatcherError::ReceiptNotFoundError)?;
+
+    Ok(receipt)
```
entropidelic: small nit
```suggestion
pending_tx
    .await
    .map_err(|_| BatcherError::TransactionSendError)?
    .ok_or(BatcherError::ReceiptNotFoundError)
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -675,28 +728,31 @@ impl Batcher {
...
-            .map_err(|_| BatcherError::TransactionSendError)?
-            .ok_or(BatcherError::ReceiptNotFoundError)?;
-
-        Ok(receipt)
+            &self.payment_service,
+        )
+        .await
+        {
+            Ok(receipt) => Ok(receipt),
+            Err(_) => {
```
entropidelic, line 752: I think we should check the reason of the error. For example, if the transaction was reverted it wouldn't make sense to retry with the fallback and we will be paying two times. We should try with the fallback only when the error is related to RPC connection/reliability issues

batcher/aligned/.env.bk
```diff
@@ -0,0 +1 @@
+PRIVATE_KEY=0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d
```
entropidelic: is it ok to commit this?

taturosati: Its an anvil private key, removed it anyways cause its not relevant for PR

entropidelic (approved):

PatStiles (approved):

PR #766, opened 2024-08-09
feat(explorer): add search by proof hash

### LINKED ISSUE #765: feat(explorer): add search by proof hash

glpecile:
# Changes

* add method to proofs API that returns the merkle root / array of merkle roots of the batch(es) that contain(s) the proof hash hex
* add new search results view that redirects when one there's one result only
* edit search component placeholder text to include batch search
* add copy to clipboard icon next to proof hashes

## Demo (with pagination)

https://github.com/user-attachments/assets/3cc150cb-fb09-455f-87cf-c5f3d43f088b


explorer/lib/explorer/models/proofs.ex
```diff
@@ -43,4 +43,23 @@ defmodule Proofs do
...
+
+    {:ok, proof_hash_binary} = Base.decode16(proof_hash_hex, case: :mixed)
+
+    query = from(p in Proofs,
+      where: p.proof_hash == ^proof_hash_binary,
+      order_by: [desc: p.id],
+      limit: 1,
+      select: p)
+
+    case Explorer.Repo.one(query) do
```
uri-99: This gets only the latest entry, but we can have a proof present in many batches

uri-99 (changes requested):
This PR is missing a view showing all batches where a proof is present, if the proof is present in more than one batch

entropidelic (approved):

explorer/lib/explorer/models/proofs.ex
```diff
@@ -42,4 +42,29 @@ defmodule Proofs do
     end
   end
 
+  def get_batch_from_proof(proof_hash_hex) do
```
uri-99: if it returns N, function name should be plural
```suggestion
  def get_batches_containing_proof(proof_hash_hex) do
```

explorer/lib/explorer_web/live/pages/search/index.ex
```diff
@@ -0,0 +1,48 @@
+defmodule ExplorerWeb.Search.Index do
+  use ExplorerWeb, :live_view
+
+  def mount(%{"hash" => hash}, _session, socket) do
+    case Proofs.get_batch_from_proof(hash) do
```
uri-99:
```suggestion
    case Proofs.get_batches_containing_proof(hash) do
```

uri-99 (changes requested):

NicolasRampoldi (commented):
I think we should add pagination, other than that everything seems ok

uri-99 (approved):

NicolasRampoldi (approved):

PR #769, opened 2024-08-12
fix (operator): audits on getBatchFromS3

### LINKED ISSUE #768: fix (operator): audits on get batch from s3
- Fix oom attack with tars
- Fix bypass length

taturosati:
- Fix oom error with gzip bomb
- Fix bypass max batch size

Both were fixed with io.LimitReader

# To test

Follow audit instructions for attack cases, check everything still works as normal

### DIFF
--- a/operator/pkg/s3.go
+++ b/operator/pkg/s3.go
@@ -21,9 +21,10 @@ func (o *Operator) getBatchFromS3(batchURL string, expectedMerkleRoot [32]byte) 
 		return nil, fmt.Errorf("error getting Proof Head from S3: %s", resp.Status)
 	}
 
-	if resp.ContentLength > o.Config.Operator.MaxBatchSize {
+	contentLength := resp.ContentLength
+	if contentLength > o.Config.Operator.MaxBatchSize {
 		return nil, fmt.Errorf("proof size %d exceeds max batch size %d",
-			resp.ContentLength, o.Config.Operator.MaxBatchSize)
+			contentLength, o.Config.Operator.MaxBatchSize)
 	}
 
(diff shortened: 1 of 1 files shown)


NicolasRampoldi (approved):

operator/pkg/s3.go
```diff
@@ -37,11 +38,20 @@ func (o *Operator) getBatchFromS3(batchURL string, expectedMerkleRoot [32]byte)
...
+	// This is to prevent the operator from downloading a larger than expected file
+	// + 1 is added to the contentLength to check if the response body is larger than expected
+	reader := io.LimitedReader{R: resp.Body, N: contentLength + 1}
+	batchBytes, err := io.ReadAll(&reader)
 	if err != nil {
 		return nil, err
 	}
 
+	// Check if the response body is larger than expected
+	if reader.N == 0 {
```
entropidelic: in the [io.LimitedReader docs](https://pkg.go.dev/io#LimitedReader) I found
> Read returns EOF when N <= 0 or when the underlying R returns EOF. 

should the condition be changed to 
```suggestion
	if reader.N <= 0 {
```
?

entropidelic:
aside from one comment regarding the `LimitedReader` documentation on its parameter N, code looks good and the attacks were handled succesfully with this solution

operator/pkg/s3.go
MauroToscano: Seems like it should be changed, yes

MauroToscano:
Is there a zip bomb to test ? We could have it around to try it

entropidelic (approved):

PR #804, opened 2024-08-19
fix: frontrunning createNewTask on servicemanager

### LINKED ISSUE #789: Frontrunning createNewTask() on AlignedLayerServiceManager allows to DOS legitimate batch submissions
Frontrunning createNewTask() on AlignedLayerServiceManager allows to DOS legitimate batch submissions

- [ ] Change batch map to index tasks by batcher_address (caller of createTask) and task hash

uri-99:
## Solution developed

To solve issue #789 , and prevent frontrunning DoS, the folowing solution was implemented:

AlignedLayerServiceManager's map to store batches state (previously `merkle_root -> batchState`), now uses msg.sender as part of its key (now `hash(merkle_root, msg.sender) -> batchState`).

This solves the issue mentioned because if another User tries to frontrun a batch with its merkle_root, it won't be classified as the same batch by the AlignedLayerServiceManager, as msg.sender will be different (thus, hash(merkle_root, msg.sender` will be different).

One may think "but what if the attacker uses the same BatcherPaymentService.sol contract to make the frontrun? Won't both submittions have the same msg.sender?" ; But this is not a problem, since the BatcherPaymentService only allows one address (the batcher) to make submitions to it.

### ServiceManager

- RespondToTask has new parameter: `address senderAddress`:

```
bytes32 batchMerkleRoot,
address senderAddress,
```

- NewBatch event has new parameter: `address senderAddress`:

```
event NewBatch(
    bytes32 indexed batchMerkleRoot,
    address senderAddress,
    uint32 taskCreatedBlock,
    string batchDataPointer
);
```

- BatchVerified event has new parameter: `address senderAddress`:

```
event BatchVerified(
    bytes32 indexed batchMerkleRoot,
    address senderAddress
);
```

### Explorer

- Reading correctly the NewBatch event
  - storing it in new sender_address column of batches table
- Reading correctly the BatchVerified event

### Operator

- Is reading NewBatch's new paramenter
  - calculates BatchIdentifierHash = hash(merkleRoot, senderAddress)
  - no longer use merkleRoot to identify batches, now using BatchIdentifierHash
  - submitting response to Aggregator with signature of (merkleRoot, senderAddress)

### Aggregator

- Recieving signature of (merkleRoot, senderAddress) from Operator
- Making correct new respondToTask 

### Batcher

- No changes needed, can still submit batches


operator/pkg/operator.go
```diff
@@ -155,15 +155,21 @@ func (o *Operator) Start(ctx context.Context) error {
...
 			signedTaskResponse := types.SignedTaskResponse{
+				BatchIdentifierHash: batchIdentifierHash,
 				BatchMerkleRoot: newBatchLog.BatchMerkleRoot,
+				SenderAddress:   newBatchLog.SenderAddress,
 				BlsSignature:    *responseSignature,
 				OperatorId:      o.OperatorId,
 			}
 
-			o.Logger.Infof("Signed hash: %+v", *responseSignature)
+			o.Logger.Infof("Signed Task Response to send: %+v", signedTaskResponse)
```
taturosati: image

We should hex encode relevant fields here, its hard to read

operator/pkg/operator.go
```diff
@@ -155,15 +155,21 @@ func (o *Operator) Start(ctx context.Context) error {
 				o.Logger.Infof("batch %x did not verify. Err: %v", newBatchLog.BatchMerkleRoot, err)
 				continue
 			}
-			responseSignature := o.SignTaskResponse(newBatchLog.BatchMerkleRoot)
+			batchIdentifier := append(newBatchLog.BatchMerkleRoot[:], newBatchLog.SenderAddress[:]...)
+			var batchIdentifierHash = *(*[32]byte)(crypto.Keccak256(batchIdentifier))
+
+			responseSignature := o.SignTaskResponse(batchIdentifierHash)
+			o.Logger.Infof("responseSignature about to send: %x", responseSignature)
```
taturosati: image
 
 I don't know if this is very useful, maybe keep it but in debug 
```suggestion
			o.Logger.Debugf("responseSignature about to send: %x", responseSignature)
```

taturosati (changes requested):
small nitpicks with logs, other than that code looks good and worked

MauroToscano (approved):

entropidelic:
lgtm, I agree on tatu's comments but aside from that everything looks good.

entropidelic (approved):

taturosati (approved):

PR #806, opened 2024-08-19
feat: operators version service

taturosati:
Add operator version tracker

Exposes two endpoints:

- POST to /versions for operator to post its version on startup, verifies operator signature
- GET to /versions to list operator versions

# To test

Start anvil and aggregator

Start operator tracker: `make tracker_devnet_start`

Register operator: `make operator_full_registration`
Build operator: `make build_operator`
Start operator: `./operator/build/aligned-operator start --config ./config-files/config.yaml`

> [!WARNING]  
> Note that you can't use `make operator_start` because operator version is not included in executable if not built.

You can then get operator versions by running `curl http://localhost:3000/versions`

You can test an update by changing OPERATOR_VERSION in makefile, rebuilding and restarting operator


Makefile
```diff
@@ -664,3 +664,33 @@ recover_db: run_db
 explorer_fetch_old_batches:
 	@cd explorer && \
 	./scripts/fetch_old_batches.sh 1728056 1729806
+
+__TRACKER__:
+
+tracker_devnet_start: tracker_run_db
+	@cd operator_tracker/ && \
+		cp .env.dev .env && \
```
JuArce: would be nice to run this target using the .env.dev file instead of overwriting the .env file

config-files/config.yaml
```diff
@@ -39,6 +39,7 @@ aggregator:
 ## Operator Configurations
 operator:
   aggregator_rpc_server_ip_port_address: localhost:8090
+  operator_tracker_ip_port_address: http://localhost:3000
```
JuArce: This variable `operator_tracker_ip_port_address: http://localhost:3000` is missing in
- config-operator.yaml
- config-operator-1.yaml
- config-operator-2.yaml
- config-operator-3.yaml

operator_tracker/sql/schema.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE IF NOT EXISTS operator_versions (
+  address CHAR(42) PRIMARY KEY,
+  version TEXT
```
JuArce: version should have a fixed length to avoid accept too long strings

operator_tracker/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+        registry_coordinator,
+    };
+
+    let listener = tokio::net::TcpListener::bind("0.0.0.0:3000")
+        .await
+        .expect("Failed to bind listener");
+
+    let router = Router::new()
+        .route("/versions", post(post_operator_version))
+        .route("/versions", get(list_operator_versions))
```
JuArce, line 94: There is missing a `/versions/:address`. Mainly for operator view in explorer to avoid request all versions every time

operator_tracker/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+    Ok(())
+}
+
+async fn list_operator_versions(state: State<AppState>) -> Json<Vec<OperatorVersion>> {
+    let rows = sqlx::query_as::<_, OperatorVersion>("SELECT * FROM operator_versions")
+        .fetch_all(&state.pool)
+        .await
+        .expect("Failed to execute query");
+
+    Json(rows)
```
JuArce: Would be nice return 204 if there is no body content

operator_tracker/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+        axum::http::Response::builder()
+            .status(axum::http::StatusCode::OK)
+            .body(axum::body::Body::from(
+                serde_json::to_string(&self).unwrap(),
+            ))
+            .unwrap()
+    }
+}
+
+async fn post_operator_version(
```
JuArce, line 103: This should return 201 if it is created correctly. Also should include a link to the entity in the `Location` header and the created entity as response body

JuArce (changes requested):

operator/cmd/actions/start.go
```diff
@@ -34,6 +39,42 @@ func operatorMain(ctx *cli.Context) error {
...
+	}
+	bodyBuffer := new(bytes.Buffer)
+
+	bodyReader := json.NewEncoder(bodyBuffer)
+	err = bodyReader.Encode(body)
+	if err != nil {
+		return err
+	}
+
+	// send version to metrics server
```
entropidelic: small nit but this comment confused me 😛. It should be the operator tracker server

operator/cmd/actions/start.go
```diff
@@ -34,6 +39,42 @@ func operatorMain(ctx *cli.Context) error {
...
+	endpoint := operatorConfig.Operator.OperatorTrackerIpPortAddress + "/versions"
+	log.Println("Sending version to metrics server: ", endpoint)
```
entropidelic: same here. Also, can we use the eigensdk logger?

operator_tracker/src/main.rs
```diff
@@ -0,0 +1,154 @@
...
+use operator_tracker::RegistryCoordinatorContract;
+use sqlx::postgres::{PgPool, PgPoolOptions};
+use sqlx::Executor;
+
+// The RegistryCoordinator contract
+// This is the contract that the service will interact with to get the status of the operator
+// The response is a uint8 that represents the status of the operator
+// 0 - Never registered
+// 1 - Registered
+// 2 - Deregistered
```
entropidelic, line 27: I don't really know how ethers handles this but couldn't we just use the contract enum?

taturosati: unfortunately I don't think so, at least I could not find a way

operator_tracker/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    InvalidSignature,
+    OperatorNotRegistered,
+}
+
+impl IntoResponse for OperatorVersionError {
+    fn into_response(self) -> axum::http::Response<Body> {
+        axum::http::Response::builder()
+            .status(axum::http::StatusCode::BAD_REQUEST)
+            .body(Body::from(serde_json::to_string(&self).unwrap()))
+            .unwrap()
```
entropidelic: let's not use unwraps, if there is any error while building the response body let's respond something

operator_tracker/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    address: String,
+    pub version: String,
+}
+
+impl IntoResponse for OperatorVersion {
+    fn into_response(self) -> axum::http::Response<Body> {
+        axum::http::Response::builder()
+            .status(axum::http::StatusCode::OK)
+            .body(Body::from(serde_json::to_string(&self).unwrap()))
+            .unwrap()
```
entropidelic: same comment about responses here

operator_tracker/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+        .recover(version)
+        .map_err(|_| OperatorVersionError::InvalidSignature)?;
+
+    info!("Operator address: {:?}", operator_address);
+
+    // Check if operator is registered on registry coordinator
+    let status = registry_coordinator
+        .get_operator_status(operator_address)
+        .await
+        .expect("Failed to get operator status");
```
entropidelic: let's return an error here

operator_tracker/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+) -> Result<Option<OperatorVersion>, OperatorVersionError> {
+    debug!(
+        "Received operator version request. Version {}",
+        payload.version
+    );
+
+    // Recover operator address from signature
+
+    // hash keccak256(version) and recover address from signature
+    let version = keccak256(payload.version.as_bytes());
```
entropidelic:
```suggestion
    let hashed_version = keccak256(payload.version.as_bytes());
```

operator_tracker/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+            Some(OperatorVersion {
+                address: operator_address.clone(),
+                version: payload.version.clone(),
+            }),
+        )
+    };
+
+    db.execute(query.bind(&operator_address).bind(&payload.version))
+        .await
+        .expect("Failed to update operator version");
```
entropidelic: return error here

operator_tracker/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    // check operator address is hex and 42 characters long
+    if address.len() != 42 || !address.starts_with("0x") {
+        return None;
+    }
+
+    sqlx::query_as::<_, OperatorVersion>("SELECT * FROM operator_versions WHERE address = $1")
+        .bind(address)
+        .fetch_optional(db)
+        .await
+        .expect("Failed to execute query")
```
entropidelic: return error

operator_tracker/src/main.rs
```diff
@@ -0,0 +1,170 @@
...
+    {
+        Ok(Some(body)) => {
+            info!("Operator version created successfully");
+
+            let body = serde_json::to_string(&body).expect("Failed to serialize response");
+
+            axum::http::Response::builder()
+                .status(axum::http::StatusCode::CREATED)
+                .body(Body::from(body))
+                .expect("Failed to build response")
```
entropidelic: same comment about responses when something fails. Let's fix that in all the response builders

JuArce (commented):
Just one comment 
If the operator receives an error, it prints something like that

```
2024-08-21T12:14:58.066-0300	ERROR	actions/start.go:75	Error sending version to metrics server: 	{"err": null}
github.com/yetanotherco/aligned_layer/operator/cmd/actions.operatorMain
	/Users/juarce/repos/yetanothercompany/aligned_layer/operator/cmd/actions/start.go:75
github.com/urfave/cli/v2.(*Command).Run
	/Users/juarce/go/pkg/mod/github.com/urfave/cli/v2@v2.27.1/command.go:279
github.com/urfave/cli/v2.(*Command).Run
	/Users/juarce/go/pkg/mod/github.com/urfave/cli/v2@v2.27.1/command.go:272
github.com/urfave/cli/v2.(*App).RunContext
	/Users/juarce/go/pkg/mod/github.com/urfave/cli/v2@v2.27.1/app.go:337
github.com/urfave/cli/v2.(*App).Run
	/Users/juarce/go/pkg/mod/github.com/urfave/cli/v2@v2.27.1/app.go:311
main.main
	/Users/juarce/repos/yetanothercompany/aligned_layer/operator/cmd/main.go:27
runtime.main
	/opt/homebrew/Cellar/go/1.22.3/libexec/src/runtime/proc.go:271
```

Would be better to print a more useful message
It happens if the tracker service is not running or if the operator does not send a valid version (ie running without build)

taturosati:
> Just one comment If the operator receives an error, it prints something like that
> 
> ```
> 2024-08-21T12:14:58.066-0300	ERROR	actions/start.go:75	Error sending version to metrics server: 	{"err": null}
> github.com/yetanotherco/aligned_layer/operator/cmd/actions.operatorMain
> 	/Users/juarce/repos/yetanothercompany/aligned_layer/operator/cmd/actions/start.go:75
> github.com/urfave/cli/v2.(*Command).Run
> 	/Users/juarce/go/pkg/mod/github.com/urfave/cli/v2@v2.27.1/command.go:279
> github.com/urfave/cli/v2.(*Command).Run
> 	/Users/juarce/go/pkg/mod/github.com/urfave/cli/v2@v2.27.1/command.go:272
> github.com/urfave/cli/v2.(*App).RunContext
> 	/Users/juarce/go/pkg/mod/github.com/urfave/cli/v2@v2.27.1/app.go:337
> github.com/urfave/cli/v2.(*App).Run
> 	/Users/juarce/go/pkg/mod/github.com/urfave/cli/v2@v2.27.1/app.go:311
> main.main
> 	/Users/juarce/repos/yetanothercompany/aligned_layer/operator/cmd/main.go:27
> runtime.main
> 	/opt/homebrew/Cellar/go/1.22.3/libexec/src/runtime/proc.go:271
> ```
> 
> Would be better to print a more useful message It happens if the tracker service is not running or if the operator does not send a valid version (ie running without build)

Fixed, changed log to warn and removed nil error

JuArce (approved):

config-files/config.yaml
uri-99: We are already using port 3000 for metrics,
found in 6_setup_aligned.md:
```
To run Prometheus and Grafana, run:

`make run_metrics`

Then you can access Grafana on `http://localhost:3000` with the default credentials `admin:admin`.
```

One solution could be to change the port to another one, 
Or a better solution could be to make the operator_version_tracker port, variable.

uri-99 (changes requested):

operator_tracker/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    env_logger::init();
+
+    let database_url = &std::env::var("DATABASE_URL").expect("DATABASE_URL must be set");
+
+    let rpc_url = &std::env::var("RPC_URL").expect("RPC_URL must be set");
+
+    let registry_coordinator_addr = &std::env::var("REGISTRY_COORDINATOR_ADDRESS")
+        .expect("REGISTRY_COORDINATOR_ADDRESS must be set");
```
uri-99, line 61: We should consider reading this value from `alignedlayer_deployment_output.json`

operator_tracker/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+        serialize_or_err(rows)
+    };
+
+    axum::http::Response::builder()
+        .status(status)
+        .body(body)
+        .unwrap_or_default() // Should never fail but dont panic
+}
+
+async fn get_operator_version(
```
uri-99, line 161: There should be a special response if there are no operator versions found, to avoid the following:

image

uri-99 (changes requested):

Makefile
```diff
@@ -664,3 +664,32 @@ recover_db: run_db
...
+tracker_install: tracker_build_db
+	cargo install --path ./operator_tracker
+
+tracker_build_db:
+	@cd operator_tracker && \
+		docker build -t tracker-postgres-image .
+
+tracker_run_db: tracker_build_db tracker_remove_db_container
+	@cd operator_tracker && \
+		docker run -d --name tracker-postgres-container -p 5432:5432 -v tracker-postgres-data:/var/lib/postgresql/data tracker-postgres-image
```
uri-99: 5432 is also exposed for the Explorer DB, it will probably bring some issues on stage, but probably not on prod.

uri-99 (changes requested):

operator_tracker/src/main.rs
taturosati: I don't think its important, we would still need to pass path to the deployment output file, and this way we avoid the extra complexity of parsing the file

Makefile
uri-99: port was changed to 5433

operator_tracker/src/main.rs
uri-99: will return 204 -no response- code

config-files/config.yaml
uri-99: port changed to 3030

operator/cmd/actions/start.go
```diff
@@ -34,7 +39,45 @@ func operatorMain(ctx *cli.Context) error {
...
+	}
+
+	// send version to operator tracker server
+	endpoint := operatorConfig.Operator.OperatorTrackerIpPortAddress + "/versions"
+	operator.Logger.Info("Sending version to operator tracker server: ", "endpoint", endpoint)
+
+	res, err := http.Post(endpoint, "application/json",
+		bodyBuffer)
+	if err != nil {
+		// Dont prevent operator from starting if metrics server is down
```
entropidelic:
```suggestion
		// Dont prevent operator from starting if operator tracker server is down
```

operator/cmd/actions/start.go
```diff
@@ -34,7 +39,45 @@ func operatorMain(ctx *cli.Context) error {
...
+	endpoint := operatorConfig.Operator.OperatorTrackerIpPortAddress + "/versions"
+	operator.Logger.Info("Sending version to operator tracker server: ", "endpoint", endpoint)
+
+	res, err := http.Post(endpoint, "application/json",
+		bodyBuffer)
+	if err != nil {
+		// Dont prevent operator from starting if operator tracker server is down
+		operator.Logger.Warn("Error sending version to metrics server: ", "err", err)
+	} else if res.StatusCode != http.StatusCreated && res.StatusCode != http.StatusNoContent {
+		operator.Logger.Warn("Error sending version to metrics server: ", "status_code", res.StatusCode)
```
entropidelic:
```suggestion
		operator.Logger.Warn("Error sending version to operator tracker server: ", "status_code", res.StatusCode)
```

entropidelic (approved):

uri-99 (approved):

PR #812, opened 2024-08-20
feat: add context timeout to avoid dos

### LINKED ISSUE #797: Add timeout when downloading proofs in the Operators to avoid DoS
Add timeout when downloading proofs in the Operators to avoid DoS

NicolasRampoldi:
# Description

- This PR adds a timeout to the `getBatchFromS3` function in order to avoid DoS from a malicious server.

# To Test

- Make sure everything is working as usual.
- To test the timeout, in `operator.go` change the `BatchDownloadTimeout` constant to `BatchDownloadTimeout = 1 * time.Millisecond`. You should see an error like so:
```
ERROR (pkg/operator.go:189) Could not get proofs from S3 bucket: Get "http://localhost:4566/aligned.storage/c8eba46fe76b26356f6edb2c66badd4b94e1daa3871aed90ce73d3dcb0a2289b.json": context deadline exceeded
```
But the operator will continue to work normally.

### DIFF
--- a/operator/pkg/s3.go
+++ b/operator/pkg/s3.go
@@ -1,6 +1,7 @@
 package operator
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


entropidelic (approved):
lgtm, workped properly. Maybe we should think about a reasonable timeout interval

operator/pkg/s3.go
```diff
@@ -9,12 +10,25 @@ import (
 	"github.com/yetanotherco/aligned_layer/operator/merkle_tree"
 )
 
-func (o *Operator) getBatchFromS3(batchURL string, expectedMerkleRoot [32]byte) ([]VerificationData, error) {
+func (o *Operator) getBatchFromS3(ctx context.Context, batchURL string, expectedMerkleRoot [32]byte) ([]VerificationData, error) {
```
entropidelic, line 14: it would be really nice to change this function name sometime 😄

taturosati (approved):
Code looks good, worked for me

PR #820, opened 2024-08-20
feat: implement aligned layer service manager interface

### LINKED ISSUE #793: Batcher: Change contract call from ```.call(){value}``` to an interface
The batcher contract calls the service manager with a low level .call(){value}, which is unnecessary, and it could just call it through something like an interface. That would be easier to maintain in the codebase as well.

NicolasRampoldi:
> [!IMPORTANT]  
> We will need to upgrade the `AlignedLayerServiceManager` and `BatcherPaymentService` contracts. So we could first test it on stage.

# Description

- This PR refactors the `AlignedLayerServiceManager` contract to improve maintainability, code clarity, and adherence to best practices. The changes involve defining an interface for `AlignedLayerServiceManager` and replacing the low-level `.call` usage with interface calls in `BatcherPaymentService`.

# To Test

- Make sure everything works as usual.


contracts/src/core/IAlignedLayerServiceManager.sol
```diff
@@ -0,0 +1,29 @@
...
+        bytes32 pubInputCommitment,
+        bytes32 provingSystemAuxDataCommitment,
+        bytes20 proofGeneratorAddr,
+        bytes32 batchMerkleRoot,
+        bytes memory merkleProof,
+        uint256 verificationDataBatchIndex
+    ) external view returns (bool);
+
+    function balanceOf(address account) external view returns (uint256);
+}
```
uri-99, line 29: The following function is missing:

```suggestion
    function balanceOf(address account) external view returns (uint256);
    
    function checkPublicInput(
        bytes calldata publicInput,
        bytes32 hash
    ) public pure returns (bool);
}
```

uri-99 (changes requested):

contracts/src/core/IAlignedLayerServiceManager.sol
NicolasRampoldi: It's not there because the function is not `external`

contracts/src/core/BatcherPaymentService.sol
```diff
@@ -105,18 +109,11 @@ contract BatcherPaymentService is
...
-                batchMerkleRoot,
-                batchDataPointer
-            )
+        alignedLayerServiceManager.createNewTask{value: feeForAggregator}(
+            batchMerkleRoot,
+            batchDataPointer
         );
 
-        require(success, "createNewTask call failed");
-
```
uri-99, line 119: This is OK, call to interface will revert if failed 👍

taturosati (approved):

taturosati:
We should test on stage before merging

uri-99 (approved):

MauroToscano (approved):

PR #822, opened 2024-08-21
feat: add chain id to signature for batcher payments

### LINKED ISSUE #794: Append network ID to the data signed by user on Batcher Payments
To verify signed data, direct keccak256 hashes are checked. It would be better to utilize EIP-712 style hashing with a domain separator to prevent cross-network replay attacks. For example, a signature signed over on a testnet could work on mainnet.

taturosati:
Add chain id to batcher payments to fix issue where batcher could reuse signatures across chains. 

Now user signs hash of verification data + nonce + chain id

Contract verifies with same flow

# To test

Just test normal flow


batcher/aligned-batcher/src/lib.rs
```diff
@@ -150,6 +151,11 @@ impl Batcher {
             .try_into()
             .unwrap();
 
+        let chain_id = eth_rpc_provider
+            .get_chainid()
+            .await
+            .expect("Failed to get chain id");
+
```
NicolasRampoldi: If it fails we can try with the `eth_rpc_provider_fallback`

NicolasRampoldi (changes requested):

batcher/aligned-batcher/src/lib.rs
taturosati: fixed on e401d33

uri-99 (approved):

NicolasRampoldi (approved):

MauroToscano (changes requested):
Batcher should have the chainID in a config, to be able to filter signed transactions with a wrong chainID

entropidelic (approved):

MauroToscano (approved):

PR #840, opened 2024-08-22
feat: add more event emissions

### LINKED ISSUE #833: fix: Critical State‐Altering Operations Lack Event Emissions, Reducing Transparency and Auditability
# Description

- Critical state‐altering operations in the contracts do not emit events, which diminishes transparency and makes it harder to audit changes effectively.

# Examples

- The creation of a new task.
- The `locking` and `unlocking` of balances before withdrawal.
- The `checkMerkleRootAndVerifySignatures` function, which is public and updates user balances in `verifySignatureAndDecreaseBalance` but does not emit any events.

NicolasRampoldi:
# Description

- This PR addresses the lack of event emissions for operations in our smart contracts.

# Events added

## AlignedLayerServiceManager
- BatcherBalanceUpdated.

## BatcherPaymentService

- TaskCreated.
- BalanceLocked.
- BalanceUnlocked.

# To Test

- Try deployment and normal flow.


entropidelic:
I think we should be careful with the emission of events, they could significantly improve the costs of the transactions. Specially with the SignerBalanceDecreased and SignatureVerified, which will be emitted lots of times in every transaction. Make sure to measure this properly

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -148,6 +157,7 @@ contract AlignedLayerServiceManager is
             batchesState[batchMerkleRoot].batcherAddress
         ] -= txCost;
```
uri-99: Should emit BatcherBalanceUpdated.

Also in `receive()` , but i can't leave a comment on that line .-.

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -148,6 +157,7 @@ contract AlignedLayerServiceManager is
...
         payable(msg.sender).transfer(txCost);
+        emit TaskResponseProcessed(batchMerkleRoot, msg.sender);
```
uri-99: isn't this almost the same as previous `emit BatchVerified(batchMerkleRoot);` event?
I think this event is a bit redundant.

contracts/src/core/BatcherPaymentService.sol
```diff
@@ -21,6 +21,10 @@ contract BatcherPaymentService is
     // EVENTS
     event PaymentReceived(address indexed sender, uint256 amount);
     event FundsWithdrawn(address indexed recipient, uint256 amount);
+    event TaskCreated(bytes32 indexed batchMerkleRoot, string batchDataPointer);
+    event BalanceLocked(address indexed user);
+    event BalanceUnlocked(address indexed user, uint256 unlockBlock);
```
uri-99, line 26: should emit in `receive()` and `withdraw()` when that PR is merged

contracts/src/core/BatcherPaymentService.sol
```diff
@@ -223,6 +231,8 @@ contract BatcherPaymentService is
         } else {
             require(layer[0] == batchMerkleRoot, "Invalid merkle root");
         }
+
+        emit MerkleRootVerified(batchMerkleRoot);
```
uri-99: I can't seem to understand the added value of this event

uri-99 (changes requested):

contracts/src/core/AlignedLayerServiceManager.sol
NicolasRampoldi: You are right,  solved in [8dea4d6](https://github.com/yetanotherco/aligned_layer/pull/840/commits/8dea4d6b4c9b6046de9af200fd103296e641758b)

contracts/src/core/AlignedLayerServiceManager.sol
NicolasRampoldi: Solved in [e4e648b](https://github.com/yetanotherco/aligned_layer/pull/840/commits/e4e648bb1b9e00c20215d4562837c516abbb2142)

contracts/src/core/BatcherPaymentService.sol
NicolasRampoldi: Solved in [e9e442d](https://github.com/yetanotherco/aligned_layer/pull/840/commits/e9e442d6699faf0b643ecfd516bbf57aa2dbe4c0). Receive doesn't lock.

contracts/src/core/BatcherPaymentService.sol
NicolasRampoldi: It was pointed out by the auditors: 
> The `checkMerkleRootAndVerifySignatures` function, which is public and updates user balances in `verifySignatureAndDecreaseBalance` but does not emit any events.

Maybe it doesn't make sense, we can discuss it

taturosati: We made that function internal, it was a mistake it being public

taturosati:
> I think we should be careful with the emission of events, they could significantly improve the costs of the transactions. Specially with the SignerBalanceDecreased and SignatureVerified, which will be emitted lots of times in every transaction. Make sure to measure this properly

I agree, do we have measurements on gas usage with / without events, I assume its a fixed value extra per proof, since we are emitting N SignatureVerified events for N proofs.

contracts/src/core/BatcherPaymentService.sol
NicolasRampoldi: Removed in [86832b2](https://github.com/yetanotherco/aligned_layer/pull/840/commits/86832b2bd076581b83817d8dacf9eb106e5d087d)

taturosati (approved):
lgtm

uri-99 (approved):

entropidelic (approved):

PR #738, opened 2024-08-07
(Operator): Update ffi_binding to check buffers

PatStiles:
Draft pr updating the ffi bindings to handle zero size and null length inputs. @NicolasRampoldi @srosati The Halo2 bindings will need more work and updates within our fork.

PR #844, opened 2024-08-22
fix (batcher): remove panics

### LINKED ISSUE #815: Remove panics from batcher
Batcher has many panics. We should remove them all

taturosati:
Remove panics expects and unwraps.

Remaining panics are on constructor / initialize, to fail early on unrecoverable errors

# To test

Test normal flow


NicolasRampoldi (approved):
Worked for me, worth mentioning that the `unwraps` present in halo2 are being removed in Pat's PR #738. Also, not related to this PR but maybe we can change it here,
In line 148 of `aligned-batcher/src/lib.rs` we get the `last_uploaded_batch_block`:
```rust
  let last_uploaded_batch_block = eth_rpc_provider
            .get_block_number()
            .await
            .expect("Failed to get block number")
            .try_into()
            .unwrap();
```
We can try with the `eth_rpc_provider_fallback` if it fails.

entropidelic (approved):

PR #853, opened 2024-08-23
fix: change aligned layer service manager address to interface

### LINKED ISSUE #852: fix: BatcherPaymentService initialize AlignedLayerServiceManager parameter from address to IAlignedLayerServiceManager
# Description

- We should use the best type available in the function arguments and declarations instead of accepting address and performing low‐level calls or later casting it to the correct type.

NicolasRampoldi:
# Description

- In order to use the best type available in the function parameters instead of accepting AlignedLayerServiceManager address, this PR makes the function accept AlignedLayerServiceManager interface.

# To Test

- Try deploying and make sure everything works as usual.


taturosati (approved):
It works on my machine

entropidelic (approved):

PR #855, opened 2024-08-23
feat: handle new batch log on separate go routine

### LINKED ISSUE #854: fix(operator): download batches concurrently
it needs a timeout, and ideally some way to hold the request to the content-length demanded. Also processing one batch shouldn't block processing other batches. checking the proofs might be quick but downloading the batch is not actually predictable because it's IO.

taturosati:
Download batches separately to avoid locking for too long while downloading. Easiest way is to handle the new batch entirely in a different go routine.

Test normal flow. you can test with two batcher to see two batchers at the same time. bigger batches work better for this since they are heavier to download, ideally test with remote bucket for added latency.

### DIFF
--- a/operator/pkg/operator.go
+++ b/operator/pkg/operator.go
@@ -179,7 +179,6 @@ func (o *Operator) HandleNewBatchLog(newBatchLog *servicemanager.ContractAligned
 		OperatorId:          o.OperatorId,
 	}
 
-	// o.Logger.Infof("Signed Task Response to send: %+v", signedTaskResponse)
 	o.Logger.Infof("Signed Task Response to send: BatchIdentifierHash=%s, BatchMerkleRoot=%s, SenderAddress=%s",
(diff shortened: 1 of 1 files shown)


operator/pkg/operator.go
```diff
@@ -154,36 +154,41 @@ func (o *Operator) Start(ctx context.Context) error {
...
+
+	signedTaskResponse := types.SignedTaskResponse{
+		BatchIdentifierHash: batchIdentifierHash,
+		BatchMerkleRoot:     newBatchLog.BatchMerkleRoot,
+		SenderAddress:       newBatchLog.SenderAddress,
+		BlsSignature:        *responseSignature,
+		OperatorId:          o.OperatorId,
+	}
+
+	// o.Logger.Infof("Signed Task Response to send: %+v", signedTaskResponse)
```
entropidelic:
```suggestion
```

entropidelic (approved):

NicolasRampoldi (approved):

PR #864, opened 2024-08-26
fix: batcher nonce verification order in handle_message function

### LINKED ISSUE #861: fix (batcher): nonce verification order in handle_message function
# Description

- The current implementation of the `handle_message` function in the batcher performs nonce verification using the `check_nonce_and_increment` method, which increments the nonce before pre-verifying the proof. This order of operations may introduce vulnerabilities, specifically to replay attacks and potential DoS.

### LINKED ISSUE #1081: replay attack on batcher disrupts a particular user
## Problem
In the Batcher, in handle_message, a ClientMessage is received, the signature is checked (but the nonce is not checked yet), then the proof count is checked and incremented for that user.

The proof count is used to estimate whether the user has enough Ethereum balance to pay for the proof. The proof count is not reset until after the batch is submitted, so any subsequent proof requests within this batch will have an incremented proof count

The proof is then validated, and finally, the nonce is checked and incremented. If the ClientMessage was replayed, the proof will be valid but the nonce will not be.
Consequently, the message will be dropped and will not be included in the batch.

However, because the proof count is still incremented, it will appear as though the user needs more balance to pay for the next proof that they submit. If an attacker submits many replays, they can eventually increase the cost estimate, such that a real proof will not be accepted.

An Aligned Layer user is expected to have a minimum balance per proof of 13,000*100 gwei = 0.0013 Ether, which, as of the writing of this report, is equivalent to 3.36 USD. Hence, if the attacker is able to submit 30 invalid proofs (increasing the proof count to 30), the Batcher will then require the user to have more than 100 USD in their balance.
Moreover, since the balance for a user is public information, the attacker can check how many invalid proofs need to be replayed.

## Remediation
Check the nonce at the start when checking the signature but refraining from incrementing it until the end, once the ClientMessage is fully validated. 

## Additional Remediation Considerations (Mauro)

Just checking the nonces without locking the whole state will lead to race conditions. Locking the whole state is also problematic, since it will make the whole batcher work sequentially. We need to move the UserState away from the BatchQueue, and have locks per user, to allow concurrency and avoid. this race condition

NicolasRampoldi:
Closes #1081 

# Description

- This PR refactors the message handling process to address the security concerns detailed in the audit.
- To achieve this, it does the following:
  1. Check the nonce before verifying the proof in both `handle_message` and `handle_nonpaying_msg`.
  2. Separate nonce checking and incrementing operations to ensure atomicity.
  3. Improve error handling and messaging for invalid nonces and proofs.

# To Test

- Test full flow with paying and non-paying address. Make sure everything is working correctly.


MauroToscano (changes requested):
User can submit a transaction with the same nonce while the proofs are verifying and this could end up in an invalid state. The nonce needs to be blocked while this process is going.

NicolasRampoldi:
> User can submit a transaction with the same nonce while the proofs are verifying and this could end up in an invalid state. The nonce needs to be blocked while this process is going.

Solved in [aff5aac](https://github.com/yetanotherco/aligned_layer/pull/864/commits/aff5aac858c3c8f43273b6df1b8307e7453aae9a)

entropidelic:
it seems like when a createNewTask transaction is reverted, the nonce is set to an invalid state and then all proofs sent with that address get an `invalid nonce` error.
To replicate, try sending two bursts of the same proof. For example:
```
make batcher_send_risc0_burst
```
when the transaction is reverted because of the repeated merkle root, you should not be able to send any more proofs with that address

NicolasRampoldi:
> it seems like when a createNewTask transaction is reverted, the nonce is set to an invalid state and then all proofs sent with that address get an `invalid nonce` error. To replicate, try sending two bursts of the same proof. For example:
> 
> ```
> make batcher_send_risc0_burst
> ```
> 
> when the transaction is reverted because of the repeated merkle root, you should not be able to send any more proofs with that address

Solved in [ea00344](https://github.com/yetanotherco/aligned_layer/pull/864/commits/ea00344cc75b3f5ad1ecf6170b7e3d15755be995). I added a condition that thought was necessary but it didn't make sense

entropidelic (approved):

batcher/aligned-batcher/src/lib.rs
```diff
@@ -541,6 +535,7 @@ impl Batcher {
                     }
                     std::cmp::Ordering::Equal => {
                         // if we are here nonce == expected_user_nonce
+
```
taturosati:
```suggestion
```

uri-99:
Being solved in #1087 and 'refactor-user-states'

PR #887, opened 2024-08-28
fix: add new_batch_v2 emit compatibility

### LINKED ISSUE #885: fix(explorer): add new_batch_v2 compatibility

glpecile:
# Changes

* merge after #888
* update abi
* add `v2` to new batch


uri-99 (changes requested):
This PR should contain in it those contract changes, otherwise it does not work.

I think these changes should be on https://github.com/yetanotherco/aligned_layer/pull/875 , or bring those changes to this branch so the explorer correlates to the contract being queried.

Alternatively, this PR could point to #875 , making 875 a the PR which adds _v2 compatibility

PR #906, opened 2024-08-29
feat(explorer): add calculator view

### LINKED ISSUE #905: feat(explorer): add calculator view
# Objective

Show an estimate of how much aligned costs and saves you.

## Cost of verifying a batch of proofs

The main components are:

* **Cost of BLS aggregator task response in Ethereum**: ~constant, 400000 gas. Can vary depending on the amount of Operators that didn't sign but we can't know this beforehand). This cost is paid by the aggregator when interacting with the *ServiceManager* and must be refunded to it.
* **Cost of creating the task**: $BaseCost$ (~100000 gas) + $CostPerProof$ (~13000 gas) x $NProofs$. The cost per proof is related to a for loop in the EVM. This cost is paid by the Batcher when interacting with the *BatcherPaymentService* and must be refunded to it.

In the end, the total cost is,

$C(n) = TaskResponseCost + BaseTaskCreationCost + CostPerProof * n$

where $n$ is the number of proofs in a batch.

The cost per proof is then,

$c(n) = \frac{TaskResponseCost + BaseTaskCreationCost}{n} + CostPerProof$

## Mockup

image

glpecile:
# Changes

* add new calculator view

### DIFF
--- a/explorer/lib/explorer_web/components/core_components.ex
+++ b/explorer/lib/explorer_web/components/core_components.ex
@@ -232,6 +232,7 @@ defmodule ExplorerWeb.CoreComponents do
   attr :rest, :global, include: ~w(disabled form name value)
   attr :icon, :string, default: nil
   attr :icon_class, :string, default: nil
+  attr :variant, :string, default: "default"
 
   slot :inner_block, required: true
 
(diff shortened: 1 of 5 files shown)


JuArce:
Stale

PR #910, opened 2024-08-30
feat: add limit on aggregator spending

### LINKED ISSUE #881: feat: Add limit on aggregator spending
Batcher must be able to specify a limit on how much Ether the aggregator is allowed to consume to respondToTask.

Make it 3x the amount of the Ether the Agg is expected to use.

(Response is usually included in 3 blocks, this covers a spike of gas)

uri-99:
# This PR

- This PR adds to the Batcher the ability to set a `max_fee_allowed_to_respond`.
  - This variable is currently being set in `lib.rs` as `MAX_FEE_FOR_AGGREGATOR_MULTIPLIER` , where it is used to multiply it to `fee_for_aggregator`, to get the `max_fee_allowed_to_respond`.
- Aggregator can't spend more fee than what was set in `max_fee_allowed_to_respond` , to respond that task

## To Deploy

This PR changes no memory, so a normal upgrade must be run for both AlignedLayerServiceManager and BatcherPaymentService.

### Happy Path
```
```
```
make aggregator_start
```
```
```
```
make batcher_start_local
```
```
make run_explorer
```

And when they finish starting up, you can send proofs:
```
make batcher_send_risc0_burst
```

### Playing around max_fee_allowed_to_respond

For this you can do, for example:
- change the MAX_FEE_FOR_AGGREGATOR_MULTIPLIER to different values and check it behaves accordingly
- change/hardcode maxFeeAllowedToRespond values on the `BatcherPaymentService` and check it behaves accordingly.
- change/hardcode maxFeeAllowedToRespond values on the `AlignedServiceManager`, and check it behaves accordingly.


contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -141,26 +153,25 @@ contract AlignedLayerServiceManager is
...
+        uint256 txCost = (initialGasLeft - gasleft() + 70_000) * tx.gasprice;
 
-        if (batchersBalances[senderAddress] < txCost) {
+        if (txCost > currentBatch.maxFeeAllowedToRespond) {
             revert InsufficientFunds(
                 senderAddress,
-                txCost,
-                batchersBalances[senderAddress]
+                currentBatch.maxFeeAllowedToRespond,
+                txCost
```
entropidelic: this should raise another error, like "ExceededMaxFee" or something similar

uri-99: solved

batcher/aligned-batcher/src/eth/mod.rs
```diff
@@ -29,14 +29,21 @@ pub struct CreateNewTaskFeeParams {
     pub fee_for_aggregator: U256,
     pub fee_per_proof: U256,
     pub gas_price: U256,
+    pub max_fee_allowed_to_respond: U256,
```
taturosati:
```suggestion
    pub max_fee_for_aggregator: U256,
```

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -49,7 +49,8 @@ contract AlignedLayerServiceManager is
 
     function createNewTask(
         bytes32 batchMerkleRoot,
-        string calldata batchDataPointer
+        string calldata batchDataPointer,
+        uint256 maxFeeAllowedToRespond
```
taturosati:
```suggestion
        uint256 maxFeeForAggregator
```

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -98,30 +105,35 @@ contract AlignedLayerServiceManager is
...
-        if (batchersBalances[senderAddress] == 0) {
-            revert BatcherHasNoBalance(senderAddress);
+        // Check that batcher has enough funds to fund response
+        if (batchersBalances[senderAddress] < currentBatch.maxFeeAllowedToRespond) {
+            revert InsufficientFunds(
+                senderAddress,
+                currentBatch.maxFeeAllowedToRespond,
+                batchersBalances[senderAddress]
+            );
         }
```
taturosati: This is fine for now and better than what we had, but a batcher may have insufficient funds to pay for a second or third pending batch

contracts/src/core/AlignedLayerServiceManagerStorage.sol
```diff
@@ -4,6 +4,7 @@ abstract contract AlignedLayerServiceManagerStorage {
     struct BatchState {
         uint32 taskCreatedBlock;
         bool responded;
+        uint256 maxFeeAllowedToRespond;
```
taturosati:
```suggestion
        uint256 maxFeeForAggregator;
```

contracts/src/core/BatcherPaymentService.sol
```diff
@@ -111,7 +110,8 @@ contract BatcherPaymentService is
         bytes32[] calldata leaves, // padded to the next power of 2
         SignatureData[] calldata signatures, // actual length (proof sumbitters == proofs submitted)
         uint256 feeForAggregator,
-        uint256 feePerProof
+        uint256 feePerProof,
+        uint256 maxFeeAllowedToRespond
```
taturosati:
```suggestion
        uint256 maxFeeForAggregator
```

contracts/src/core/IAlignedLayerServiceManager.sol
```diff
@@ -9,28 +9,29 @@ interface IAlignedLayerServiceManager {
         bytes32 indexed batchMerkleRoot,
         address senderAddress,
         uint32 taskCreatedBlock,
-        string batchDataPointer
+        string batchDataPointer,
+        uint256 maxFeeToRespond
```
taturosati:
```suggestion
        uint256 maxFeeForAggregator
```

taturosati (commented):
just refactors, I think code looks good, need to test thoroughly

batcher/aligned-batcher/src/lib.rs
```diff
@@ -56,6 +56,7 @@ const ADDITIONAL_SUBMISSION_GAS_COST_PER_PROOF: u128 = 13_000;
 const CONSTANT_GAS_COST: u128 = AGGREGATOR_GAS_COST + BATCHER_SUBMISSION_BASE_GAS_COST;
 const DEFAULT_MAX_FEE_PER_PROOF: u128 = ADDITIONAL_SUBMISSION_GAS_COST_PER_PROOF * 100_000_000_000; // gas_price = 100 Gwei = 0.0000001 ether (high gas price)
 const MIN_FEE_PER_PROOF: u128 = ADDITIONAL_SUBMISSION_GAS_COST_PER_PROOF * 100_000_000; // gas_price = 0.1 Gwei = 0.0000000001 ether (low gas price)
+const MAX_FEE_FOR_AGGREGATOR_MULTIPLIER: u128 = 2;
```
NicolasRampoldi: Maybe we can add a comment explaining why `x2` is enough.

contracts/src/core/AlignedLayerServiceManager.sol
uri-99: Yes, but for now this can't be checked on createNewTask.
Also, Batcher may have N pending batches that will never be accepted, and he should not need to have funds for these batches.

taturosati (changes requested):
Happy path is not working, getting revert with `0x5c54305e0000000000000000000000007969c5ed335650692bc04293b07f5bf2e7a673c00000000000000000000000000000000000000000000000000002e357ab61e200000000000000000000000000000000000000000000000000000171abd5b0f100` which is InsufficientFunds

taturosati:
> Happy path is not working, getting revert with `0x5c54305e0000000000000000000000007969c5ed335650692bc04293b07f5bf2e7a673c00000000000000000000000000000000000000000000000000002e357ab61e200000000000000000000000000000000000000000000000000000171abd5b0f100` which is InsufficientFunds

More info regarding this, doing a cast send does not fix this issue. I think the cause of this is that the batcher payment service needs to have balance and not the batcher itself.

I had to change the multiplier to 1 for some batches to get it started. This gives it some balance for it to have enough for the 2x multiplier.

Possible solution to this is adding a method on the batcher payment service to fund itself on the aligned service manager (which would basically be a transfer from contract to contract).

batcher/aligned-batcher/src/eth/mod.rs
uri-99: respondToTaskFeeLimit

contracts/src/core/AlignedLayerServiceManager.sol
uri-99: respondToTaskFeeLimit

contracts/src/core/AlignedLayerServiceManagerStorage.sol
uri-99: respondToTaskFeeLimit

contracts/src/core/BatcherPaymentService.sol
uri-99: respondToTaskFeeLimit

contracts/src/core/IAlignedLayerServiceManager.sol
uri-99: respondToTaskFeeLimit

uri-99:
@taturosati 
Added deposit funds to AlignedLayerServiceManager in `make anvil_deploy_aligned_contracts` .
No new function was needed to implement this, the existing `depositToBatcher(address)` was used

entropidelic (approved):

JuArce (approved):

PR #916, opened 2024-09-03
feat: make signing eip 712 compliant

### LINKED ISSUE #877: Make signing EIP-712 compliant
Signatures should be fully [EIP-712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) compliant

Domain should be extended to include name and version, and data types hashed properly

We can also change our library to alloy and use their functionalities to not redo-it ourselves

NicolasRampoldi:
> [!IMPORTANT]  
> Added `chain` param to `submit` and `submit_multiple` in the SDK. This is needed to get the `BatcherPaymentService` contract address.

> [!IMPORTANT]  
> The version of the Aligned contracts was changed from =0.8.12 to ^0.8.12 to enable compatibility with open-zeppelin 5.0.0 (which is needed to use EIP712). We still use open-zeppelin from eigenlayer-middleware for all the imports but just for the EIP712 we use open-zeppelin 5.0.0.

> [!NOTE]  
> The GAP variable was lowered by 1 because of the added NONCED_VERIFICATION_DATA_TYPEHASH bytes32 constant.

> [!NOTE]  
> As stated in the code's comment, VerificationData was used as a bytes32 instead of a VerificationData struct because we don't have the necessary fields in the contract (BatcherPaymentService) to use the VerificationData struct. Also, chain_id is not part of the type hash because it is now part of the domain.

If you want to read more about the [EIP-712](https://eips.ethereum.org/EIPS/eip-712).

# Description

- This PR makes the `NoncedVerificationData` signing EIP712 compliant. The EIP712Domain has:

  ```
  name: Aligned
  version: 1
  chain_id: <current_chain_id>
  verifying_contract: <current_payment_service contract_addr>
  ```

# Deploying
## Devnet:

`make anvil_add_type_hash_to_batcher_payment_service`

## Testnet:

- Stage or prod variables should be set accordingly on the contracts/scripts/.env file

`make upgrade_add_type_hash`

# Working with Metamask

https://github.com/user-attachments/assets/decd274b-6423-4d8e-89e3-718d0f49fa9c

# To Test
- Run `make deps`.
- Run everything as usual and make sure it is working properly.


batcher/aligned-sdk/src/core/types.rs
```diff
@@ -265,3 +319,15 @@ pub enum Chain {
...
+    match chain_id.as_u64() {
+        ANVIL_CHAIN_ID => H160::from_str("0x7969c5eD335650692Bc04293B07F5BF2e7A673C0").ok(),
+        HOLESKY_CHAIN_ID => H160::from_str("0x815aeCA64a974297942D2Bbf034ABEe22a38A003").ok(),
+        MAINNET_CHAIN_ID => {
+            //FIXME: Add the payment service address for mainnet
+            None
+        }
+        _ => None,
+    }
+}
```
taturosati: If we want to support different batchers I think the payment service address will need to be a field in the struct, and be passed by the client. I think most sdk methods use the payment service address now so it shouldn't be too hard to add

batcher/aligned-sdk/src/core/types.rs
```diff
@@ -170,14 +186,62 @@ pub struct ClientMessage {
...
+        let nonce_hash = hasher.finalize_reset();
+
+        hasher.update(
+            [
+                nonced_verification_data_type_hash.as_slice(),
+                verification_data_hash.as_slice(),
+                nonce_hash.as_slice(),
+            ]
+            .concat(),
+        );
```
taturosati: check if this is equivalent to calling hasher.update with each slice instead of using concat:

```suggestion
        hasher.update(nonced_verification_data_type_hash.as_slice());
        hasher.update(verification_data_hash.as_slice());
        hasher.update(nonce_hash.as_slice());
```

NicolasRampoldi: Yes, it is equivalent. Thanks for point it out, solved in [f259a96](https://github.com/yetanotherco/aligned_layer/pull/916/commits/f259a96e4707d809fb89e5d5d6601e82e9668c17)

batcher/aligned-sdk/src/core/types.rs
NicolasRampoldi: I agree, I think we can add that in another PR

batcher/aligned-sdk/src/core/types.rs
```diff
@@ -278,3 +329,15 @@ pub enum Chain {
...
+    match chain_id.as_u64() {
+        ANVIL_CHAIN_ID => H160::from_str("0x7969c5eD335650692Bc04293B07F5BF2e7A673C0").ok(),
+        HOLESKY_CHAIN_ID => H160::from_str("0x815aeCA64a974297942D2Bbf034ABEe22a38A003").ok(),
+        MAINNET_CHAIN_ID => {
+            //FIXME: Add the payment service address for mainnet
+            None
+        }
+        _ => None,
+    }
+}
```
taturosati: There is another problem with this, we are missing holesky stage

taturosati (changes requested):

batcher/aligned-sdk/src/core/types.rs
NicolasRampoldi: Solved in [aab9476](https://github.com/yetanotherco/aligned_layer/pull/916/commits/aab94768d8f020c4d77f62506875916c00893c97)

batcher/aligned-sdk/src/core/types.rs
```diff
@@ -196,32 +255,18 @@ impl ClientMessage {
...
-        let hashed_data: [u8; 32] =
-            ClientMessage::hash_with_nonce_and_chain_id(&self.verification_data);
+        let recovered = self.signature.recover_typed_data(&self.verification_data)?;
+
+        // We can expect here because encode_eip712 can only error if
+        // struct_hash or domain_separator return an error, which is not possible
+        let hashed_data = self
+            .verification_data
+            .encode_eip712()
+            .expect("Failed to encode verification data for signature verification");
```
entropidelic: 👮 return error instead of expect

NicolasRampoldi: I used expect because of this:
```
We can expect here because encode_eip712 can only error if
struct_hash or domain_separator return an error, which is not possible
```
I added it as a comment in lines 260 and 261

NicolasRampoldi: I used expect because `encode_eip712` couldn't error since we defined `struct_hash` and `domain_separator` as I mentioned in the comment above the function call. But you are right it is better to return the error anyways. I solved it in [e8be0d7](https://github.com/yetanotherco/aligned_layer/pull/916/commits/e8be0d7b553b1fb80c0f96a60fa491a116a47526) adding a custom error.

contracts/src/core/BatcherPaymentService.sol
```diff
@@ -68,12 +70,17 @@ contract BatcherPaymentService is
     // map to user data
     mapping(address => UserInfo) public userData;
 
+    bytes32 private constant NONCED_VERIFICATION_DATA_TYPEHASH =
+        keccak256(
+            "NoncedVerificationData(bytes32 verification_data_hash,bytes32 nonce,uint256 max_fee)"
+        );
```
uri-99: The content of the variable should be set in a reinitializer function, and possibly make it public with a onlyOwner setter in case we might want to add more fields to this parameter

uri-99 (changes requested):
Didn't finish the review but I leave you this request changes on the mean time

contracts/src/core/BatcherPaymentService.sol
uri-99: Also maybe a verifier of this value might turn useful?
a public view that takes string as input and verifies `keccak(string) == stored NONCED_VERIFICATION_DATA_TYPEHASH`

entropidelic (dismissed):

contracts/src/core/BatcherPaymentService.sol
NicolasRampoldi: Solved in [4ffd79c](https://github.com/yetanotherco/aligned_layer/pull/916/commits/4ffd79c1222671c465818bbbadf416992ef58b88)

contracts/src/core/BatcherPaymentService.sol
```diff
@@ -87,14 +91,28 @@ contract BatcherPaymentService is
...
 
         alignedLayerServiceManager = _alignedLayerServiceManager;
         batcherWallet = _batcherWallet;
+        noncedVerificationDataTypeHash = _noncedVerificationDataTypeHash;
+    }
+
+    function initializeNoncedVerificationDataTypeHash(
+        bytes32 _noncedVerificationDataTypeHash
+    ) public reinitializer(2) onlyOwner {
+        noncedVerificationDataTypeHash = _noncedVerificationDataTypeHash;
```
uri-99, line 85:
```suggestion
        setNoncedVerificationDataTypeHash(_noncedVerificationDataTypeHash);
```

NicolasRampoldi: I think it's more expensive and doesn't add any improvement in readability

uri-99 (approved):

entropidelic (approved):

MauroToscano:
Signed Fee should appear. Nonce would be better if it's not hashed, but I think this in inherited from some weird decision on the code

MauroToscano (changes requested):
Signed Fee should appear. Nonce would be better if it's not hashed, but I think this in inherited from some weird decision on the code

MauroToscano (approved):

MauroToscano (approved):

PR #918, opened 2024-09-03
feat: aggregator should check if gas is enough before respondToTask

### LINKED ISSUE #882: feat: Aggregator should check if gas is enough before respondToTask
Agg should be updated to check if the gas is enough to pay before sending tx

uri-99:
# This PR

- Aggregator runs a series of new checks before submitting `RespondToTask` tx:
  - runs a simulation of `RespondToTask` and checks `RespondToTaskFeeLimit` is enough
  - checks Batcher has enough funds to pay for `RespondToTaskFeeLimit`
  - checks Agg has enough funds to pay for `RespondToTaskFeeLimit`
 
In this PR, the variable `batchersBalances` was changed from `internal` to `public`; because the aggregator must check how much balance does the Batcher has, to compare it with `RespondToTaskFeeLimit`.

## To Deploy

Aggregator code has changed (adding new checks) , so it must be redeployed.
AlignedLayerServiceManager variable was changed from `internal` to `public`, so it must be redeployed. 

### Happy Path
```
```
```
make aggregator_start
```
```
```
```
make batcher_start_local
```
```
make run_explorer
```

And when they finish starting up, you can send proofs:
```
make batcher_send_risc0_burst
```

## Unhappy paths
- Empty Aggregator balance:
```
cast send 0x0000000000000000000000000000000000000000 --value 9999998499999944500000 --private-key 0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a --rpc-url "http://localhost:8545"
```
Note cast may not show the tx as completed, but you can verify batchers balance with:
```
cast balance 0x15d34AAf54267DB7D7c367839AAf71A00a2C6A65 --rpc-url localhost:8545
```
Then send proofs, Aggregator should not try to submit the tx
```
make batcher_send_risc0_burst
```

- change the `DEFAULT_AGGREGATOR_FEE_MULTIPLIER` and `RESPOND_TO_TASK_FEE_LIMIT_MULTIPLIER` to different values, and check it behaves accordingly
- hardcode respondToTaskFeeLimit values on the BatcherPaymentService or AlignedServiceManagerand, and check it behaves accordingly
- (These 2 cases are pretty similar but one may be favourable over the other depending which one you understand better and which value you want to test more in detail)


core/chainio/avs_writer.go
```diff
@@ -91,3 +102,86 @@ func (w *AvsWriter) SendAggregatedResponse(batchMerkleRoot [32]byte, senderAddre
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (w *AvsWriter) compareAggregatorBalance(amount *big.Int, aggregatorAddress common.Address) error {
+	// Get Agg wallet balance
+	aggregatorBalance, err := w.Client.BalanceAt(context.TODO(), aggregatorAddress, nil)
```
NicolasRampoldi: Change the `context.TODO()` with `context.Background` or `context.WithTimeout`. Also, the proper thing would be to receive the context as a parameter in this function `func (w *AvsWriter) compareAggregatorBalance(ctx context.Context)`. More on this: https://stackoverflow.com/a/74239191/13590282

core/chainio/avs_writer.go
```diff
@@ -91,3 +102,86 @@ func (w *AvsWriter) SendAggregatedResponse(batchMerkleRoot [32]byte, senderAddre
...
+	if err != nil {
+		aggregatorBalance, err = w.ClientFallback.BalanceAt(context.TODO(), aggregatorAddress, nil)
```
NicolasRampoldi: Change the `context.TODO()` with `context.Background` or `context.WithTimeout`. Also, the proper thing would be to receive the context as a parameter in this function `func (w *AvsWriter) compareAggregatorBalance(ctx context.Context)`. More on this: https://stackoverflow.com/a/74239191/13590282

NicolasRampoldi (changes requested):

core/chainio/avs_writer.go
```diff
@@ -91,3 +103,86 @@ func (w *AvsWriter) SendAggregatedResponse(batchMerkleRoot [32]byte, senderAddre
...
+	}
+	w.logger.Info("Aggregator balance", "balance", aggregatorBalance)
+	if aggregatorBalance.Cmp(amount) < 0 {
+		return fmt.Errorf("cost is higher than Aggregator balance")
+	}
+	return nil
+}
+
+func (w *AvsWriter) compareBatcherBalance(amount *big.Int, senderAddress [20]byte) error {
+	// Get batcher balance
```
taturosati, line 169: maybe this should get a context as well
```suggestion
func (w *AvsWriter) compareBatcherBalance(amount *big.Int, senderAddress [20]byte) error {
        ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()
	// Get batcher balance
```

core/chainio/avs_writer.go
```diff
@@ -91,3 +103,86 @@ func (w *AvsWriter) SendAggregatedResponse(batchMerkleRoot [32]byte, senderAddre
...
+		if err != nil {
+			// Fallback also failed, proceed with simulated cost checks
+			w.logger.Error("Failed to get batch state", "error", err)
+			w.logger.Info("Proceeding with simulated cost checks")
+			if err := w.compareAggregatorBalance(simulatedCost, aggregatorAddress); err != nil {
+				return err
+			}
+			if err := w.compareBatcherBalance(simulatedCost, senderAddress); err != nil {
+				return err
+			}
```
taturosati: smol nitpick, this sequence could be moved to one function since we are doing it twice

core/chainio/avs_writer.go
uri-99: We are using the `ServiceManager` and `ServiceManagerFallback` for the `get BatchersBalances` call, which doesn't need a context to be passed to it (i'm guessing it is already configured at this point).
While in the other case we use `eth.Client`, which needs a context

core/chainio/avs_writer.go
uri-99: Sure why not, done

entropidelic (approved):

PatStiles (approved):

PR #934, opened 2024-09-06
fix(explorer): count only active operators for weight

### LINKED ISSUE #933: fix(explorer): inactive operators should not count their stake
Their stake should not be counted for the total, and their % should be 0, even though they may have some stake, because he is not involved in the voting system

uri-99:
# Changes

* Feat: Conditionally add weight to DB only if the operator is active 
* Fix: Update operator.total_stake after restaking changes
* Fix: correctly manage operator unregistration

## Test

Follow the usual steps to get aligned up and running. These are:

Each command with a number is meant to be run on a separate terminal instance/tab.

1. Starts anvil

```
```

2. On another Starts the aggregator
```
make aggregator_start
```
3. Registers and starts the operator:

```
```

4. Start Explorer
For better test experience, you should change the following code line:
In `explorer/lib/explorer/periodically.ex`
Line 20, change from:
`:timer.send_interval(one_second * seconds_in_an_hour, :restakings) # every 1 hour`
to
`:timer.send_interval(one_second * 30, :restakings) # every 30 seconds`
Now, you can run_explorer

```
make run_explorer
```

5. (On Another Tab) Fetch old operator changes in explorer

```
make explorer_fetch_old_operators_strategies_restakes
```

You will know this script is done when you are handed control over the `iex` (interactive elixir console). You will get control with the classic `>` symbol.
You can close this interactive console by pressing `ctrl + c` twice.

6.
After this script, you should be able to see the new, updated, `Total Active Operators` value in the Explorer.

7.
To disable an operator, you must `eject` it.
For this, you can must run the following command:

```sh
cast send \
	0x851356ae760d987E095750cCeb3bC6014560891C \
	 --rpc-url localhost:8545 \
	 "ejectOperator(address, bytes)" 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 0x00 \
	 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
```

After ejecting an Operator, his status should change to disabled, and his stake go down to 0, as well as the total Stake should decrement accordingly.

8.
You can also run more than 1 operator, and eject only 1, to see values change accordingly.
To run more than 1 operator, you can change line 5 of `Makefile` to the following:
```
CONFIG_FILE?=config-files/config-operator-2.yaml
```

And then on another terminal:
```
```


glpecile (approved):

MauroToscano (approved):

MauroToscano:
May need extra testing

MauroToscano (changes requested):
Add steps to test

explorer/lib/explorer/models/operators.ex
```diff
@@ -32,6 +32,17 @@ defmodule Operators do
     Operators.changeset(%Operators{}, Map.from_struct(operator))
   end
 
+  def generate_new_total_stake_changeset(%{operator_address: operator_address}) do
+    new_total_stake = StakeRegistryManager.get_stake_of_quorum_for_operator(%Restakings{operator_address: operator_address})
+    dbg operator_address
+    dbg new_total_stake
```
glpecile:
```suggestion
```

These debugs should be deleted

glpecile (changes requested):
need re-review

explorer/lib/explorer/periodically.ex
```diff
@@ -17,7 +17,7 @@ defmodule Explorer.Periodically do
     seconds_in_an_hour = 60 * 60
 
     :timer.send_interval(one_second * 12, :batches) # every 12 seconds, once per block
-    :timer.send_interval(one_second * seconds_in_an_hour, :restakings) # every 1 hour
+    :timer.send_interval(one_second * 30, :restakings) # every 1 hour
```
glpecile: Is this change a leftover from debugging?

NicolasRampoldi, line 20: Should this be reverted to `seconds_in_an_hour`? I think it was changed for testing purposes

Makefile
```diff
@@ -162,7 +162,7 @@ operator_deposit_into_mock_strategy:
 	@go run operator/cmd/main.go deposit-into-strategy \
 		--config $(CONFIG_FILE) \
 		--strategy-address $(STRATEGY_ADDRESS) \
-		--amount 1000
+		--amount 100000000000000000
```
glpecile, line 173: Is this meant to be changed?

Makefile
```diff
@@ -143,7 +143,7 @@ operator_register_with_eigen_layer:
 
 operator_mint_mock_tokens:
 	@echo "Minting tokens"
-	. ./scripts/mint_mock_token.sh $(CONFIG_FILE) 1000
+	. ./scripts/mint_mock_token.sh $(CONFIG_FILE) 100000000000000000
```
glpecile, line 154: Should this be changed back to 1000?

explorer/lib/explorer/models/restakings.ex
```diff
@@ -29,6 +29,7 @@ defmodule Restakings do
     Operators.get_operators()
       |> Enum.map(fn operator -> StakeRegistryManager.has_operator_changed_staking(%{fromBlock: from_block, operator_id: operator.id, operator_address: operator.address}) end)
       |> Enum.reject(fn {_operator_id, _operator_address, has_changed_stake} -> not has_changed_stake end)
+      |> Enum.reject(fn {operator_id, _operator_address, _has_changed_stake} -> not Operators.get_operator_by_id(operator_id).is_active end)
```
NicolasRampoldi, line 32: Could this be done in the same function call? Like:
```elixir
|> Enum.reject(fn {operator_id, _operator_address, has_changed_stake} -> not has_changed_stake or not Operators.get_operator_by_id(operator_id).is_active end)
```

explorer/lib/explorer/models/strategies.ex
```diff
@@ -99,4 +99,14 @@ defmodule Strategies do
     Strategies.changeset(strategy, %{total_staked: new_stake})
   end
 
+  def discount_restaking(restaking) do
+    query = from(s in Strategies,
+      where: s.strategy_address == ^restaking.strategy_address,
+      select: s)
+    strategy = Explorer.Repo.one(query)
+
+    new_stake = Decimal.sub(strategy.total_staked, restaking.stake)
```
glpecile: What happens in this case if the new_stake is negative for example?

explorer/lib/explorer/periodically.ex
uri-99: yes, reverted

Makefile
uri-99: Yes, 1000 wei is too small value to be shown in operator stakes. Now its 0.1 eth

Makefile
uri-99: No, 1000 wei is too small value to be shown in operator stakes. Now its 0.1 eth

explorer/lib/explorer/models/restakings.ex
uri-99: It would work but it would be less efficient because `Operators.get_operator_by_id()` would be called for every operator

explorer/lib/explorer/models/strategies.ex
uri-99: Logically it should never happen, but I added a check just in case

MauroToscano (approved):

entropidelic (approved):

PR #949, opened 2024-09-10
chore(aggregator): implement telemetry for aggregator

JuArce:
# Description
This PR allows the aggregator to export telemetry traces to have a better observability of the system

# Tasks

- [x] Docker compose with Otel Collector and Jaeger UI
- [x] Init traces on NewTask
- [x] Add span on Operator responses
- [x] Add span on Quorum Reached
- [x] Add span on Tasks errors (ie task expired)
- [x] Close traces after a period of time the quorum was reached to wait for most operators responses 
    

# How to Test

> [!NOTE]  
> Docker is needed to test the PR.

1. Run telemetry
```
make run_telemetry
```
It will run the Otel-Collector and Jaeger UI.
> [!NOTE]  
> Jaeger UI is available on http:localhost:16686
image

2. Run anvil
```
```

3. Run aggregator
```
make aggregator_start
```

4. Run operator
```
export OPERATOR_ADDRESS=0x70997970C51812dc3A010C7d01b50e0d17dc79C8
make operator_full_registration CONFIG_FILE=config-files/config-operator-1.yaml
make operator_start CONFIG_FILE=config-files/config-operator-1.yaml
```

5. Run batcher
```
make batcher_start_local
```

6. Send tasks 

7. Once you sent a batch and it has been responded by the aggregator, you should be able to check the trace of the task in the Jaeger UI (`http:localhost:16686`)

In the search, select service: `aggregator` and press `find traces`
It should show you the tasks you sent to aligned


aggregator/internal/pkg/aggregator.go
```diff
@@ -251,10 +268,13 @@ func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsA
 
 	agg.taskMutex.Lock()
 	agg.AggregatorConfig.BaseConfig.Logger.Info("- Locked Resources: Fetching merkle root")
-	batchIdentifierHash := agg.batchesIdentifierHashByIdx[blsAggServiceResp.TaskIndex]
+	batchIdentifierHash = agg.batchesIdentifierHashByIdx[blsAggServiceResp.TaskIndex]
```
taturosati: This is already set on line 231
```suggestion
```

telemetry/telemetry.go
taturosati, line 1: We should probably add a TaskExpiredTrace when a task expires instead of reaching threshold, ideally should included quorum % and operators who responded if we have access to that information.

taturosati:
Other than comments, worked perfectly on my machine

uri-99 (changes requested):
Documentation of this PR would be highly valuable, maybe even with some screenshots.
(review still WIP)

PatStiles (approved):

uri-99:
image

This Warning seems to be appearing when the task was not yet responded, and it disappears afterwards. Can we change the text of the warning?

uri-99:
Is there a way to make Jaeger auto-refresh?

uri-99:
Have you considered setting up the `/monitor` ? Seems like it is a real-time visualization of the items found with the search.

image

telemetry/telemetry.go
```diff
@@ -0,0 +1,263 @@
...
+		initialCtx:                ctx,
+		shutdownTracerProvider:    shutdownTracerProvider,
+		shutdownMeterProvider:     shutdownMeterProvider,
+	}
+}
+
+// Telemetry functions
+
+// InitNewTrace Init a new trace for the given batchMerkleRoot
+// User must make sure to call FinishTrace()
```
uri-99, line 87: What happens if Agg crashes before calling FinishTrace() ?

update: it never uploads the Trace

telemetry/telemetry.go
```diff
@@ -0,0 +1,263 @@
...
+func (t *Telemetry) QuorumReachedTrace(batchMerkleRoot [32]byte) trace.Span {
+	ctx := t.getCtx(batchMerkleRoot)
+	_, span := t.Tracer.Start(
+		ctx,
+		"Quorum reached",
+		trace.WithAttributes(
+			attribute.String("merkle_root", fmt.Sprintf("0x%s", hex.EncodeToString(batchMerkleRoot[:]))),
+			attribute.String("status", "ok"),
+		),
+	) // TODO add quorum %
```
uri-99, line 159: Should this TODO be resolved in this PR?

aggregator/internal/pkg/aggregator.go
uri-99, line 1: What do you think about adding traces around lines 317 of this file? If the aggregator fails to `SendAggregatedResponse` after reaching quorum

telemetry/telemetry.go
JuArce: This will be added in another pr

entropidelic (approved):

JuArce:
Superseded by #1077

PR #997, opened 2024-09-17
fix(audit): Add Fuzzer for cbor serialization in Operator

### LINKED ISSUE #902: Marshall/Unmarshall conflict
There has been observed that some marshalled data does not always match the original data. We are now changing the serialization to CBOR so we can ignore the problem with JSON, but we should check if this problem is present in the marshall/unmarshall of CBOR too, replicating the fuzzer provided for that case.

See `Marshal_Unmarshal_conflict.md` for details

PatStiles:
Adds a fuzzer for cbor serialization within the operator, closes #902 . 

To test on :

macOS
```
make operator_verification_data_fuzz_macos
```
Linux
```
operator_verification_data_fuzz_linux
```


MauroToscano (changes requested):
I'm getting ```no Go files in ../workspace/aligned_layer/operator```. We also need to document this somewhere

PatStiles:
Added make targets for fuzzing.

operator/pkg/marshall_unmarshall_test.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+// Test roundtrip of cbor serialization used in Aligned.
+func FuzzMarshalUnmarshal(f *testing.F) {
+	f.Fuzz(func(t *testing.T, data []byte, seed int64) {
+		// MarshalUnmarshal
+
+		var marshalled VerificationData
+		decoder := codec.NewDecoderBytes(data, new(codec.CborHandle))
+		err := decoder.Decode(marshalled)
```
uri-99: [Decode() must be passed a pointer](https://github.com/ugorji/go/blob/master/codec/decode.go#L1492)
```suggestion
		err := decoder.Decode(&marshalled)
```

uri-99 (changes requested):

uri-99 (changes requested):
Fuzzer is returning an error:

image

This seems to be when Decode() is called with the data parameter: `[]byte("\xa7_[\x00\x00\x10\x000000")`

uri-99:
Was @Oppen debugging the panic on the Decode function? Any news?

uri-99:
Note: same crash happens on Linux

IAvecilla:
After some debugging and research, we decided to try changing the CBOR library we were using. That change seems to have fixed the issue.

I’ve opened PR #1110, which updates the library to the one proposed by @Oppen in #1093 .

MauroToscano (changes requested):
If the data is invalid unmarshall will fatal. Just check the decoding works and doesn't panics. 

We can have another one that starts from a valid piece of data

PatStiles:
> If the data is invalid unmarshall will fatal. Just check the decoding works and doesn't panics.
> 
> We can have another one that starts from a valid piece of data

I split the fuzzer into marshall and unmarshall.

MauroToscano:
Superseeded by: https://github.com/yetanotherco/aligned_layer/pull/1110

PR #1004, opened 2024-09-18
fix(audit): Periodically Remove  oldTask data from aggregator

### LINKED ISSUE #977: fix(aggregator): aggregator memory leak could lead to denial of service
When a new task is created, several aggregator maps are written to with task-specific information. If the service runs for long enough, it could run out of memory, causing the Aggregator to stop functioning. 

Check `aggregator/internal/pkg/aggregator.go`
These are the maps we should clean once the task is responded
```
	agg.batchesIdxByIdentifierHash[batchMerkleRoot] = batchIndex
	agg.batchCreatedBlockByIdx[batchIndex] = uint64(taskCreatedBlock)
	agg.batchesIdentifierHashByIdx[batchIndex] = batchMerkleRoot
	agg.batchDataByIdentifierHash[batchMerkleRoot] = BatchData{
		BatchMerkleRoot: batchMerkleRoot,
		SenderAddress:   [20]byte{},
	}
```

PatStiles:
# This PR:
Removes Task data from aggregator response maps after a response has been responded to or expired.

closes #977 

#To Test:
Setup local dev net as per the guide

```
```

```
 make aggregator_start
```

```
make batcher_start_local
```

(optional:)
```
```

And leave a task sender running until the end of the test
```
```

 Everything should work as normally.

Also, every 120 seconds, you should see the log: `Cleaning finalized tasks from maps` , with some extra information about which tasks are being removed from the Aggregator. These are the tasks being removed, all older than 10 blocks of age.

In prod this variables change to bigger numbers, but the flow is the same.

Note: in dev, first iteration may be small because of the constraints, but the second iteration should include more tasks.


MauroToscano:
What happens if an answer comes after the batch is closed ?

MauroToscano (commented):
What happens if an answer comes after the batch is closed ? It should probably be tested with more operators

PatStiles:
I ran a local testnet with 3 operators produced. The aggregator did not experience any errors and and successfully removed added task information from the aggregator after successfully responding.

aggregator/internal/pkg/aggregator.go
```diff
@@ -282,7 +282,18 @@ func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsA
 			agg.logger.Info("Aggregator successfully responded to task",
 				"taskIndex", blsAggServiceResp.TaskIndex,
 				"batchIdentifierHash", "0x"+hex.EncodeToString(batchIdentifierHash[:]))
-				
+
+			// If Aggregator successfully responds to task we acquire the task mutex and
+			// remove task information from the aggregator maps to prevent a memory leak.
+			agg.taskMutex.Lock()
+			agg.AggregatorConfig.BaseConfig.Logger.Info("- Locked Resources: Removing Task Info from Aggregator")
```
Oppen: The logs should probably be outside the critical section.

entropidelic:
Changed this PR to draft since we should prioritize other issues first

aggregator/internal/pkg/aggregator.go
PatStiles: Addessed, thank you @Oppen

MarcosNicolau:
> What happens if an answer comes after the batch is closed? It should probably be tested with more operators

The aggregator when receiving the response checks that the batch indeed exists [here](https://github.com/yetanotherco/aligned_layer/blob/ea592f4d234a78011ef6916a83b135c94506d718/aggregator/internal/pkg/server.go#L64-L69). Anyway, I tested it on my machine doing:
1. Registered four operators
2. Killed one
3. Sent a bunch of proofs
4. Wait for the aggregator to respond to the task
5. Woke the killed operator and when it got the latest task, it sent his response and the aggregator ignored it, logging the following message:

```bash
[2024-09-27 11:54:10.904 -03] INFO (pkg/server.go:63) - Locked Resources: Starting processing of Response
[2024-09-27 11:54:10.904 -03] INFO (pkg/server.go:67) - Unlocked Resources: Task not found in the internal map
```

MarcosNicolau (approved):
Everything worked as expected on my machine!

JuArce:
Imo the deletion of the task in the maps should be delayed, so it will accept responses from operators after reaching the quorum

It can be done with a go routine like
```go
go func() {
		time.Sleep(10 * time.Second)
		// Deletion logic
	}()
```

The deletion is done once the task was verified in Ethereum (so there is a time between the quorum is reached and the transaction is accepted in Ethereum). So I'm not sure if we should add an extra delay in the deletion

aggregator/internal/pkg/aggregator.go
```diff
@@ -61,7 +61,7 @@ type Aggregator struct {
 	// and can start from zero
 	batchesIdxByIdentifierHash map[[32]byte]uint32
 
-	// Stores the taskCreatedBlock for each batch bt batch index
+	// Stores the taskCreatedBlock for each batch but batch index
```
uri-99:
```suggestion
	// Stores the taskCreatedBlock for each batch by batch index
```

aggregator/internal/pkg/aggregator.go
```diff
@@ -265,6 +265,20 @@ func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsA
 			agg.logger.Info("Aggregator successfully responded to task",
 				"taskIndex", blsAggServiceResp.TaskIndex,
 				"batchIdentifierHash", "0x"+hex.EncodeToString(batchIdentifierHash[:]))
+
+			// If Aggregator successfully responds to task we acquire the task mutex and
+			// remove task information from the aggregator maps to prevent a memory leak.
+			go func() {
```
Oppen: I'm not sure a goroutine is the right tool here. I think doing this sequentially will be more reliable.

aggregator/internal/pkg/aggregator.go
```diff
@@ -278,6 +292,19 @@ func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsA
 		"merkleRoot", "0x"+hex.EncodeToString(batchData.BatchMerkleRoot[:]),
 		"senderAddress", "0x"+hex.EncodeToString(batchData.SenderAddress[:]),
 		"batchIdentifierHash", "0x"+hex.EncodeToString(batchIdentifierHash[:]))
+
+	// If the aggregator fails to respond to the task we remove the batch information as well.
+	go func() {
```
Oppen: Here as well.

Oppen:
> The deletion is done once the task was verified in Ethereum (so there is a time between the quorum is reached and the transaction is accepted in Ethereum). So I'm not sure if we should add an extra delay in the deletion

That sounds to me like what we really want is a kind of queue (jargon for "an array") and a periodic task that checks if we're done with stuff.

EDIT: actually, I think it's simpler than that. We already have indices for each batch. Every once in a while (the appropriate time is probably around the time Ethereum takes to confirm our transactions) we check the oldest (lowest) index, find its Merkle root, and use that and the index to clean the dictionaries. This could be a long-lived goroutine started by main.

aggregator/internal/pkg/aggregator.go
```diff
@@ -202,17 +211,29 @@ func (agg *Aggregator) Start(ctx context.Context) error {
...
+
 func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsAggregationServiceResponse) {
 	if blsAggServiceResp.Err != nil {
 		agg.taskMutex.Lock()
 		batchIdentifierHash := agg.batchesIdentifierHashByIdx[blsAggServiceResp.TaskIndex]
 		agg.logger.Error("BlsAggregationServiceResponse contains an error", "err", blsAggServiceResp.Err, "batchIdentifierHash", hex.EncodeToString(batchIdentifierHash[:]))
+		// Task errored, mark as finalized
+		// TODO: 
+		// Actually, if I remove this task from the maps, Agg will go and fetch it again from chain
+		// agg.batchIsFinalizedChan <- blsAggServiceResp.TaskIndex
```
MauroToscano: This is ready for review but this TODO seems odd

MauroToscano (changes requested):
There is no docs on why a garbage collector is needed. This seems overengineered.

aggregator/internal/pkg/aggregator.go
```diff
@@ -178,6 +178,8 @@ func NewAggregator(aggregatorConfig config.AggregatorConfig) (*Aggregator, error
 		telemetry:             aggregatorTelemetry,
 	}
 
+	go aggregator.clearTasksFromMaps(garbageCollectorPeriod, garbageCollectorTasksAge)
```
Oppen: This doesn't look like the right place to instantiate the task. I would do it at the call-site. Here we're just asking for a struct to be initialized and returned.

uri-99 (approved):
biased self-approve

MauroToscano (approved):

aggregator/internal/pkg/aggregator.go
uri-99: done, with supervisor

aggregator/cmd/main.go
```diff
@@ -38,6 +39,9 @@ func main() {
 	}
 }
 
+const garbageCollectorPeriod = time.Second * 150 //TODO change to time.Day * 1
```
JuArce: When should we change it? If we have to change it, it may be a config variable, so we do not need to change code

aggregator/cmd/main.go
```diff
@@ -38,6 +39,9 @@ func main() {
...
+const garbageCollectorTasksAge = uint64(10)     //TODO change to 2592000, 1 month of blocks
```
JuArce: When should we change it? If we have to change it, it may be a config variable, so we do not need to change code

JuArce: Also `Age` does not tell much more. What does it mean?

aggregator/internal/pkg/aggregator.go
```diff
@@ -377,3 +387,44 @@ func (agg *Aggregator) AddNewTask(batchMerkleRoot [32]byte, senderAddress [20]by
 	agg.AggregatorConfig.BaseConfig.Logger.Info("- Unlocked Resources: Adding new task")
 	agg.logger.Info("New task added", "batchIndex", batchIndex, "batchIdentifierHash", "0x"+hex.EncodeToString(batchIdentifierHash[:]))
 }
+
+// long-lived gorouting that periodically checks and removes old Tasks from stored Maps
+func (agg *Aggregator) ClearTasksFromMaps(period time.Duration, blocksOld uint64) {
+	defer func() {
```
JuArce, line 393: Would be nice to document it. At least a brief explanation of how it works and its motivation

aggregator/internal/pkg/aggregator.go
```diff
@@ -377,3 +387,44 @@ func (agg *Aggregator) AddNewTask(batchMerkleRoot [32]byte, senderAddress [20]by
...
+		if err != nil {
+			agg.logger.Error(err.(string))
+		}
+	}()
+
+	agg.AggregatorConfig.BaseConfig.Logger.Info(fmt.Sprintf("- Removing finalized Task Infos from Maps every %v", period))
+	lastIdxDeleted := uint32(0)
+
+	for {
+		time.Sleep(period)
```
JuArce: Normally we sleep the function after execute the main task, this way we can know easily if the task is being executed correctly (even if it is not deleted anything yet, at least we now it is being executed)

core/chainio/avs_reader.go
```diff
@@ -102,3 +103,48 @@ func (r *AvsReader) GetNotRespondedTasksFrom(fromBlock uint64) ([]servicemanager
 
 	return tasks, nil
 }
+
+// This function is a helper to get a task hash of aproximately nBlocksOld blocks ago
+func (r *AvsReader) GetOldTaskHash(nBlocksOld uint64) (*[32]byte, error) {
+	// r.ChainReader.ethClient.CallContext(context.Background(), &blockNumber, "eth_blockNumber")
```
JuArce: Maybe delete this comment?

JuArce (changes requested):

aggregator/cmd/main.go
uri-99: It means it garbage-collects task of that Age.
In this case, it cleans tasks older than 10 blocks.
I have added a comment description of this.

aggregator/cmd/main.go
uri-99: done

aggregator/cmd/main.go
uri-99: done

aggregator/internal/pkg/aggregator.go
uri-99: done

aggregator/internal/pkg/aggregator.go
uri-99: done

core/chainio/avs_reader.go
uri-99: done

aggregator/internal/pkg/aggregator.go
uri-99: Wait, no, actually we have to execute it at the beggining, because we have `continue` closures that want to break the current execution and wait for next execution

JuArce (approved):

aggregator/cmd/main.go
```diff
@@ -38,6 +38,7 @@ func main() {
 	}
 }
 
+
```
Oppen:
```suggestion
```

aggregator/internal/pkg/aggregator.go
```diff
@@ -377,3 +385,51 @@ func (agg *Aggregator) AddNewTask(batchMerkleRoot [32]byte, senderAddress [20]by
 	agg.AggregatorConfig.BaseConfig.Logger.Info("- Unlocked Resources: Adding new task")
 	agg.logger.Info("New task added", "batchIndex", batchIndex, "batchIdentifierHash", "0x"+hex.EncodeToString(batchIdentifierHash[:]))
 }
+
+// long-lived gorouting that periodically checks and removes old Tasks from stored Maps
```
Oppen:
```suggestion
// Long-lived goroutine that periodically checks and removes old Tasks from stored Maps
```

aggregator/internal/pkg/aggregator.go
```diff
@@ -377,3 +385,51 @@ func (agg *Aggregator) AddNewTask(batchMerkleRoot [32]byte, senderAddress [20]by
...
 }
+
+// Long-lived goroutine that periodically checks and removes old Tasks from stored Maps
+// It runs every period and removes all tasks older than blocksOld
+// This was added because each task occupies memory in the maps, and we need to free it to avoid a memory leak
+func (agg *Aggregator) ClearTasksFromMaps(period time.Duration, blocksOld uint64) {
+	defer func() {
+		err := recover() //stops panics
+		if err != nil {
+			agg.logger.Error(err.(string))
```
Oppen: Casting to string may fail and `panic` for good, but `error` has an `Error()` method to format as string:
```suggestion
			agg.logger.Error(err.Error())
```

config-files/config-aggregator.yaml
```diff
@@ -35,6 +35,9 @@ aggregator:
   enable_metrics: true
   metrics_ip_port_address: localhost:9091
   telemetry_ip_port_address: localhost:4001
+  garbage_collector_period: 60s #The period of the GC process #TODO make it 1 day
+  garbage_collector_tasks_age: 3 #The age of tasks that will be removed by the GC #TODO make it 1 month
```
Oppen: Missing TODOs. Will we do them now or leave for admin to do?

Oppen (approved):

config-files/config-aggregator.yaml
uri-99: Left recommended values for devops to set on prod

Issue #999, opened 2024-09-18
fix: Batcher sends proof even if eip712 signature contents are incompatible DoS


uri-99:
EIP712 signature has content such as chainId and contract address. If this information is not appropriate (i.e. chainId is mainnet, but sent on Holesky), even if the signature is appropriate, BatcherPaymentsService will reject it, as it should.

The problem is, Batcher will accept this information, and send it to BatcherPaymentsService, resulting in a reverted tx. The Batcher should not accept this proof to avoid the whole batch being dropped.

PR #1005, opened 2024-09-18
fix: batcher sends proof even if eip712 signature contents are incompatible

uri-99:
# This PR
There was a bug that the batcher accepted a proof if incompatible `--chain` and `--batcher-payments-address` where provided, which resulted in an Ethereum tx revert for the whole batch submition.

Resolving this, it was found that these 2 parameters where redundant, so they were both removed and replaced for the `--network` , which accepts either `devnet`, `holesky-stage` or `holesky` .

Then, this parameter is used by the SDK to the appropriate contract addresses, while the chainId is resolved from the RPC (this has not changed, which also lead to confusion).

## To Deploy
- As the SDK changes with this PR, a new release with a new tag should be made (maybe v0.7.1 or v0.8.0)
- The Batcher should be redeployed, since a new check is being made
  - `if client_msg.verification_data.payment_service_addr != self.payment_service.address() {`
- The `task sender` should be restarted using this new parameter (make targets have been migrated)
- ZKQuiz contract has also changed so it must be redeployed and its address repointed to the new on-chain one

## To Test
You can test the following work correctly:
- the new task senders
- the batcher
- the examples
  - Validating Public Inputs can be tested on testnet, as normal (`make submit_fibonacci_proof`)
  - ZKQuiz changed a value in the contract so it should be tested on devnet before deploying it on Staging

### Test ZKQuiz in devnet
1) Run anvil
```
```

2) Deploy ZKQuiz verifier contract:
```
cd examples/zkquiz
make deploy_verifier
```

3) Configure variables in `examples/zkquiz/contracts/.env`:
```
RPC_URL=http://localhost:8545
PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
ALIGNED_SERVICE_MANAGER_ADDRESS=0x1613beB3B2C4f22Ee086B2b38C1476A3cE7f78E8
BATCHER_PAYMENT_SERVICE_ADDRESS=0x7969c5eD335650692Bc04293B07F5BF2e7A673C0
```

4) And configure variables in `examples/zkquiz/Makefile/.env`: (double check `CONTRACT_ADDRESS` is the same address as outputed when deployed the zkquiz verifier contract)
```
CONTRACT_ADDRESS=0xc351628EB244ec633d5f21fBD6621e1a683B1181
RPC_URL=http://localhost:8545
KEYSTORE_PATH=../../../../config-files/devnet/keys/anvil.ecdsa.key.json
```

5) Change some code in `examples/zkquiz/quiz/script/src/main.rs`"
- line 20, to:
```
const BATCHER_URL: &str = "ws://localhost:8080";
```
- line 23, to:
```
const NETWORK: Network = Network::Devnet;
```
- line 33, to:
```
        default_value = "http://localhost:8545"
```
- line 49, to:
```
.with_chain_id(31337u64);
```

6) Run zkquiz:
```
make answer_quiz
```

Solves https://github.com/yetanotherco/aligned_layer/issues/999


examples/validating-public-input/aligned-integration/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-aligned-sdk = { git = "https://github.com/yetanotherco/aligned_layer", tag = "v0.6.0" }
+aligned-sdk = { git = "https://github.com/yetanotherco/aligned_layer", rev = "5ac0862ebace4e5971d1fd4c58f2c1973576d557" }
```
entropidelic: should we leave this version commit?

uri-99: The version should be the current one. Should I point it to the current branch? If i point it to an existing release, it won't contain the changes made in this PR

batcher/aligned-sdk/src/core/types.rs
```diff
@@ -324,7 +325,7 @@ pub enum ResponseMessage {
 }
 
 #[derive(Debug, Clone)]
-pub enum Chain {
+pub enum Network {
```
entropidelic, line 333: very small nit, but we can derive the `Copy` trait here.
Doing this, the references in the function arguments like `submit_multiple_and_wait_verification` and the others are not needed.
Since it is just an enum, it is very cheap to copy the value (actually cheaper than the reference itself) and the code will look a little cleaner.

examples/validating-public-input/aligned-integration/Cargo.toml
entropidelic: yes you are right. Let's make it point to the last commit of this branch, but we have to remember to update it once we make a release

MauroToscano: Why is it not using the repository version ? I assume we may want to keep them working while operators upgrade maybe ?

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -114,6 +114,10 @@ pub async fn send_messages(
                 error!("Invalid replacement message!");
                 return Err(SubmitError::InvalidReplacementMessage);
             }
+            ValidityResponseMessage::InvalidPaymentServiceAddress => {
+                error!("Invalid payment service address!");
+                return Err(SubmitError::InvalidPaymentServiceAddress);
+            }
```
entropidelic: for debugging, it may be useful to know which was the expected payment service address and which one the received in the message

entropidelic:
aside from some nits, code looks good and worked correctly

batcher/aligned-sdk/src/core/types.rs
uri-99: refactored 👍

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: implemented 👍

examples/validating-public-input/aligned-integration/Cargo.toml
uri-99: The actual content of the requests does not change so the examples should be usable even before the SDK is released, right?

MauroToscano (changes requested):
Review how to deploy new version of examples with @JuArce

examples/validating-public-input/aligned-integration/Cargo.toml
uri-99: image

uri-99:
Opening back to ready 4 review after changed cargo.toml to local dir, and bringing Pats pr

batcher/aligned-batcher/src/lib.rs
```diff
@@ -441,7 +441,7 @@ impl Batcher {
 
         if client_msg.verification_data.chain_id != self.chain_id {
             warn!(
-                "Received message with incorrect chain id: {}",
+                "Received message with incorrect chain id: {}", //This check does not save against "Holesky" and "HoleskyStage", since both are chain_id 17000
```
Oppen, line 444: Is there any other check that can cover the difference?

batcher/aligned-sdk/src/sdk.rs
```diff
@@ -153,10 +146,30 @@ pub async fn submit_multiple(
...
+
+pub fn get_aligned_service_manager_address(network: Network) -> ethers::types::H160 {
+    match network {
+        Network::Devnet => H160::from_str("0x1613beB3B2C4f22Ee086B2b38C1476A3cE7f78E8").unwrap(),
+        Network::Holesky => H160::from_str("0x58F280BeBE9B34c9939C3C39e0890C81f163B623").unwrap(),
+        Network::HoleskyStage => {
+            H160::from_str("0x9C5231FC88059C086Ea95712d105A2026048c39B").unwrap()
+        }
+    }
+}
```
Oppen, line 167: These addresses should probably not be hardcoded into the program.

batcher/aligned-batcher/src/lib.rs
uri-99: yes, on line `459`: `This checks saves against "Holesky" and "HoleskyStage", since each one has a different payment service address`.
So with those 2 checks, the batcher is certain the user signed and sent the message with the expected chainId and Network

batcher/aligned-sdk/src/sdk.rs
uri-99: They are coded into the SDK, not the actual Batcher program.

This allows to ask the SDK directly for Aligned contract's addresses, so program implementations just need to pass in which `Network` they want to communicate.

The improvement is that before, these value were all around the place, expecting them in a parameter but then rewriting them according to other parameter values. (`payment_service_addr` and `chain`). Unifying this parameter (now called `Network`) avoids the possibility of sending uncompatible `payment_service_addr` and `chain` on the same request.

If the addresses where removed from the SDK, users would have to pass all relevant addresses as parameters, which would re-enable the possibility of sending uncompatible parameters on a request.

If necesarry, this could be done, but we should be extra careful with the error codes since uncompatible addresses might fail in unexpected ways, and error the error codes might not be obviously correlated to uncompatible addresses.

Oppen: I meant more that maybe they should be part of some config file. But I don't think that's really important at this time, so feel free to ignore the comment.

examples/zkquiz/quiz/script/src/main.rs
```diff
@@ -106,21 +108,20 @@ async fn main() {
                 pub_input: None,
             };
 
-            // Set a `max_fee` of 0.5 Eth
+            // Set a `max_fee` of 0.1 Eth
```
JuArce: This is wrong. The max_fee here is 0.5 eth

entropidelic (approved):

JuArce (approved):

MauroToscano (approved):

PR #1008, opened 2024-09-18
refactor(batcher): batch building algorithm

entropidelic:
This PR changes the batch building algorithm.
The implementation in this PR now starts iterating from the batch queue entry with the minumum max fee, and asks if the fee per proof obtained by posting the whole batch is less than that max fee of that entry. If it is, that means that every user is willing to pay for the price and the whole batch is posted. If it is not, then it is saved for another batch and we continue with the next entry, repeating the process, calculating the fee per proof with a proof less in the queue.

### Testing
* Start and test everything as usual. Make sure it works as expected.
* Some tests have been added for the new algorithm. Test them manually and check that they work. A CI workflow should be added later for them to run automatically. See issue #1047 
```bash
cd batcher
cargo test
```

**NOTE**: the `try_build_batch` function needs to calculate the batch size using cbor serialization. Until now, if there was an error in the serialization, the error was ignored. This is very unlikely to happen because the entries had been already serialized and deserialized in the moment `try_build_batch` is called, unless there is something really bad with the cbor serialization library.
Nevertheless, I changed the `try_build_batch` function to return a `Result` so that the error is more visible, but I didn't refactor the rest of the code to accomodate to the new return type, since it would be a little out of scope for this PR. issue #1046 has been done to address this and can be changed in a subsequent refactor of the code.


PatStiles (approved):
Unit tests passes and running with multiple operators did not cause any errors.

batcher/aligned-batcher/src/types/batch_queue.rs
```diff
@@ -87,9 +110,428 @@ impl Ord for BatchQueueEntryPriority {
         if ord == std::cmp::Ordering::Equal {
             self.nonce.cmp(&other.nonce).reverse()
         } else {
-            ord
+            ord.reverse()
```
MauroToscano: Cmp function can be changed instead of using a reverse here

batcher/aligned-batcher/src/types/batch_queue.rs
```diff
@@ -87,9 +110,428 @@ impl Ord for BatchQueueEntryPriority {
...
+
+            // It is safe to call `.unwrap()` here since any serialization error should have been caught
+            // when calculating the total size of the batch with the `calculate_batch_size` function
+            let verification_data_size =
+                cbor_serialize(&entry.nonced_verification_data.verification_data)
+                    .unwrap()
+                    .len();
+            batch_size -= verification_data_size;
+
+            let (not_working_entry, not_woring_priority) = batch_queue_copy.pop().unwrap();
```
MauroToscano: Typo, not_woring_priority

batcher/aligned-batcher/src/types/batch_queue.rs
```diff
@@ -87,9 +110,428 @@ impl Ord for BatchQueueEntryPriority {
...
+/// of proofs from users that are willing to pay for it:
+/// 1. Traverse each batch priority queue, starting from the one with minimum max fee.
+/// 2. Calculate the `fee_per_proof` for the whole batch and compare with the `max_fee` of the entry.
+/// 3. If `fee_per_proof` is less than the `max_fee` of the current entry, submit the batch. If not, pop this entry
+///     from the queue and push it to `resulting_priority_queue`, then repeat step 1.
+///
+/// `resulting_priority_queue` will be the batch queue composed of all entries that were not willing to pay for the batch.
+/// This is outputted in along with the finalized batch.
+pub(crate) fn try_build_batch(
+    batch_queue_copy: &mut BatchQueue,
```
MauroToscano: If it needs to be a copy, either don't use a reference, or use a non mutable reference and clone inside the function, to avoid mistakes

entropidelic: solved in 31c9ae2877f7746a868e029fb0a2c50df68181d6

batcher/aligned-batcher/src/types/batch_queue.rs
entropidelic: solved in 75cdc89fd644906855f13a890c4867095acfc5b6

MauroToscano (approved):

PR #1014, opened 2024-09-19
fix(code quality): Improve error for S3 upload in Batcher

### LINKED ISSUE #1011: Error when uploading to bucket fail that is not clear ```Task creation error: dispatch failure```
There is a dispatch error when uploading to bucket fail that is not clear ```Task creation error: dispatch failure```

It's not clear what a dispatch is, it should be easier to understand the problem was uploading the data

PatStiles:
Improves the error for uploading batches to S3 in the batcher.

closes #1011

### DIFF
--- a/batcher/aligned-batcher/src/lib.rs
+++ b/batcher/aligned-batcher/src/lib.rs
@@ -1072,7 +1072,7 @@ impl Batcher {
             &file_name,
         )
         .await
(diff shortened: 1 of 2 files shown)


MauroToscano (changes requested):
How does this error get printed ? Also, maybe change S3BatchUploadError to BatchUploadError, since we are not using S3

MauroToscano (approved):

uri-99 (approved):

PR #1040, opened 2024-09-20
feat(testing): Port infinite task sender to rust.

### LINKED ISSUE #1015: create rust program with the SDK to mimic infinite proof submission scripts
PR: #1040 

* sleep and repeat

Some details:

PatStiles:
Closes #1015 creates a rust cli tool to generate and submit infinite task to the network to replace the current shell scripting framework. 

Suggestions:
- Currently the tool executes a go program from the cli to generate a gnark proof that in turn depends on specific folders to write the generated proofs. This is pretty fragile and we could eliminate this external dependency by using rust -> go ffi bindings via [rust-bindgen] to pass the proof data directly at during execution without writing and reading from files.

## TODO:
- [x] Test locally.
- [x] Test on Holesky with local Keystore.

## To Test:
```
make deps
```
```
```
- Run anvil, aggregator and batcher:
```
```
```
make aggregator_start
```
```
make batcher_start_local
```
- Run the aligned-task-sender by either running:
```
cd batcher/aligned-task-sender/ && cargo run --release -- send-infinite
```

```
make batcher_send_infinite_burst
```


batcher/aligned/src/main.rs
```diff
@@ -726,7 +726,7 @@ fn save_response(
     Ok(())
 }
 
-pub async fn get_user_balance(
+async fn get_user_balance(
```
uri-99, line 699: why this change?

batcher/aligned-task-sender/src/main.rs
```diff
@@ -0,0 +1,383 @@
...
+    let mut nonce_bytes = [0; 32];
+
+    (nonce + U256::from(proof_count)).to_big_endian(&mut nonce_bytes);
+
+    write_file(nonce_file.as_str(), &nonce_bytes)?;
+
+    Ok(nonce)
+}
+
+//TODO: Duplicated from `batcher/aligned`.
```
uri-99: solve TODO

batcher/aligned-task-sender/src/main.rs
```diff
@@ -0,0 +1,383 @@
...
+                    SubmitError::InvalidEthereumAddress(format!(
+                        "Error while parsing address: {}",
+                        e
+                    ))
+                })?;
+            std::fs::create_dir_all(GROTH_16_PROOF_DIR)
+                .map_err(|e| SubmitError::IoError(PathBuf::from(GROTH_16_PROOF_DIR), e))?;
+
+            let mut count = 1;
+            //TODO: sort out error messages
```
uri-99: solve TODO

batcher/aligned-task-sender/src/main.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+impl From<ChainArg> for Chain {
+    fn from(chain_arg: ChainArg) -> Self {
+        match chain_arg {
+            ChainArg::Devnet => Chain::Devnet,
+            ChainArg::Holesky => Chain::Holesky,
+            ChainArg::HoleskyStage => Chain::HoleskyStage,
+        }
+    }
```
uri-99: this will change after #1005 is merged

uri-99 (changes requested):
Temporarily blocking PR, leaving some comments and also it should be merged after #1005

batcher/aligned-task-sender/src/main.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<(), AlignedError> {
+    env_logger::Builder::from_env(Env::default().default_filter_or("info")).init();
+    let base_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    let args: TaskSenderArgs = TaskSenderArgs::parse();
+
+    match args.command {
+        SendInfinite(submit_args) => {
```
uri-99: We are planning on adding some more variations, right? we could add them as a comment.
If not, having a switch here doesn't make much sense

batcher/aligned-task-sender/src/main.rs
```diff
@@ -0,0 +1,383 @@
...
+                            e.to_string()
+                        );
+                        return Ok(());
+                    }
+                }
+            };
+
+            let eth_rpc_url = submit_args.eth_rpc_url.clone();
+
+            let chain_id = get_chain_id(eth_rpc_url.as_str()).await.map_err(|e| {
```
uri-99: It would be cleaner to use `get_chain_id` from the SDK

batcher/aligned-task-sender/src/main.rs
```diff
@@ -0,0 +1,383 @@
...
+    #[clap(about = "Send Infinite Proofs to the batcher")]
+    SendInfinite(SubmitArgs),
+}
+
+#[derive(Parser, Debug, Clone)]
+#[command(version, about, long_about = None)]
+pub struct SubmitArgs {
+    #[arg(
+        name = "Batcher connection address",
+        long = "batcher_url",
```
uri-99: we have been using `kebab-case` for parameters

PatStiles:
Will close and reopen

PR #1041, opened 2024-09-20
fix: build `hashStruct` following EIP description and do not perform keccak on `nonce` and `maxFee`

JuArce:
# From Audit

Right over here where you construct the EIP712 hashed struct to be verified against the signature, you perform keccak256 on the nonce and the maxFee:

https://github.com/yetanotherco/aligned_layer/blob/ce5c9e76002365d23b5e23c9a4ebd9c930e3b7d7/contracts/src/core/BatcherPaymentService.sol#L283-L294

But you shouldn’t do that. In fact, the standard expects you to encode them directly [(linked here)](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#definition-of-encodedata)

For dynamic-size objects like bytes and string, and arrays we should do keccak256, but for other things we just do provide it directly. So it should be something like keccak256(abi.encode(typedHash, leaf, signature.nonce, signature.maxFee))

# Description

Change structHash using the [EIP description](https://eips.ethereum.org/EIPS/eip-712#definition-of-hashstruct)

```
hashStruct(s : 𝕊) = keccak256(typeHash ‖ encodeData(s))
```

Also, do not perfom the keccak on the nonce and the maxFee

# How to test

Send proofs as usual

> [!Note]
> Remember to rebuild aligned using `make install_aligned_compiling`

# How to deploy

This PR requires upgrading the BatcherPaymentService.sol

To upgrade you need to run
```sh
make upgrade_batcher_payment_service
```

Once you upgraded the contrat, you need to rebuild the batcher and the client

```
make install_batcher
```

```
```


PatStiles (approved):

uri-99 (approved):

uri-99 (approved):
Reverting my approve until tested with MM

uri-99 (changes requested):
Reverting my approve until tested with MM

uri-99 (approved):
It is still working on Metamask

PatStiles:
Confirmed with @uri-99 script this still functions on metamask

Issue #964, opened 2024-09-13
centralize constants in one file


entropidelic:
We have constants scattered along many files. It will be useful and make the code much cleaner if we had constants in one file and read them from there.
If one file is too much work, we could have at least one per entity e.g., batcher, operator, aggregator, etc.
Some example constants:
- Rust gas costs
- contract addresses in make
- other number constants in make
- contract addresses to deploy batcherPayments
- retries and wait times

PR #1045, opened 2024-09-23
feat(utils): Utility functions for computing and estimating the `max_fee` of a proof

PatStiles:
This pr adds three functions to `aligned_sdk`: 
- `estimate_max_fee`: Returns to the user there estimated `max_fee`: Min, Default, Instant based on a theoretical batch of 32 proofs. 32 was decided upon as a suggested arbitrary large batch size.
- `fee_per_proof`: Returns the `fee_per_proof` of a batch size of size `num_proofs_per_batch`
- `compute_max_feee`: returns the computing `max_fee` for a proof based on (`num_proofs` / `num_proofs_per_batch`).

It deduplicates constants originally found in `aligned-batcher` that can be imported once #964 is completed.


MauroToscano (changes requested):
Decouple calculation from the fetching of the GasCost, and add tests assuming you have the GasCost. It can be just checking that one is bigger than another one

PatStiles:
@MauroToscano tests added

MauroToscano (changes requested):
Mention this function in the one that is used to send a task, with ///, so the user knows this one should be used to get the right estimation

batcher/aligned-sdk/src/sdk.rs
```diff
@@ -35,6 +35,18 @@ use futures_util::{
     StreamExt, TryStreamExt,
 };
 
+//Public constants for convenience
+pub const HOLESKY_PUBLIC_RPC_URL: &str = "https://ethereum-holesky-rpc.publicnode.com";
```
entropidelic: this is used only for testing, it should be in the testing section at the end of the file

batcher/aligned-sdk/src/sdk.rs
```diff
@@ -35,6 +35,18 @@ use futures_util::{
...
+
+const AGGREGATOR_GAS_COST: u128 = 400_000;
+const ADDITIONAL_SUBMISSION_GAS_COST_PER_PROOF: u128 = 13_000;
+const CONSTANT_GAS_COST: u128 = ((AGGREGATOR_GAS_COST * DEFAULT_AGGREGATOR_FEE_MULTIPLIER)
+    / DEFAULT_AGGREGATOR_FEE_DIVIDER)
+    + BATCHER_SUBMISSION_BASE_GAS_COST;
+const DEFAULT_AGGREGATOR_FEE_MULTIPLIER: u128 = 3; // to set the feeForAggregator variable higher than what was calculated
+const DEFAULT_AGGREGATOR_FEE_DIVIDER: u128 = 2;
+const BATCHER_SUBMISSION_BASE_GAS_COST: u128 = 125_000;
```
entropidelic: duplication of these constants make the code difficult to maintain. If here is a good place for them to be held until we centralize all constants into some file, then remove them from other places and make them be imported from here

batcher/aligned-sdk/src/sdk.rs
```diff
@@ -99,6 +111,103 @@ pub async fn submit_multiple_and_wait_verification(
     Ok(aligned_verification_data)
 }
 
+/*
+Returns the estimated `max_fee` depending on the batch inclusion preference of the user, based on the max priority gas price.
+NOTE: The `max_fee` is computed from an rpc nodes max priority gas price.
+
+For batch cost estimates we estimate a batch will have ~32 proofs present.
```
Oppen: Batch size sounds like the kind of parameter you might want to tune eventually. It should probably be a global constant instead of a magic number, so we can change it in one place.

batcher/aligned-sdk/src/sdk.rs
```diff
@@ -99,6 +111,103 @@ pub async fn submit_multiple_and_wait_verification(
...
+                "Failed to compute `fee_per_proof`: {:?}",
+                e
+            )))
+        }
+    };
+
+    let proof_price = match estimate {
+        PriceEstimate::Min => fee_per_proof,
+        PriceEstimate::Default => U256::from(10) * fee_per_proof,
+        PriceEstimate::Instant => U256::from(32) * fee_per_proof,
```
Oppen: I will guess: this is the batch size, because "instant" means "close the batch now, even if this is the only proof, I'll pay for the whole batch if necessary", right?

batcher/aligned-sdk/src/sdk.rs
```diff
@@ -704,4 +813,44 @@ mod test {
...
+    }
+
+    #[tokio::test]
+    async fn computed_max_fee_for_more_proofs_larger_than_for_less_proofs() {
+        let small_fee = compute_max_fee(HOLESKY_PUBLIC_RPC_URL, 5, 20)
+            .await
+            .unwrap();
+        let large_fee = compute_max_fee(HOLESKY_PUBLIC_RPC_URL, 5, 10)
+            .await
+            .unwrap();
```
Oppen, line 602: I'm worried these tests may be flaky, for the following reasons:
1. Querying the API means it can change with time, due to congestion, so at some points in time where prices lower `large_fee` may become cheaper than `small_fee`;
2. We may end up hitting rate limits.

batcher/aligned-sdk/src/sdk.rs
entropidelic: some docs explaining these multipliers would be nice

batcher/aligned-sdk/src/sdk.rs
```diff
@@ -99,6 +111,103 @@ pub async fn submit_multiple_and_wait_verification(
...
+        Err(e) => {
+            return Err(SubmitError::GenericError(format!(
+                "Failed to fetch gas price: {:?}",
+                e
+            )))
+        }
+    };
+
+    Ok(gas_price)
+}
```
entropidelic: I think it doesn't make much sense to define a function on our SDK just to call the rpc provider's `get_gas_price` method. Maybe a better pattern will be to instantiate a provider only once, and ask for the gas price directly to it. This data can be used directly in the other functions without the need need to instantiate a provider every time. 
@Oppen if you have any thoughts on this please feel free to share them

PatStiles: Decoupling fetching the gas cost was done in response to this comment above https://github.com/yetanotherco/aligned_layer/pull/1045#pullrequestreview-2326200196 in https://github.com/yetanotherco/aligned_layer/pull/1045/commits/514ca8d0602e19c2250a60866afba232851377e2 . Happy to revert the change if that is the consensus.

batcher/aligned-sdk/src/sdk.rs
PatStiles: Extracted to constants

batcher/aligned-sdk/src/sdk.rs
PatStiles: Now passes provider

PR #1052, opened 2024-09-23
Add consensus state input checks

gabrielbosio:


gabrielbosio:
Wrong parent branch. Closing.

PR #1064, opened 2024-09-24
refactor: remove 0 check in recovered address in BatcherPaymentService

### LINKED ISSUE #1061: Remove the 0 address check for the recovered signer as it is unnecessary

MarcosNicolau:
**Changes**
This PR removes the check for zero addresses in the recovered signature within the `BatcherPaymentService` contract. The library handling the signature recovery already performs this check when splitting the fields (`r`, `v`, `s`) first and then calling the `ecrecover` precompile, where it does the check (see [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/dbb6104ce834628e473d2173bbc9d47f81a9eec3/contracts/utils/cryptography/ECDSA.sol#L142-L145)). 

**Testing**
I've tested that everything works by setting up a local environment and sending some `sp1` and `plonk` proofs. Here is the process:

1. Start aggregator: `make aggregator_start`.
2. Start operator: `make operator_register_and_start`.
3. Start batcher: `make batcher_start_local`.
4. Send some proofs and make sure everything works as expected.

Contracts have been updated with `make anvil_deploy_aligned_contracts`.

Closes #1061


entropidelic (approved):

MauroToscano (changes requested):
Re-deploy aligned contracts and test it, and push them

uri-99 (approved):

MauroToscano (approved):

PR #1066, opened 2024-09-24
refactor: batcher non-paying messages

### LINKED ISSUE #1006: refactor: remove batcher duplicated code when handling nonpaying
PR: #1066 

The batcher has a poor implementation with repeated code around [here](https://github.com/yetanotherco/aligned_layer/blob/staging/batcher/aligned-batcher/src/lib.rs#L458).

In here, batcher is divided in 2 cases;
```
if self.is_nonpaying(&addr) {
     ....
} else {
     ....
}
```
Both logics are quire similar, and the code inside the function `handle_nonpaying_msg()` implements almost the same logic that is implemented after [the else](https://github.com/yetanotherco/aligned_layer/blob/staging/batcher/aligned-batcher/src/lib.rs#L461).

Appropriate functions should be made to handle this and avoid the code repetition.

MarcosNicolau:
**Changes**
Refactors the `handle_message` function in the `batcher`, unifying the `non_paying` case.

**Test**
These new changes can be tested by setting up a local environment and: 
1. Sending proofs to the `non-paying` address and verifying that the batcher pays for us.
2. Sending proofs to a custom address and verifying that we paid for the verification.

Closes #1006

### DIFF
--- a/batcher/aligned-batcher/src/lib.rs
+++ b/batcher/aligned-batcher/src/lib.rs
@@ -452,7 +452,7 @@ impl Batcher {
         }
 
         info!("Verifying message signature...");
-        let Ok(addr) = client_msg.verify_signature() else {
+        let Ok(mut addr) = client_msg.verify_signature() else {
             error!("Signature verification error");
             send_message(
                 ws_conn_sink.clone(),
@@ -466,12 +466,18 @@ impl Batcher {
         let is_non_paying = self.is_nonpaying(&addr);
         if is_non_paying {
             debug!("Non paying message");
+            // if the message is nonpaying we want to use the configured batcher address
+            // to pay for the verification
+            addr = self
+                .non_paying_config
+                .as_ref()
+                .unwrap()
+                .replacement
+                .address();
         }
 
         debug!("Verifying balance");
         if is_non_paying {
-            let non_paying_config = self.non_paying_config.as_ref().unwrap();
-            let addr = non_paying_config.replacement.address();
             let user_balance = self.get_user_balance(&addr).await;
 
             if user_balance == U256::from(0) {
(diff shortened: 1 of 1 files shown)


batcher/aligned-batcher/src/lib.rs
```diff
@@ -453,146 +452,168 @@ impl Batcher {
...
 
-                // When pre-verification is enabled, batcher will verify proofs for faster feedback with clients
-                if self.pre_verification_is_enabled
-                    && !zk_utils::verify(&nonced_verification_data.verification_data).await
-                {
-                    error!("Invalid proof detected. Verification failed.");
-                    send_message(ws_conn_sink.clone(), ValidityResponseMessage::InvalidProof).await;
-                    return Ok(()); // Send error message to the client and return
-                }
+            if user_balance == U256::from(0) {
```
MarcosNicolau, line 483: I don't understand why here we check if the balance is zero instead of checking for the actual cost of the verification.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -732,6 +753,47 @@ impl Batcher {
...
+    /// Handles a nonpaying message
+    /// this types of message are used in testing environments
+    /// here the batcher pays for verifying the proof for the user
+    async fn handle_nonpaying_message(
+        &self,
+        mut batch_state: tokio::sync::MutexGuard<'_, BatchState>,
+        nonced_verification_data: NoncedVerificationData,
+        ws_conn_sink: Arc<RwLock<SplitSink<WebSocketStream<TcpStream>, Message>>>,
+        nonce_value: U256,
+    ) -> bool {
```
MarcosNicolau, line 771: This function can never fail, but the return of the boolean is just to keep the consistency with the other methods and to account for a future failing point. Please tell me if I should better remove it.

batcher/aligned-batcher/src/lib.rs
Oppen: I think this is because it's a non-paying message, so there isn't any particular balance the user needs to reach, as long as it's not empty. If that's the case, it's worth a code comment.

uri-99 (changes requested):
Blocking PR until Entropidelic's refactor is pushed

uri-99:
Closing as solved in #1106

Issue #962, opened 2024-09-13
fix(operator, aggregator): aggregator or operator downtime can break eventual consistency


entropidelic:
## Tasks

- [ ] Solve issue on Operator
- [ ] Solve issue on Aggregator

## Operator
If an Operator experiences downtime, NewBatch events emitted when the Operator was offline will not get processed by the Operator. If over one-third of the Operators experience downtime and subsequently miss a NewBatch event (e.g., due to a poorly rolled out upgrade), it will be impossible for the Aggregator to accumulate a quorum of verifications for the corresponding batch.

### Recommended remediation
To address Operator downtime, we recommend that an Operator log timestamps to a file periodically. When the Operator is back online after downtime, it could read the file to learn when it went offline and process all the batches since the downtime that do not already have a corresponding BatchVerified event.
Additionally, we recommend having the Operator write to a log any batch it has previously verified and skip re-verifying those batches after downtime, which the Aggregator has already verified, but which have not yet triggered a BatchVerified event by the smart contract either because a quorum has not yet been reached, or because the respondToTask transaction is still in the mempool

## Aggregator
If the Aggregator were to experience downtime, it could similarly cause batches to be skipped. If an Operator's message fails to reach the Aggregator, the Operator will continue to attempt to reach the Aggregator for 100 seconds. If the Aggregator fails to become available within this time, data loss will occur.

### Recommended remediation
o address Aggregator downtime, we recommend solving the Operator-to-Aggregator message loss by
placing a highly available queue in front of the Aggregator, such as Amazon SQS. However, Operator
signatures accumulated by the Aggregator that had not yet reached a quorum would be lost in an
Aggregator crash event. To remediate this, Operator signatures could be logged to a file, which could be
read on service startup. However, if an Aggregator that is experiencing issues is replaced by another
Aggregator, these accumulated messages would not be available, as the log file would be on a separate
instance. A shared datastore, such as Redis, would be required to facilitate node replacement. For
catching up on missed events from AlignedLayerServiceManage, the timestamp logging solution
should suffice.

Oppen:
> To address Operator downtime, we recommend that an Operator log timestamps to a file periodically. When the Operator is back online after downtime, it could read the file to learn when it went offline and process all the batches since the downtime that do not already have a corresponding BatchVerified event.
[...]

PR #1073, opened 2024-09-24
fix: operator downtime

### LINKED ISSUE #978: fix: getLatestTaskFromEthereum should use filterer
Currently, getLatestTaskFromEthereum in `avs_subscriber.go` , parses an ABI to make a client request of a NewBatch log. It says:
```
The alternative to this is using the FilterNewBatch method from the contract's filterer, but it requires
to iterate over all the logs, which is not efficient and not needed since we only need the latest task.
```

Which is not true, since bindings filterer accepts, in `filteropts, the `fromBlock` and `toBlock` parameters.

This is why, `getLatestTaskFromEthereum` should use the bindings filterer. Using them, errors here appear on compilation and not on runtime (which was an error we had in the operators, where `NewBatch` did not exist)

MarcosNicolau:
**Changes**
This pr adds a way to recover the operator's tasks from downtime. The way it works now is:
1. We store the latest block at which a batch was finished processing.
2. On startup we read the block and query the `avs_reader` for all the tasks that have not been responded to starting from that block minus `100` (this is to account for the case that a higher block number task finished first than a previous one, since batches can be processed in parallel). This is to be improved with an algorithm by @Oppen.
3. We process the batches as per usual.

A new check is also added to not send already responded blocks in the `avs_subscriber` 'get_latest_task` interval. This problem happened at startup specially, it would always run the latest task doesn't matter how old it was.

**Test**

To make sure everything works, we can run the following test: setup

1. Setup a local testnet as usual: `make anvil_start_with_block`, `make aggregator_start`, `make batcher_start_local`
5. Now for the operators, we'll start four different ones:
    - `1`: `make operator_register_and_start`
    - `2`: `make operator_register_and_start CONFIG_FILE=config-files/config-operator-1.yml`
    - `3`: `make operator_register_and_start CONFIG_FILE=config-files/config-operator-2.yml`
    - `4`: `make operator_register_and_start CONFIG_FILE=config-files/config-operator-3.yml`
7. Send some proofs to the batcher to make sure everything is working: `make batcher_send_plonk_bls12_381_burst `
8. Stop operator `2`, `3` and `4`
9. Again send more proofs : `make batcher_send_risc0_burst` (it is better if you send many batches)
10. You should see that aggregator is waiting for more verifications to send the response
11. Start operator `2` and `3` again: 
      - `make operator_start CONFIG_FILE=config-files/config-operator-1.yml`
      - `make operator_start CONFIG_FILE=config-files/config-operator-2.yml`
13. You should see quorum is reached and the aggregator sends the response
14. Start operator `4`: `make operator_start CONFIG_FILE=config-files/config-operator-3.yml`
15. You should see that it doesn't in fact perform the tasks as the batches have already been responded to.

## Deploy 

> [!WARNING]
> This PR adds the following parameter to the operator configuration `last_processed_batch_filepath: '<>'`

Advances on #962
Closes #978


operator/pkg/operator.go
```diff
@@ -181,6 +238,29 @@ func (o *Operator) Start(ctx context.Context) error {
 	}
 }
 
+// Here we query all the batches that have not yet been verified starting from
+// the latest verified batch by the operator. We also get the prior 5 and check if we need to verify them as well
+// This last thing of getting the last 5 is to make sure we have not missed a batch since they are process in parallel
+// and a higher batch number might have been processed first than the lower one.
+// So getting the last five accounts for that case
```
Oppen: I'm not sure the magic 5 solves the issue if they run in parallel. Imagine a given batch, say batch `3` takes longer than the others, then the others threads advance through blocks `4`, `5`, `6`, up to, say, `10`. Then you'll miss `3`, because `10-5=5`.

operator/pkg/operator.go
```diff
@@ -181,6 +238,29 @@ func (o *Operator) Start(ctx context.Context) error {
...
+func (o *Operator) ProcessMissedBatchesWhileOffline(c chan int) {
+	// this means there was no file or no batches have been verified
+	if o.lastProcessedBatch.BlockNumber == 0 {
+		c <- 0
+		return
+	}
```
Oppen: I think we should add a log for this. I picture an error condition where:
1. Service died during marshal, and because of that
2. Recovery finds a corrupted file, so
3. We silently don't recover as expected.

operator/pkg/operator.go
```diff
@@ -194,6 +274,7 @@ func (o *Operator) Start(ctx context.Context) error {
 func (o *Operator) handleNewBatchLogV2(newBatchLog *servicemanager.ContractAlignedLayerServiceManagerNewBatchV2) {
 	o.Logger.Infof("Received new batch log V2")
 	err := o.ProcessNewBatchLogV2(newBatchLog)
+	o.UpdateLastProcessBatch(newBatchLog.TaskCreatedBlock)
```
Oppen: This should happen only after we finish verifying. Possibly `defer` this.

Oppen: Also, the way this is updated looks like a data race. We should probably write a buffered channel instead, so a single goroutine updates the value and writes the file. Possibly as part of the `select` on the main loop.

MarcosNicolau: Yeah that's a data race 100%

operator/pkg/operator.go
MarcosNicolau: Yes... I guess no number can cover this cases 100%. But given that each batch is produced every 12 seconds, It might be too improbable for 5 or 10 new batches to finish before the oldest ones. Although increasing the number might help, maybe we should came up with a better method or maybe it is just unnecessary. We should discuss more.

MarcosNicolau: Updated!

core/chainio/avs_reader.go
```diff
@@ -59,6 +70,73 @@ func (r *AvsReader) GetErc20Mock(tokenAddr gethcommon.Address) (*contractERC20Mo
 	return erc20Mock, nil
 }
 
-func (r *AvsReader) IsOperatorRegistered(address gethcommon.Address) (bool, error) {
+func (r *AvsReader) IsOperatorRegistered(address ethcommon.Address) (bool, error) {
 	return r.ChainReader.IsOperatorRegistered(&bind.CallOpts{}, address)
 }
+
+// Returns all the "NewBatchV3" logs that have not been responded starting from the given block number
+func (r *AvsReader) GetNotRespondedTasksFrom(fromBlock uint64) ([]servicemanager.ContractAlignedLayerServiceManagerNewBatchV3, error) {
```
MarcosNicolau, line 73: Should we create two functions `GetNotRespondedTasksFromV3` and `GetNotRespondedTasksFromV2` here and choose which one to call based on the operator version?

operator/pkg/operator.go
```diff
@@ -136,6 +149,51 @@ func (o *Operator) SubscribeToNewTasksV3() (chan error, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal batch: %v", err)
+	}
+
+	err = os.WriteFile(o.lastProcessedBatchLogFile, json, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("failed to write to file: %v", err)
+	}
+
+	o.Logger.Info("Updated latest block json file")
```
Oppen: Consider also printing the block number here.

operator/pkg/operator.go
```diff
@@ -174,13 +235,43 @@ func (o *Operator) Start(ctx context.Context) error {
...
+			go o.handleNewBatchLogV3(newBatchLogV3, batchProcessorChan)
+		case bacthProcessed := <-batchProcessorChan:
+			_ = o.UpdateLastProcessBatch(bacthProcessed)
+
 		}
 	}
 }
 
+// Here we query all the batches that have not yet been verified starting from
+// the latest verified batch by the operator. We also get the prior 5 and check if we need to verify them as well
```
Oppen:
```suggestion
// the latest verified batch by the operator. We also get the prior 100 and check if we need to verify them as well
```

operator/pkg/operator.go
```diff
@@ -174,13 +235,43 @@ func (o *Operator) Start(ctx context.Context) error {
 				o.Logger.Fatal("Could not subscribe to new tasks V3")
 			}
 		case newBatchLogV2 := <-o.NewTaskCreatedChanV2:
-			go o.handleNewBatchLogV2(newBatchLogV2)
+			go o.handleNewBatchLogV2(newBatchLogV2, batchProcessorChan)
 		case newBatchLogV3 := <-o.NewTaskCreatedChanV3:
-			go o.handleNewBatchLogV3(newBatchLogV3)
+			go o.handleNewBatchLogV3(newBatchLogV3, batchProcessorChan)
+		case bacthProcessed := <-batchProcessorChan:
+			_ = o.UpdateLastProcessBatch(bacthProcessed)
```
Oppen: We should at least log the error, if any.

core/chainio/avs_reader.go
Oppen: I'm not sure if two functions are needed, but we should be able to handle both versions, right?

operator/pkg/operator.go
MarcosNicolau: Updated!

Oppen (approved):

operator/pkg/operator.go
```diff
@@ -136,6 +149,51 @@ func (o *Operator) SubscribeToNewTasksV3() (chan error, error) {
 	return o.avsSubscriber.SubscribeToNewTasksV3(o.NewTaskCreatedChanV3)
 }
 
+type OperatorLastProcessedBatch struct {
+	BlockNumber uint32 `json:"block_number"`
+}
```
uri-99: Is this overhead really necessary and useful?

operator/pkg/operator.go
```diff
@@ -136,6 +149,51 @@ func (o *Operator) SubscribeToNewTasksV3() (chan error, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal batch: %v", err)
+	}
+
+	err = os.WriteFile(o.lastProcessedBatchLogFile, json, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("failed to write to file: %v", err)
+	}
+
+	o.Logger.Info("Updated latest block json file, new block: %v", blockNumber)
```
uri-99: check how much is this logging, i think it does it on every batch. is it too much?

config-files/config-operator-1.yaml
```diff
@@ -31,6 +31,7 @@ operator:
   enable_metrics: true
   metrics_ip_port_address: localhost:9092
   max_batch_size: 268435456 # 256 MiB
+  last_processed_batch_filepath: "operator-1.json"
```
uri-99: this file is being stored in the root of the project, it should be stored somewhere more relevant, such as in config-files.

uri-99: Actually, why is this another file? it could be an extra parameter in the already existing `config-operator.yaml`

operator/pkg/operator.go
```diff
@@ -174,13 +235,44 @@ func (o *Operator) Start(ctx context.Context) error {
...
 			}
 		case newBatchLogV2 := <-o.NewTaskCreatedChanV2:
-			go o.handleNewBatchLogV2(newBatchLogV2)
+			go o.handleNewBatchLogV2(newBatchLogV2, batchProcessorChan)
 		case newBatchLogV3 := <-o.NewTaskCreatedChanV3:
-			go o.handleNewBatchLogV3(newBatchLogV3)
+			go o.handleNewBatchLogV3(newBatchLogV3, batchProcessorChan)
+		case bacthProcessed := <-batchProcessorChan:
+			err = o.UpdateLastProcessBatch(bacthProcessed)
+			o.Logger.Errorf("Error while updating last process batch", "err", err)
```
uri-99: im getting this error out of the box
image

operator/pkg/operator.go
```diff
@@ -174,13 +235,44 @@ func (o *Operator) Start(ctx context.Context) error {
...
+func (o *Operator) ProcessMissedBatchesWhileOffline(c chan uint32) {
+	// this is the default value
+	// and it means there was no file so no batches have been verified
+	if o.lastProcessedBatch.BlockNumber == 0 {
+		o.Logger.Info("Not continuing with missed batch processing, as operator hasn't verified anything yet...")
+		return
+	}
+
+	o.Logger.Info("Getting missed tasks")
+	logs, err := o.avsReader.GetNotRespondedTasksFrom(uint64(o.lastProcessedBatch.BlockNumber - 100))
```
uri-99: magic number `100`. This number should be a better defined constant. Either in the config file, or maybe a bit simpler as a const in this file (similar to near line 63 of this file)

operator/pkg/operator.go
```diff
@@ -174,13 +235,44 @@ func (o *Operator) Start(ctx context.Context) error {
...
+
 		}
 	}
 }
 
+// Here we query all the batches that have not yet been verified starting from
+// the latest verified batch by the operator. We also get the prior 100 and check if we need to verify them as well
+// This last thing of getting the last 100 is to make sure we have not missed a batch since they are process in parallel
+// and a higher batch number might have been processed first than the lower one.
+// So getting the last 100 accounts for such cases
```
uri-99:
```suggestion
// Here we query all the batches that have not yet been verified starting from
// the latest verified batch by the operator. We also read from the previous
// 100 blocks, because as batches are processed in parallel, there could be
// unverified batches slightly before the latest verified batch
```

operator/pkg/operator.go
```diff
@@ -154,6 +212,9 @@ func (o *Operator) Start(ctx context.Context) error {
 		metricsErrChan = make(chan error, 1)
 	}
 
+	batchProcessorChan := make(chan uint32)
```
uri-99: This channel is to mark a batch as processed, right?
```suggestion
	batchProcessedChan := make(chan uint32)
```

operator/pkg/operator.go
```diff
@@ -174,13 +235,44 @@ func (o *Operator) Start(ctx context.Context) error {
 				o.Logger.Fatal("Could not subscribe to new tasks V3")
 			}
 		case newBatchLogV2 := <-o.NewTaskCreatedChanV2:
-			go o.handleNewBatchLogV2(newBatchLogV2)
+			go o.handleNewBatchLogV2(newBatchLogV2, batchProcessorChan)
 		case newBatchLogV3 := <-o.NewTaskCreatedChanV3:
-			go o.handleNewBatchLogV3(newBatchLogV3)
+			go o.handleNewBatchLogV3(newBatchLogV3, batchProcessorChan)
```
uri-99: Instead of passing the `batchProcessedChan` as parameter, it would be best for the relevant functions to call something like `o.registerBatchProcessed(int)` which sends a message to the channel. 
Hint: `type Operator struct {` should probably have a chan for this to work.

operator/pkg/operator.go
```diff
@@ -191,9 +283,16 @@ func (o *Operator) Start(ctx context.Context) error {
 // different events enables the smooth operator upgradeability
 
 // Process of handling batches from V2 events:
-func (o *Operator) handleNewBatchLogV2(newBatchLog *servicemanager.ContractAlignedLayerServiceManagerNewBatchV2) {
+func (o *Operator) handleNewBatchLogV2(newBatchLog *servicemanager.ContractAlignedLayerServiceManagerNewBatchV2, batchProcessedChan chan uint32) {
+	var err error
+	defer func() {
+		if err == nil {
+			batchProcessedChan <- uint32(newBatchLog.Raw.BlockNumber)
```
uri-99: Here instead of sending a message to the chan it received as a parameter, it would be best to either call a function that sends the message to the chan, or call a chan stored in the operator struct. I prefer the former to avoid repeting the code that sends a message to a chan.

operator/pkg/operator.go
```diff
@@ -262,9 +361,15 @@ func (o *Operator) ProcessNewBatchLogV2(newBatchLog *servicemanager.ContractAlig
 }
 
 // Process of handling batches from V3 events:
-func (o *Operator) handleNewBatchLogV3(newBatchLog *servicemanager.ContractAlignedLayerServiceManagerNewBatchV3) {
+func (o *Operator) handleNewBatchLogV3(newBatchLog *servicemanager.ContractAlignedLayerServiceManagerNewBatchV3, batchProcessedChan chan uint32) {
+	var err error
+	defer func() {
+		if err == nil {
+			batchProcessedChan <- uint32(newBatchLog.Raw.BlockNumber)
```
uri-99: Ditto

core/chainio/avs_reader.go
```diff
@@ -59,6 +70,73 @@ func (r *AvsReader) GetErc20Mock(tokenAddr gethcommon.Address) (*contractERC20Mo
...
+		Topics:    [][]ethcommon.Hash{{newBatchEvent.ID, {}}},
+	}
+
+	logs, err := r.AvsContractBindings.ethClient.FilterLogs(context.Background(), query)
+	if err != nil {
+		logs, err = r.AvsContractBindings.ethClientFallback.FilterLogs(context.Background(), query)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get logs: %w", err)
+		}
+	}
```
uri-99: FilterNewBatchV3 from go bindings (contracts/bindings/AlignedLayerServiceManager/binding.go) takes care of this. In its FilterOpts you can set the FromBlock (called Start).

All this will be removed when using out go bindings

core/chainio/avs_reader.go
```diff
@@ -59,6 +70,73 @@ func (r *AvsReader) GetErc20Mock(tokenAddr gethcommon.Address) (*contractERC20Mo
...
+
+	var tasks []servicemanager.ContractAlignedLayerServiceManagerNewBatchV3
+
+	for _, logEntry := range logs {
+		var task servicemanager.ContractAlignedLayerServiceManagerNewBatchV3
+
+		err := alignedLayerServiceManagerABI.UnpackIntoInterface(&task, "NewBatchV3", logEntry.Data)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unpack log data: %w", err)
+		}
```
uri-99: Ditto

core/chainio/avs_reader.go
```diff
@@ -59,6 +70,73 @@ func (r *AvsReader) GetErc20Mock(tokenAddr gethcommon.Address) (*contractERC20Mo
...
+
+		// The second topic is the batch merkle root, as it is an indexed variable in the contract
+		task.BatchMerkleRoot = logEntry.Topics[1]
```
uri-99: This will also be simpler to read, avoiding random index access

core/chainio/avs_subscriber.go
```diff
@@ -102,7 +103,9 @@ func (s *AvsSubscriber) SubscribeToNewTasksV2(newTaskCreatedChan chan *servicema
 					s.logger.Debug("Failed to get latest task from blockchain", "err", err)
 					continue
 				}
-				s.processNewBatchV2(latestBatch, batchesSet, newBatchMutex, newTaskCreatedChan)
+				if latestBatch != nil {
+					s.processNewBatchV2(latestBatch, batchesSet, newBatchMutex, newTaskCreatedChan)
```
uri-99, line 103: nice new feat

core/chainio/avs_subscriber.go
```diff
@@ -358,6 +363,19 @@ func (s *AvsSubscriber) getLatestTaskFromEthereumV2() (*servicemanager.ContractA
 	// The second topic is the batch merkle root, as it is an indexed variable in the contract
 	latestTask.BatchMerkleRoot = lastLog.Topics[1]
 
+	// return the task if has not been responded only
```
uri-99: This functionality currently is useful, but the function name does not indicate this.
This functionality should also refactor its name to `getLatestNonRespondedTask` or similar, to indicate it.

core/chainio/avs_subscriber.go
```diff
@@ -422,6 +440,19 @@ func (s *AvsSubscriber) getLatestTaskFromEthereumV3() (*servicemanager.ContractA
...
```
uri-99: Ditto

core/chainio/avs_subscriber.go
uri-99: Also, unrelated, I understand you inspired your getTask using the query from this function. But it is wrong and there is an issue about this incorrect implementation: https://github.com/yetanotherco/aligned_layer/issues/978

uri-99 (changes requested):

operator/pkg/operator.go
MarcosNicolau: Not really, it just makes it easier when declaring new fields if needed in the future and when writing in the file. I'll see if removing it is worthwhile.

operator/pkg/operator.go
MarcosNicolau: That typically happens when you try to read at startup and no file has been created. If not created, the operator will create the file at the provided path when writing. I'll make sure to create the file on read as well or change the logging for a `warn` and tell the user the reason.

config-files/config-operator-1.yaml
MarcosNicolau: Personally, I don't think saving the block to the config file is a good idea, we should discuss it.

operator/pkg/operator.go
MarcosNicolau: nvm, it is because we should check `err != nil`.

core/chainio/avs_reader.go
MarcosNicolau: Updated!

operator/pkg/operator.go
MarcosNicolau: I think is better to leave it, as we can put the channel here as well!

operator/pkg/operator.go
```diff
@@ -332,6 +440,18 @@ func (o *Operator) ProcessNewBatchLogV3(newBatchLog *servicemanager.ContractAlig
 	return nil
 }
 
+func (o *Operator) afterHandlingBatchV2(log *servicemanager.ContractAlignedLayerServiceManagerNewBatchV2, succeeded bool) {
+	if succeeded {
+		o.lastProcessedBatch.batchProcessedChan <- uint32(log.Raw.BlockNumber)
```
uri-99, line 449: why not simply
```
err = o.UpdateLastProcessBatch(blockNumber)
```

operator/pkg/operator.go
```diff
@@ -332,6 +440,18 @@ func (o *Operator) ProcessNewBatchLogV3(newBatchLog *servicemanager.ContractAlig
...
+	}
+}
+
+func (o *Operator) afterHandlingBatchV3(log *servicemanager.ContractAlignedLayerServiceManagerNewBatchV3, succeeded bool) {
+	if succeeded {
+		o.lastProcessedBatch.batchProcessedChan <- uint32(log.Raw.BlockNumber)
+	}
+}
```
uri-99, line 457: why not simply
```
err = o.UpdateLastProcessBatch(blockNumber)
```

operator/pkg/operator.go
uri-99, line 1: `batchProcessedChan` may be unnecessary, why not call the function directly? 
It would save on overhead code, and the handling of an extra chan

core/chainio/avs_subscriber.go
```diff
@@ -317,59 +318,50 @@ func (s *AvsSubscriber) getLatestTaskFromEthereumV2() (*servicemanager.ContractA
...
+		return nil, nil
+	}
 
-	return &latestTask, nil
+	return lastLog, nil
 }
 
 // getLatestTaskFromEthereum queries the blockchain for the latest task using the FilterLogs method.
 // The alternative to this is using the FilterNewBatch method from the contract's filterer, but it requires
 // to iterate over all the logs, which is not efficient and not needed since we only need the latest task.
```
uri-99: Missing function comment change

uri-99 (changes requested):

operator/pkg/operator.go
MarcosNicolau: It is necessary to prevent any race condition while writing to the file as tasks are processed in parallel. Channels assure us that all messages are executed one by one synchronously and not concurrently.

uri-99: go chans are a good way of [controlling concurrent  accesses](https://stackoverflow.com/questions/29981050/concurrent-writing-to-a-file).

Oppen: That channel is a synchronization mechanism. Since we have one value and many tasks trying to write to it, we need to put someone "in charge" of it, and let everyone else message that someone.

config-files/config-operator-1.yaml
Oppen, line 1: I believe these changes are unrelated to the PR. Maybe just check the old version.

config-files/config-operator-2.yaml
Oppen, line 1: Ditto.

config-files/config-operator-3.yaml
Oppen, line 1: Ditto.

config-files/config-operator.yaml
Oppen, line 1: Ditto.

config-files/config.yaml
Oppen, line 1: Ditto.

operator/pkg/operator.go
```diff
@@ -177,10 +240,50 @@ func (o *Operator) Start(ctx context.Context) error {
...
+	logs, err := o.avsReader.GetNotRespondedTasksFrom(fromBlock)
+	if err != nil {
+		return
+	}
+	o.Logger.Infof(fmt.Sprintf("Missed tasks retrieved, total tasks to process: %v", len(logs)))
+
+	o.Logger.Infof("Starting to verify missed batches while offline")
+	for _, logEntry := range logs {
+		go o.handleNewBatchLogV3(&logEntry)
+	}
```
uri-99: should return if len(logs) == 0
image

uri-99 (changes requested):
final detail and i think its ready to approve

uri-99:
also add the new .json file to .gitignore to avoid having files with random variables being pushed to the repo

config-files/config-operator-1.yaml
MarcosNicolau: It adds a new field for the operator: `last_processed_batch_filepath`. I'll make sure to not push formatting changes along with actual changes next time!

uri-99 (approved):

core/chainio/avs_subscriber.go
```diff
@@ -165,12 +164,14 @@ func (s *AvsSubscriber) SubscribeToNewTasksV3(newTaskCreatedChan chan *servicema
 			case newBatch := <-internalChannel:
 				s.processNewBatchV3(newBatch, batchesSet, newBatchMutex, newTaskCreatedChan)
 			case <-pollLatestBatchTicker.C:
-				latestBatch, err := s.getLatestTaskFromEthereumV3()
+				latestBatch, err := s.getLatestNotRespondedTaskFromEthereumV3()
```
Oppen, line 167: Maybe "Pending" is better than "NotResponded".

uri-99: +1

Oppen:
OK, seems to work here.

PR #1077, opened 2024-09-25
feat(telemetry): implement telemetry for aggregator

### LINKED ISSUE #998: rewrite operator_tracker in elixir, add telemetry to it

### LINKED ISSUE #1075: feat(telemetry): implement telemetry for aggregator
# Description
This issue supersedes #936 

The aggregator has to communicate to the telemetry_api using HTTP endpoints
The telemetry_api will handle traces creation

# Tasks

- [x] Docker compose with Otel Collector and Jaeger UI
- [x] Init traces on NewTask
- [x] Add span on Operator responses
- [x]  Add span on Quorum Reached
- [x] Add span on Tasks errors (ie task expired)
- [x] Close traces after a period of time the quorum was reached to wait for most operators responses

JuArce:
# Description
This PR supersedes #949 

The aggregator has to communicate to the telemetry_api using HTTP endpoints
The telemetry_api will handle traces creation

closes #1075 

# Tasks

- [x] Docker compose with Otel Collector and Jaeger UI
- [x] Init traces on NewTask
- [x] Add span on Operator responses
- [x]  Add span on Quorum Reached
- [x] Add span on Tasks errors (ie task expired)
- [x] Close traces after a period of time the quorum was reached to wait for most operators responses
- [x] Show operator not responding

# How to Test

> [!TIP]
> You will need a lot of terminals 😎

1. Run anvil

```
```

2. Run aggregator

```
make aggregator_start
```

3. Run batcher

```
make batcher_start_local
```

4. Build operator

```
make build_operator
```

5. Register operator 1

```
export OPERATOR_ADDRESS=0x70997970C51812dc3A010C7d01b50e0d17dc79C8
make operator_full_registration CONFIG_FILE=config-files/config-operator-1.yaml
```

6. Register operator 2

```
export OPERATOR_ADDRESS=0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC
make operator_full_registration CONFIG_FILE=config-files/config-operator-2.yaml
```

7. Register operator 3

```
export OPERATOR_ADDRESS=0x90F79bf6EB2c4f870365E785982E1f101E93b906
make operator_full_registration CONFIG_FILE=config-files/config-operator-3.yaml
```

8. Run Telemetry

```
make telemetry_create_env
make telemetry_run_db
make open_telemetry_start
make telemetry_start 
```

9. Run Operator 1

```
./operator/build/aligned-operator start --config ./config-files/config-operator-1.yaml
```

11. Run Operator 2

```
./operator/build/aligned-operator start --config ./config-files/config-operator-2.yaml
```

13. Run Opreator 3

```
./operator/build/aligned-operator start --config ./config-files/config-operator-3.yaml
```

14. Send proofs

```
```

15. Go to [Jaeger UI](http://localhost:16686/search) and find the new traces

You should see traces like the following one

image

image


aggregator/internal/pkg/telemetry.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+type TaskErrorMessage struct {
+	MerkleRoot string `json:"merkle_root"`
+	TaskError  string `json:"error"`
+}
+
+type Telemetry struct {
+	client  *http.Client
+	baseURL *url.URL
```
Oppen: Do these need to be pointers?

telemetry_api/lib/telemetry_api/periodic/operator_fetcher.ex
```diff
@@ -0,0 +1,32 @@
...
+
+  @wait_time_ms (
+    case Integer.parse(wait_time_str) do
+      :error -> raise("OPERATOR_FETCHER_WAIT_TIME_MS is not a number, received: #{wait_time_str}")
+      {num, _} -> num
+    end
+  )
+
+  def start_link(_) do
+    Task.start_link(&poll_serivce/0)
```
Oppen:
```suggestion
    Task.start_link(&poll_service/0)
```

aggregator/internal/pkg/aggregator.go
```diff
@@ -294,6 +305,7 @@ func (agg *Aggregator) sendAggregatedResponse(batchIdentifierHash [32]byte, batc
 	if err != nil {
 		agg.walletMutex.Unlock()
 		agg.logger.Infof("- Unlocked Wallet Resources: Error sending aggregated response for batch %s. Error: %s", hex.EncodeToString(batchIdentifierHash[:]), err)
+		agg.telemetry.LogTaskError(batchMerkleRoot, err)
```
uri-99, line 308: Should something similar be done on line 317 of this file? it is another `if err != nil`

telemetry_api/config/dev.exs
```diff
@@ -65,3 +65,6 @@ config :phoenix, :plug_init_mode, :runtime
 
 # Configure ethereumex url
 config :ethereumex, url: "http://localhost:8545"
+
+# For development, we use the stdout exporter to ensure everything is working properly
+# config :opentelemetry, traces_exporter: {:otel_exporter_stdout, []}
```
uri-99, line 70: Is this uncommented during dev but commented to deploy in staging?

telemetry_api/lib/telemetry_api/operators.ex
```diff
@@ -29,14 +29,40 @@ defmodule TelemetryApi.Operators do
...
+
+  ## Examples
+
+      iex> get_operator_by_id("some_id")
+      {:ok, %Operator{}}
+
+      iex> get_operator_by_id("non_existent_id")
+      {:error, :not_found, "Operator not found for id: non_existent_id"}
+  """
+  def get_operator_by_id(id) do
```
uri-99: I recommend overloading the `get_operator()` method, so if you send an address it will call `get_operator(address)`, and if you send an id it will call `get_operator(id)`.

You can also do this by meking `get_operator` receive a struct, and you send the parameters like `get_operator(%{id: id})` or `get_operator(%{address: address})` .

You could even receive an `Operator` struct as a param so you are restricted by sending 1 (or n) parameters that already exist in the `Operator` structure.

telemetry_api/lib/telemetry_api/periodic/operator_fetcher.ex
```diff
@@ -0,0 +1,32 @@
...
+
+  defp poll_service() do
+    receive do
+    after
+      @wait_time_ms ->
+        case Operators.fetch_all_operators() do
+          {:ok, _} -> :ok
+          {:error, message} -> IO.inspect "Couldn't fetch operators: #{IO.inspect message}"
+        end
+        poll_service()
```
uri-99: The reason why we didn't do the `periodically` this way on the Explorer is that if one call of `poll_service` panics, or gets stuck for any reason, no one will start the next `poll_service` . 

I recommend you to use `:timer.send_interval` , this way always every n ms, a new Task will be created with the target code

telemetry_api/lib/telemetry_api/traces.ex
```diff
@@ -0,0 +1,189 @@
...
+      iex> register_operator_response(merkle_root, operator_id)
+      :ok
+  """
+  def register_operator_response(merkle_root, operator_id) do
+    with {:ok, operator} <- Operators.get_operator_by_id(operator_id),
+         {:ok, trace} <- set_current_trace(merkle_root) do
+      operator_stake = String.to_integer(operator.stake)
+      new_stake = trace.current_stake + operator_stake
+      new_stake_fraction = new_stake / trace.total_stake
+      operator_stake_fraction = operator_stake / trace.total_stake
```
uri-99: Are these numbers not `Decimal`? I think they should be, ints in postgres are not big enough to handle some of these numbers.
Then, i.e. for divisioind, the following is the way to divide Decimals: `Decimal.div(operator_stake, total_stake)`

uri-99 (changes requested):

aggregator/internal/pkg/telemetry.go
avilagaston9: They didn't. Changed them!

aggregator/internal/pkg/aggregator.go
avilagaston9: That's right! Added!

telemetry_api/config/dev.exs
avilagaston9: We're actually not using it at all. We just commented it in case we need it in the future.

telemetry_api/lib/telemetry_api/operators.ex
avilagaston9: Changed to the map version:

```Elixir
get_operator(%{address: address})
get_operator(%{id: id})
```

If I use an `Operator` struct, the function will match any `%Operator{}` because it automatically assigns nil to all fields internally.

telemetry_api/lib/telemetry_api/periodic/operator_fetcher.ex
avilagaston9: Nice! Changed to a `GenServer` using `:timer.send_interval`

telemetry_api/lib/telemetry_api/traces.ex
avilagaston9: We are storing the stakes as strings in the db. I have changed them to `Decimals` in Elixir now.

entropidelic (approved):

uri-99 (approved):

PR #1080, opened 2024-09-25
feat: add operator identity check on registry manager contract before creation

JulianVentura:
### Motivation

Whenever we receive an update request from a running operator, we want to validate it's identity with the `RegistryCoordinatorManager` before updating its version.

### Description

This PR adds a validation on which the `RegistryCoordinatorManager` contract is fetched to ask for caller operator identity when the `/api/operators` endpoint is fetched.


telemetry_api/lib/abi/IRegistryCoordinator.json
JuArce, line 1: I think ABI should be on `priv` directory

JulianVentura: Done

Issue #1069, opened 2024-09-24
Usage of block count instead of block timestamp (INFORMATIONAL)


MauroToscano:
Using block count is less intuitive than using a block timestamp. When using a timestamp a user could be informed of the exact time that his balance becomes unlocked right at the moment he calls the unlock() function. 

Whereas, with a block count the time cannot be exactly known upfront due to the block time variability.

PR #1085, opened 2024-09-25
feat(audit): Use block timestamp instead of block number.

PatStiles:
closes: #1069

Switches the userData to check and return in events the block timestamp instead of block number.

## To Test:
- [x] `anvil_deploy_aligned_contracts` and `anvil_start_with_block_time`
- [x] Run local testnet and ensure nothing errors.

# To Test
- Change modify the [block time](https://github.com/yetanotherco/aligned_layer/pull/1085/files#diff-20c3c151edaf33f5a8a50a8dc5945aeda8c89733dc7fc1ca53f63a8b05609585L23) constant. You should observe the unlock time decrease in the local testnet.


Oppen (approved):

MauroToscano (changes requested):
New JSON of contracts should be pushed. Also, it needs to have in the description how it should be tested. It can include changing the time for something smaller like 45 seconds just to check really quick

MarcosNicolau (approved):
I tested it with `cast` running the following commands:
1. Fund account: `cast send --value 100000 0x7969c5eD335650692Bc04293B07F5BF2e7A673C0  --from <ADDRESS>  --private-key <PK> --rpc-url "http://localhost:8545"`
 cast send 0x7969c5eD335650692Bc04293B07F5BF2e7A673C0  "unlock()" --rpc-url http://localhost:8545 --private-key <PK> --from <ADDRESS>
2. Unlock to withdraw: ` cast send 0x7969c5eD335650692Bc04293B07F5BF2e7A673C0  "unlock()" --rpc-url http://localhost:8545 --private-key <PK>  --from <ADDRESS> `
3. Advance `anvil` seconds to not wait: 
```bash
curl http://localhost:8545 \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[2000],"id":1}'
```
5. And now you should be able to withdraw: `cast send 0x7969c5eD335650692Bc04293B07F5BF2e7A673C0  "withdraw(uint256)" 10000 --rpc-url http://localhost:8545 --private-key 0x2a871d0798f97d79848a013d4936a73bf4cc922c825d33c1cf7073dff6d409c6      
`

And checking the `BalanceUnlocked` data when doing `unlock` returns the correct timestamp.

It seems that it works as expected.

contracts/src/core/BatcherPaymentService.sol
```diff
@@ -40,7 +40,7 @@ contract BatcherPaymentService is
     error UserHasNoFundsToUnlock(address user); // b38340cf
     error UserHasNoFundsToLock(address user); // 6cc12bc2
     error PayerInsufficientBalance(uint256 balance, uint256 amount); // 21c3d50f
-    error FundsLocked(uint256 unlockBlock, uint256 currentBlock); // bedc4e5a
+    error FundsLocked(uint256 unlockBlock, uint256 currentBlockTime); // bedc4e5a
```
IAvecilla: I think that the first parameter `unlockBlock` should also change to `unlockBlockTime`

contracts/src/core/BatcherPaymentService.sol
```diff
@@ -167,15 +167,17 @@ contract BatcherPaymentService is
...
+            UNLOCK_BLOCK_TIME;
+        emit BalanceUnlocked(msg.sender, userData[msg.sender].unlockBlockTime);
     }
 
     function lock() external whenNotPaused {
         if (userData[msg.sender].balance == 0) {
             revert UserHasNoFundsToLock(msg.sender);
         }
-        userData[msg.sender].unlockBlock = 0;
+        userData[msg.sender].unlockBlockTime = 0;
```
IAvecilla, line 180: This is more of a general question about the `unlock/lock` flow: we are not preventing a user from repeatedly calling unlock and lock, which would reset their `unlockBlockTime`. Should we consider preventing that scenario? It might be possible for a user to accidentally reset the timer, and it could be helpful to add some checks before allowing that, from what I understand a check `unlockBlockTime == 0` for the unlock would be sufficient. This seems beyond the scope of this PR, but it may be worth considering for the future.

IAvecilla (approved):

JuArce:
Deployed in stage

MauroToscano (approved):

PR #1087, opened 2024-09-25
fix: replay attack on batcher DoS a particular user

### LINKED ISSUE #1081: replay attack on batcher disrupts a particular user

uri-99:
> [!NOTE]
> This PR should be merged after #1106

# This PR
fixes #1081 

## To Deploy
This PR changes the Batcher, so it should be redeployed, with no impact on the rest of Aligned

## To Test
1) Test normal flow

### DIFF
--- a/batcher/aligned-batcher/src/lib.rs
+++ b/batcher/aligned-batcher/src/lib.rs
@@ -462,7 +462,7 @@ impl Batcher {
                     .await
             } else {
                 if !self
-                    .check_user_balance_and_increment_proof_count(&addr)
+                    .check_user_balance(&addr)
                     .await
                 {
                     send_message(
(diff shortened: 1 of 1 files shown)


Oppen (approved):

MauroToscano:
Not needed

PR #1088, opened 2024-09-25
fix: operator double signature

MarcosNicolau:
**Changes**
Updates go eigensdk to version 0.1.12 which fixes the double signature err. All files using the SDK have been updated accordingly to the new version. Finally, our check has been removed since the SDK handles it now.

**Test**
When modifying the operator and sending two signed responses after verifying one batch, the aggregator should ignore the second one and the system should continue working as usual. See here for a more detailed testing #1097.

### DIFF
--- a/aggregator/internal/pkg/aggregator.go
+++ b/aggregator/internal/pkg/aggregator.go
@@ -146,7 +146,7 @@ func NewAggregator(aggregatorConfig config.AggregatorConfig) (*Aggregator, error
 		return taskResponseDigest, nil
 	}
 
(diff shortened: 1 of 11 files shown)


MauroToscano (changes requested):
- Default config of operator not registering

MarcosNicolau:
> * Default config of operator not registering

Fixed! We needed to wait for the receipt when depositing into the strategy contract. Tbh, I don't know why it was working with the rest of the configurations.

PatStiles (approved):
Warn observed LGTM

entropidelic (approved):

MauroToscano (approved):

PR #1090, opened 2024-09-25
feat: disable verifiers

### LINKED ISSUE #1082: Add a Bitmap in Ethereum to disable verifiers
- [ ] Add a bitmap in Ethereum to disable verifiers in the ServiceManager. Start in 0 when no verifier is disabled
- [ ] Add a method to update it with the admin sig. Make sure it's easy to use. 
- [ ] Make the operators fetch the last bitmap before resolving the task, if it's blacklisted, assume the proof is false. 
- [ ] Make the same logic in the Batcher

MarcosNicolau:
**Changes**
This pr introduces proving systems or verifiers blacklisting through a new bitmap field in the `ServiceManager` contract. The batcher and operator, before doing a verification, they make sure that the proving system is indeed whitelisted by calling `isVerifierValid` function in the `ServiceManager` contract. 

It also includes the verification for disabled verifiers in the operators and the batcher logic. For the batcher it will now check for every new block if the disabled verifiers have changed, in the afirmative case it checks all the proofs inside the batcher queue and remove all of them indistinctly if it was a proof using the disabled verifier or not.

**Test**
You can test this following these steps:
- Set up and run all the different components (anvil, aggregator, operator and batcher)
- In another terminal run `make batcher_send_burst_groth16` just to keep generating new proofs and avoid problems trying to submit repeated batches.
- To update the blacklisted verifiers bitmap you can use the following makefile target that sends a transaction to the service manager with the function signature `disableVerifier(uint8)`. The `VERIFIER_ID=3` variable will disable the SP1 verifier:
```
make verifier_disable_devnet VERIFIER_ID=3
```
- Try to send a sp1 proof that it should be rejected by the batcher: `make batcher_send_sp1_task`.
- Whitelist the verifier doing:
```
make verifier_enable_devnet VERIFIER_ID=3
```
- Try to send a sp1 proof and it should be working again: `make batcher_send_sp1_task`.

If you want to try and disable all the other verifiers run the same command mentioned above changing the `VERIFIER_ID`
that is the ID corresponding to the `ProvingSystemId` enum in the `proving_systems.go` file, being `GnarkPlonkBls12_381` the id 0.

**Disabled verifiers in the batch queue testing**

1. Start the Aligned components (anvil, aggregator, and operator).
2. Open the batcher's configuration file at `config-files/config-batcher.yaml` and adjust the following parameters:
   ```
   block_interval: 100000
   batch_size_interval: 4
   ```
   > This ensures the batch won't be submitted while sending valid proofs, but the batch remains open, allowing you to disable the verifier later.
   
3. Run the batcher:  
   `make batcher_start_local`
   
4. Send 3 proofs using any of the verifiers:  
   - `make batcher_send_sp1_task`
   - `make batcher_send_risc0_task`
   - `make batcher_send_groth16_bn254_task`
   
5. Disable the SP1 verifier:  
   `make verifier_disable_devnet VERIFIER_ID=3`
   
6. Verify that all the clients stop and encounter an `BatchReset` error. The batcher's queue should now be empty.

---
This closes #1082


contracts/src/core/AlignedLayerServiceManagerStorage.sol
```diff
@@ -11,6 +11,11 @@ abstract contract AlignedLayerServiceManagerStorage {
     // KEY is keccak256(batchMerkleRoot,senderAddress)
     mapping(bytes32 => BatchState) public batchesState;
 
+    // Bitmap representing blacklisted verifiers
+    // Each verifier is blacklisted if its corresponding bit is set to 1
+    // The verifier index follows its corresponding value in the `ProvingSystemId` enum being 0 the first verifier.
+    uint256 public blacklistedVerifiers;
```
uri-99: We should discuss if a bitmap is the best way to implement this system. It is not really easy to query since there is no on-chain way of knowing which bit is related to which Proving System.

Would a map be a better choice?

Reading is free for both cases (`cast call` and not `cast send`) , and while writing may be cheaper in a bitmap since it uses less memory, the usage of this write is really exceptional and may not justify the extra cost in usability (thus auditability and thus security), of a bitmap.

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -181,10 +183,38 @@ contract AlignedLayerServiceManager is
...
+    function isVerifierBlacklisted(
+        uint256 verifierIdx
+    ) external view validVerifierIdx(verifierIdx) returns (bool) {
+        uint256 bit = blacklistedVerifiers & (1 << verifierIdx);
+        return bit > 0;
+    }
+
+    function getBlacklistedVerifiers() external view returns (uint256) {
+        return blacklistedVerifiers;
+    }
```
uri-99: `blacklistedVerifiers` is public so this getter is unnecesary.

contracts/src/core/IAlignedLayerServiceManager.sol
```diff
@@ -62,4 +65,12 @@ interface IAlignedLayerServiceManager {
...
 
     function setAggregator(address _aggregator) external;
+
+    function isVerifierBlacklisted(
+        uint256 verifierIdx
+    ) external view returns (bool);
+    function blacklistVerifier(uint256 verifierIdx) external;
+    function whitelistVerifier(uint256 verifierIdx) external;
+    function setVerifiersBlacklist(uint256 bitmap) external;
+    function getBlacklistedVerifiers() external view returns (uint256);
```
uri-99: Redundant function since the variable is already public

uri-99 (changes requested):

contracts/src/core/IAlignedLayerServiceManager.sol
IAvecilla: True, fixed!

contracts/src/core/AlignedLayerServiceManagerStorage.sol
uri-99: So, as discussed, we will use a bitmap and follow the SDK. In the future, we will remove the values coded directly on the SDK and read them from a config file, so this same config file will be used to enable/disable verifiers.

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -326,4 +352,11 @@ contract AlignedLayerServiceManager is
         }
         _;
     }
+
+    modifier validVerifierIdx(uint256 verifierIdx) {
+        if (verifierIdx >= 64) {
```
uri-99:
```suggestion
        if (verifierIdx >= 256 ) {
```

uri-99: Actually this whole modifier can be removed if `verifierIdx` is a uint8

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -202,10 +204,34 @@ contract AlignedLayerServiceManager is
             senderAddress,
             batchersBalances[senderAddress]
         );
-
         payable(alignedAggregator).transfer(txCost);
     }
 
+    function isVerifierBlacklisted(
+        uint256 verifierIdx
```
uri-99, line 211:
```suggestion
        uint8 verifierIdx
```

uri-99:
```suggestion
        uint8 verifierIdx
```

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -202,10 +204,34 @@ contract AlignedLayerServiceManager is
...
+    ) external view validVerifierIdx(verifierIdx) returns (bool) {
+        uint256 bit = blacklistedVerifiers & (1 << verifierIdx);
+        return bit > 0;
+    }
+
+    function blacklistVerifier(
+        uint256 verifierIdx
```
uri-99:
```suggestion
        uint8 verifierIdx
```

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -202,10 +204,34 @@ contract AlignedLayerServiceManager is
...
+    ) external validVerifierIdx(verifierIdx) onlyOwner {
+        blacklistedVerifiers |= (1 << verifierIdx);
+        emit VerifierBlacklisted(verifierIdx);
+    }
+
+    function whitelistVerifier(
+        uint256 verifierIdx
```
uri-99:
```suggestion
        uint8 verifierIdx
```

contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -326,4 +352,11 @@ contract AlignedLayerServiceManager is
         }
         _;
     }
+
+    modifier validVerifierIdx(uint256 verifierIdx) {
+        if (verifierIdx >= 64) {
+            revert VerifierIdxOutOfBounds();
+        }
+        _;
+    }
```
uri-99:
```suggestion
```

contracts/src/core/IAlignedLayerServiceManager.sol
```diff
@@ -20,6 +20,8 @@ interface IAlignedLayerServiceManager {
     );
     event BatchVerified(bytes32 indexed batchMerkleRoot, address senderAddress);
     event BatcherBalanceUpdated(address indexed batcher, uint256 newBalance);
+    event VerifierBlacklisted(uint256 indexed verifierIdx);
+    event VerifierWhitelisted(uint256 indexed verifierIdx);
```
uri-99:
```suggestion
    event VerifierBlacklisted(uint8 indexed verifierIdx);
    event VerifierWhitelisted(uint8 indexed verifierIdx);
```

contracts/src/core/IAlignedLayerServiceManager.sol
```diff
@@ -63,4 +66,11 @@ interface IAlignedLayerServiceManager {
     function balanceOf(address account) external view returns (uint256);
 
     function setAggregator(address _aggregator) external;
+
+    function isVerifierBlacklisted(
+        uint256 verifierIdx
```
uri-99:
```suggestion
        uint8 verifierIdx
```

contracts/src/core/IAlignedLayerServiceManager.sol
```diff
@@ -63,4 +66,11 @@ interface IAlignedLayerServiceManager {
...
+    ) external view returns (bool);
+    function blacklistVerifier(uint256 verifierIdx) external;
+    function whitelistVerifier(uint256 verifierIdx) external;
```
uri-99:
```suggestion
    function blacklistVerifier(uint8 verifierIdx) external;
    function whitelistVerifier(uint8 verifierIdx) external;
```

uri-99 (changes requested):

operator/pkg/operator.go
```diff
@@ -332,7 +334,25 @@ func (o *Operator) ProcessNewBatchLogV3(newBatchLog *servicemanager.ContractAlig
 	return nil
 }
 
+func (o *Operator) IsVerifierValid(blacklisted_verifiers_bitmap *big.Int, verifierId common.ProvingSystemId) bool {
+	verifierIdInt := uint64(verifierId)
```
uri-99: uint8

operator/pkg/operator.go
```diff
@@ -332,7 +334,25 @@ func (o *Operator) ProcessNewBatchLogV3(newBatchLog *servicemanager.ContractAlig
...
+	bit := blacklisted_verifiers_bitmap.Uint64() & (1 << verifierIdInt)
```
uri-99: uint8

IAvecilla: Fixed in [ee0d97d](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/ee0d97d123d0d7a144a19b35c935b4ba4305ff8a)

operator/pkg/operator.go
IAvecilla: Fixed in [ee0d97d](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/ee0d97d123d0d7a144a19b35c935b4ba4305ff8a)

contracts/src/core/IAlignedLayerServiceManager.sol
IAvecilla: Fixed in [ee0d97d](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/ee0d97d123d0d7a144a19b35c935b4ba4305ff8a)

contracts/src/core/IAlignedLayerServiceManager.sol
IAvecilla: Fixed in [ee0d97d](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/ee0d97d123d0d7a144a19b35c935b4ba4305ff8a)

contracts/src/core/IAlignedLayerServiceManager.sol
IAvecilla: Fixed in [ee0d97d](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/ee0d97d123d0d7a144a19b35c935b4ba4305ff8a)

contracts/src/core/AlignedLayerServiceManager.sol
IAvecilla: Fixed in [ee0d97d](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/ee0d97d123d0d7a144a19b35c935b4ba4305ff8a)

contracts/src/core/AlignedLayerServiceManager.sol
IAvecilla: Fixed in [ee0d97d](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/ee0d97d123d0d7a144a19b35c935b4ba4305ff8a)

contracts/src/core/AlignedLayerServiceManager.sol
IAvecilla: Fixed in [ee0d97d](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/ee0d97d123d0d7a144a19b35c935b4ba4305ff8a)

uri-99 (approved):

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1174,14 +1204,20 @@ impl Batcher {
...
-                send_message(ws_conn_sink.clone(), ValidityResponseMessage::InvalidProof).await;
-                return Ok(()); // Send error message to the client and return
+            if self.pre_verification_is_enabled {
+                let result = zk_utils::verify(
+                    &client_msg.verification_data.verification_data,
+                    blacklisted_verifiers,
+                )
+                .await;
+                if let ValidityResponseMessage::InvalidProof(reason) = result {
+                    error!("Invalid proof detected. Verification failed.");
```
Oppen: Let's also log the reason.

operator/pkg/operator.go
```diff
@@ -331,7 +334,25 @@ func (o *Operator) ProcessNewBatchLogV3(newBatchLog *servicemanager.ContractAlig
 	return nil
 }
 
+func (o *Operator) IsVerifierValid(blacklisted_verifiers_bitmap *big.Int, verifierId common.ProvingSystemId) bool {
+	verifierIdInt := uint8(verifierId)
+	bit := uint8(blacklisted_verifiers_bitmap.Uint64()) & (1 << verifierIdInt)
```
Oppen: Are we sure we won't have more than eight different verifiers?

MarcosNicolau: The bitmap is implemented as a `u256` so it makes sense to enforce `u8`s as indexes as we save ourselves from checking if the provided number is between the boundary of the bitmap (0-255). Though, why are we casting `blacklisted_verifiers_bitmap` to an `uint64` instead of `uint256`? Doesn't go have `uint256`?

Oppen: We're casting the bitmap to `uint8` before applying the mask, which is a single bit shifted by the id number, so that limits us to 8 different verifiers.

IAvecilla: You're right, thanks for pointing that out. It works now because we have fewer than 8 verifiers, but it's cutting it close. I changed it to use `uint64` directly, which should be fine since we likely won't need more than 64.

Oppen: > 64 verifiers ought to be enough for anybody
- Not Bill Gates

Oppen (approved):
I believe this can be merged, but:
1. The choice of checking the blacklist inside the `verify` function seems odd to me. Blacklists are policy, and for one reason or another we may want to verify internally something we wouldn't allow a client to send. IMO this should be separate.
2. Some of the code would be simpler if `VerifierId`was `Copy`.

I'd consider these as tech debt, to be ignored for now and fix at a later time, as it's not a functional change.

batcher/aligned-sdk/src/core/types.rs
```diff
@@ -328,6 +344,56 @@ pub enum ValidityResponseMessage {
     InvalidPaymentServiceAddress(Address, Address),
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum ProofInvalidReason {
+    Unknown,
+    MissingVerificationData,
```
entropidelic, line 350: really like these enums. Given the context where this one was used, I think that instead of `VerificationData` it was meant to say `VerificationKey`

operator/pkg/operator.go
```diff
@@ -455,6 +462,18 @@ func (o *Operator) afterHandlingBatchV3(log *servicemanager.ContractAlignedLayer
 }
 
 func (o *Operator) verify(verificationData VerificationData, results chan bool) {
+	blacklisted_verifiers_bitmap, err := o.avsReader.BlacklistedVerifiers()
```
entropidelic: very smol nit, but let's keep the camelCase go convention for names :)

operator/pkg/operator.go
```diff
@@ -455,6 +462,18 @@ func (o *Operator) afterHandlingBatchV3(log *servicemanager.ContractAlignedLayer
...
+	if err != nil {
+		o.Logger.Errorf("Could not check verifier status: %s", err)
+		results <- false
+		return
+	}
+	is_verifier_valid := o.IsVerifierValid(blacklisted_verifiers_bitmap, verificationData.ProvingSystemId)
```
entropidelic: ditto

JuArce (changes requested):
I'm blocking the merge because I'm checking if the storage is correctly added  in the contract

Makefile
```diff
@@ -209,6 +209,23 @@ operator_deposit_and_register: operator_deposit_into_strategy operator_register_
 
 operator_full_registration: operator_get_eth operator_register_with_eigen_layer operator_mint_mock_tokens operator_deposit_into_mock_strategy operator_whitelist_devnet operator_register_with_aligned_layer
 
+# The verifier ID to blacklist or whitelist corresponds to the index of the verifier in the `ProvingSystemIDd` enum.
+verifier_whitelist_devnet:
```
JuArce: Maybe I'd change to enable/disable instead of whitelist/blacklist. I think this way is clearer

Makefile
```diff
@@ -209,6 +209,23 @@ operator_deposit_and_register: operator_deposit_into_strategy operator_register_
 
 operator_full_registration: operator_get_eth operator_register_with_eigen_layer operator_mint_mock_tokens operator_deposit_into_mock_strategy operator_whitelist_devnet operator_register_with_aligned_layer
 
+# The verifier ID to blacklist or whitelist corresponds to the index of the verifier in the `ProvingSystemIDd` enum.
```
JuArce:
```suggestion
# The verifier ID to blacklist or whitelist corresponds to the index of the verifier in the `ProvingSystemId` enum.
```

Makefile
JuArce: In order to go faster, we can ignore this change

contracts/src/core/AlignedLayerServiceManagerStorage.sol
JuArce: This variable should be added after `address public alignedAggregator;` because if not, it will shift the current storage and will break it.

Also, the GAP should be reduced, in this case, by 1, because we are adding an uint256

contracts/src/core/IAlignedLayerServiceManager.sol
```diff
@@ -63,4 +65,11 @@ interface IAlignedLayerServiceManager {
     function balanceOf(address account) external view returns (uint256);
 
     function setAggregator(address _aggregator) external;
+
+    function isVerifierBlacklisted(
```
JuArce: To keep consistency, add a blank line between functions declaration

JuArce (changes requested):

Makefile
IAvecilla: I did it since it was not that big of a change and I think it will improve the understanding of what these functions do.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -197,6 +213,7 @@ impl Batcher {
             chain_id,
             payment_service,
             payment_service_fallback,
+            service_manager,
```
JuArce, line 232: ServiceManager should have a fallback like the PaymentService

batcher/aligned-batcher/src/lib.rs
```diff
@@ -376,12 +392,33 @@ impl Batcher {
...
-        if self.pre_verification_is_enabled
-            && !zk_utils::verify(&nonced_verification_data.verification_data).await
-        {
-            error!("Invalid proof detected. Verification failed.");
-            send_message(ws_conn_sink.clone(), ValidityResponseMessage::InvalidProof).await;
-            return Ok(());
+        if self.pre_verification_is_enabled {
+            let disabled_verifiers = self.disabled_verifiers().await;
+            let verification_data = nonced_verification_data.verification_data.clone();
+            if disabled_verifiers & (U256::one() << verification_data.proving_system as u64)
```
JuArce: I know how it works, but maybe it could be documented or separated in a function called is_disabled() or something else to be more clear on the code flow

JuArce: Also, having this in a functions, makes easier add unit tests in the future

operator/pkg/operator.go
```diff
@@ -455,6 +455,18 @@ func (o *Operator) afterHandlingBatchV3(log *servicemanager.ContractAlignedLayer
 }
 
 func (o *Operator) verify(verificationData VerificationData, results chan bool) {
+	disabledVerifiersBitmap, err := o.avsReader.DisabledVerifiers()
```
JuArce: It is getting the verification bitmat for each proof. If the batch is too big, it will add a lot of latency

operator/pkg/utils.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"math/big"
+
+	"github.com/yetanotherco/aligned_layer/common"
+)
+
+func IsVerifierDisabled(disabledVerifiersBitmap *big.Int, verifierId common.ProvingSystemId) bool {
+	verifierIdInt := uint8(verifierId)
+	bit := disabledVerifiersBitmap.Uint64() & (1 << verifierIdInt)
```
JuArce, line 17: If the bitmap is uint256, I think it will not fit in a uint64

JuArce (changes requested):

batcher/aligned-batcher/src/lib.rs
IAvecilla: Added it in[e4ff7c4](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/e4ff7c4b3ec51b52f3ad67f0cac613febc89dc99). Just for the record, right now the behavior is exactly the same as the payment service if both service manager and the fallback fails, then the whole batch gets rejected.

batcher/aligned-batcher/src/lib.rs
IAvecilla: Agree, added it in [6d96950](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/6d969506235717d6eb9ecb53a94bdf0e1bd4bae4). I also added some unit tests for this function.

operator/pkg/operator.go
IAvecilla: Good catch, fixed it in [dd8ff97](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/dd8ff97a61673c474afd8361470fd4a9d6f7a5d7)

operator/pkg/utils.go
IAvecilla: Yes, that's correct. The thing is that to perform the bitwise `and` operation, I need to cast the big integer into a `uint64`. We should never have a bitmap larger than 64 bits, since we’ll likely never have more than 63 proving systems. If we accidentally set a bitmap larger than expected, the function will wrap the number and return 0, meaning no verifier will be disabled. I’m not sure if this is the intended behavior, but it seems acceptable.

uri-99 (changes requested):
Blocking until IAvecilla finishes some edge cases

batcher/aligned-batcher/src/lib.rs
```diff
@@ -383,12 +421,40 @@ impl Batcher {
...
 
         // When pre-verification is enabled, batcher will verify proofs for faster feedback with clients
-        if self.pre_verification_is_enabled
-            && !zk_utils::verify(&nonced_verification_data.verification_data).await
-        {
-            error!("Invalid proof detected. Verification failed.");
-            send_message(ws_conn_sink.clone(), ValidityResponseMessage::InvalidProof).await;
-            return Ok(());
+        if self.pre_verification_is_enabled {
+            let disabled_verifiers = match self.disabled_verifiers().await {
```
uri-99: This makes a call to the RPC. In here it should query the internal value, that i think it is updated on every block

batcher/aligned-batcher/src/lib.rs
```diff
@@ -383,12 +421,40 @@ impl Batcher {
...
+                Ok(disabled_verifiers) => disabled_verifiers,
+                Err(e) => {
+                    error!("Failed to get disabled verifiers: {e:?}");
+                    send_message(ws_conn_sink.clone(), ValidityResponseMessage::EthRpcError).await;
+                    return Ok(());
+                }
+            };
+            let verification_data = &nonced_verification_data.verification_data;
+            if zk_utils::is_verifier_disabled(disabled_verifiers, verification_data) {
```
uri-99: You could skip the previous get, and have is_verified_disabled to query the stored value

batcher/aligned-batcher/src/lib.rs
```diff
@@ -950,13 +1028,26 @@ impl Batcher {
...
+        let gas_price_future = self.get_gas_price();
+        let disabled_verifiers_future = self.disabled_verifiers();
+
+        let (gas_price, disable_verifiers) =
+            tokio::join!(gas_price_future, disabled_verifiers_future);
+        let gas_price = gas_price.ok_or(BatcherError::GasPriceError)?;
+        let new_disable_verifiers =
+            disable_verifiers.map_err(|e| BatcherError::DisabledVerifiersError(e.to_string()))?;
+
+        let mut disabled_verifiers_lock = self.disabled_verifiers.lock().await;
```
uri-99: this value could be stored in the batcher, so later checks can query the value and avoid the rpc call

batcher/aligned-batcher/src/lib.rs
```diff
@@ -950,13 +1028,26 @@ impl Batcher {
...
+
+        if new_disable_verifiers != *disabled_verifiers_lock {
+            *disabled_verifiers_lock = new_disable_verifiers;
+            let has_disabled_verifiers = self
+                .queue_has_disabled_verifiers(new_disable_verifiers)
+                .await;
+            if has_disabled_verifiers {
+                self.flush_queue_and_clear_nonce_cache().await;
```
uri-99: if we are going to flush the whole queue if 1 invalidated verifier is present, we could just flush the whole queue without checking if there is 1 present, to avoid reading the whole queue

uri-99 (changes requested):
still missing to review a couple of files

contracts/src/core/AlignedLayerServiceManagerStorage.sol
```diff
@@ -16,7 +16,12 @@ abstract contract AlignedLayerServiceManagerStorage {
 
     address public alignedAggregator;
 
+    // Bitmap representing disabled verifiers
+    // Each verifier is disabled if its corresponding bit is set to 1
+    // The verifier index follows its corresponding value in the `ProvingSystemId` enum being 0 the first verifier.
+    uint256 public disabledVerifiers;
```
uri-99, line 22: should we make sure we set the value as 0? to avoid picking up possible trash that was in the gap.
With a simple call to `setDisabledVerifiers(0)` in the upgrade it should work

operator/pkg/operator.go
```diff
@@ -475,7 +490,13 @@ func (o *Operator) afterHandlingBatchV3(log *servicemanager.ContractAlignedLayer
 	}
 }
 
-func (o *Operator) verify(verificationData VerificationData, results chan bool) {
+func (o *Operator) verify(verificationData VerificationData, disabledVerifiersBitmap *big.Int, results chan bool) {
+	IsVerifierDisabled := IsVerifierDisabled(disabledVerifiersBitmap, verificationData.ProvingSystemId)
+	if IsVerifierDisabled {
+		o.Logger.Infof("Verifier %s is disbled. Skipping verifie", verificationData.ProvingSystemId.String())
```
entropidelic:
```suggestion
		o.Logger.Infof("Verifier %s is disbled. Returning false", verificationData.ProvingSystemId.String())
```
Just to make it a little more explicit

uri-99:
```suggestion
		o.Logger.Infof("Verifier %s is disbled. Skipping verification", verificationData.ProvingSystemId.String())
```

operator/pkg/utils.go
uri-99: I agree with IAvecilla. We are on a tight schedule and we won't have more than 64 verifiers in the near future.

It is a good observation though, I would add a comment in the code about this.

operator/pkg/utils_test.go
```diff
@@ -1,9 +1,55 @@
...
+func TestIsVerifierDisabled(t *testing.T) {
+	t.Run("All verifiers are enabled", func(t *testing.T) {
+		disabledVerifiersBitmap := big.NewInt(0)
+		proving_systems := []common.ProvingSystemId{common.GnarkPlonkBls12_381, common.GnarkPlonkBn254, common.Groth16Bn254, common.SP1, common.Risc0}
+		for _, verifierId := range proving_systems {
+			got := IsVerifierDisabled(disabledVerifiersBitmap, verifierId)
+			want := false
+
+			if got != want {
+				t.Errorf("Verifier %s is disable but it should not", verifierId.String())
```
uri-99:
```suggestion
				t.Errorf("Verifier %s is disabled but it shouldn't be", verifierId.String())
```

operator/pkg/utils_test.go
```diff
@@ -1,9 +1,55 @@
...
+	t.Run("All verifiers are disabled", func(t *testing.T) {
+		// This is the bitmap for all verifiers disabled since it is 11111 in binary.
+		disabledVerifiersBitmap := big.NewInt(31)
+		proving_systems := []common.ProvingSystemId{common.GnarkPlonkBls12_381, common.GnarkPlonkBn254, common.Groth16Bn254, common.SP1, common.Risc0}
+		for _, verifierId := range proving_systems {
+			got := IsVerifierDisabled(disabledVerifiersBitmap, verifierId)
+			want := true
+
+			if got != want {
+				t.Errorf("Verifier %s is enabled but it should not", verifierId.String())
```
uri-99:
```suggestion
				t.Errorf("Verifier %s is enabled but it shouldn't be", verifierId.String())
```

operator/pkg/utils_test.go
```diff
@@ -1,9 +1,55 @@
...
+	t.Run("Some verifiers are disabled", func(t *testing.T) {
+		// This is the bitmap for the first and last verifiers disabled since it is 10001 in binary.
+		disabledVerifiersBitmap := big.NewInt(17)
+		proving_systems := []common.ProvingSystemId{common.GnarkPlonkBls12_381, common.GnarkPlonkBn254, common.Groth16Bn254, common.SP1, common.Risc0}
+		for _, verifierId := range proving_systems {
+			got := IsVerifierDisabled(disabledVerifiersBitmap, verifierId)
+			want := verifierId == common.GnarkPlonkBls12_381 || verifierId == common.Risc0
+
+			if got != want {
+				t.Errorf("Verifier %s is enabled but it should not", verifierId.String())
```
uri-99:
```suggestion
				t.Errorf("Verifier %s is enabled but it shouldn't be", verifierId.String())
```

uri-99 (changes requested):

batcher/aligned-batcher/src/lib.rs
```diff
@@ -950,13 +1027,21 @@ impl Batcher {
...
+        let gas_price = gas_price.ok_or(BatcherError::GasPriceError)?;
+        let new_disable_verifiers =
+            disable_verifiers.map_err(|e| BatcherError::DisabledVerifiersError(e.to_string()))?;
+
+        let mut disabled_verifiers_lock = self.disabled_verifiers.lock().await;
+
+        if new_disable_verifiers != *disabled_verifiers_lock {
+            *disabled_verifiers_lock = new_disable_verifiers;
+            self.flush_queue_and_clear_nonce_cache().await;
+        }
```
entropidelic: probably the `disabled_verifiers_lock` should be dropped to avoid unnecesary locking

entropidelic: it could good to encapsulate this part in a scope `{}`

entropidelic: nvm, I added it 👍

operator/pkg/operator.go
```diff
@@ -475,7 +490,13 @@ func (o *Operator) afterHandlingBatchV3(log *servicemanager.ContractAlignedLayer
 	}
 }
 
-func (o *Operator) verify(verificationData VerificationData, results chan bool) {
+func (o *Operator) verify(verificationData VerificationData, disabledVerifiersBitmap *big.Int, results chan bool) {
+	IsVerifierDisabled := IsVerifierDisabled(disabledVerifiersBitmap, verificationData.ProvingSystemId)
+	if IsVerifierDisabled {
+		o.Logger.Infof("Verifier %s is disbled. Skipping verification", verificationData.ProvingSystemId.String())
```
entropidelic:
```suggestion
		o.Logger.Infof("Verifier %s is disabled. Returning false", verificationData.ProvingSystemId.String())
```

entropidelic (approved):

batcher/aligned-batcher/src/lib.rs
IAvecilla: Done! [16398e0](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/16398e0396b6e08b8100de26e9e215ffcdf056ee)

batcher/aligned-batcher/src/lib.rs
IAvecilla: Done! [16398e0](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/16398e0396b6e08b8100de26e9e215ffcdf056ee)

batcher/aligned-batcher/src/lib.rs
IAvecilla: This was already present in the batcher, now we used as the primary source instead of calling the service manager everytime

batcher/aligned-batcher/src/lib.rs
IAvecilla: Yeah, it could be a good first version for this. We can make sure that future versions checks for the proofs in the queue instead of flushing all of them everytime

contracts/src/core/AlignedLayerServiceManagerStorage.sol
IAvecilla: I did the changes in this commit [8accb46](https://github.com/yetanotherco/aligned_layer/pull/1090/commits/8accb46eaa56a1b6f125909601e6859e3b19871b), please make sure that is the intended way to do this. I also updated the anvil state

uri-99: Awesome, we're just missing the make targets

uri-99 (approved):

JuArce (approved):

PR #1105, opened 2024-09-26
feat(telemetry): modify the aggregator to use the elixir Tracker

avilagaston9:
**Motivation**

This PR introduces the basic structure for telemetry implementation in the aggregator.

**Description**

- Creates the `Telemetry` struct with the following methods:
    - `InitTaskTrace()`
    - `LogOperatorResponse()`
    - `LogQuorumReached()`
    - `LogTaskError()`
    - `FinishTaskTrace()`
- Each method sends an `HTTP POST` request to the new Elixir Tracker.
- Integrates the `Telemetry` struct into the `Aggregator` and invokes its methods as needed.
- This PR also implements the missing endpoints in the Elixir Tracker.

image

### DIFF
--- a/aggregator/internal/pkg/aggregator.go
+++ b/aggregator/internal/pkg/aggregator.go
@@ -12,7 +12,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/yetanotherco/aligned_layer/metrics"
 
-	"github.com/Layr-Labs/eigensdk-go/chainio/clients"
 	sdkclients "github.com/Layr-Labs/eigensdk-go/chainio/clients"
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/Layr-Labs/eigensdk-go/services/avsregistry"
@@ -86,8 +85,12 @@ type Aggregator struct {
 
 	logger logging.Logger
 
+	// Metrics
 	metricsReg *prometheus.Registry
 	metrics    *metrics.Metrics
+
+	// Telemetry
+	telemetry *Telemetry
 }
 
 func NewAggregator(aggregatorConfig config.AggregatorConfig) (*Aggregator, error) {
@@ -125,7 +128,7 @@ func NewAggregator(aggregatorConfig config.AggregatorConfig) (*Aggregator, error
 	aggregatorPrivateKey := aggregatorConfig.EcdsaConfig.PrivateKey
 
 	logger := aggregatorConfig.BaseConfig.Logger
-	clients, err := clients.BuildAll(chainioConfig, aggregatorPrivateKey, logger)
+	clients, err := sdkclients.BuildAll(chainioConfig, aggregatorPrivateKey, logger)
 	if err != nil {
 		logger.Errorf("Cannot create sdk clients", "err", err)
 		return nil, err
@@ -154,6 +157,9 @@ func NewAggregator(aggregatorConfig config.AggregatorConfig) (*Aggregator, error
 	reg := prometheus.NewRegistry()
 	aggregatorMetrics := metrics.NewMetrics(aggregatorConfig.Aggregator.MetricsIpPortAddress, reg, logger)
 
+	// Telemetry
+	aggregatorTelemetry := NewTelemetry(aggregatorConfig.Aggregator.TelemetryIpPortAddress, logger)
+
 	nextBatchIndex := uint32(0)
 
 	aggregator := Aggregator{
@@ -176,6 +182,7 @@ func NewAggregator(aggregatorConfig config.AggregatorConfig) (*Aggregator, error
 		logger:                logger,
 		metricsReg:            reg,
 		metrics:               aggregatorMetrics,
+		telemetry:             aggregatorTelemetry,
 	}
 
 	return &aggregator, nil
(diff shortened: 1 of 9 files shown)


aggregator/internal/pkg/aggregator.go
```diff
@@ -250,17 +268,15 @@ func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsA
...
-	taskCreatedBlock := agg.batchCreatedBlockByIdx[blsAggServiceResp.TaskIndex]
 
 	// Delete the task from the map
 	delete(agg.operatorRespondedBatch, blsAggServiceResp.TaskIndex)
 
-	agg.AggregatorConfig.BaseConfig.Logger.Info("- Unlocked Resources: Fetching merkle root")
+	agg.AggregatorConfig.BaseConfig.Logger.Info("- Unlocked Resources: Freeing task data")
 	agg.taskMutex.Unlock()
 
+	defer agg.telemetry.LogQuorumReached(batchData.BatchMerkleRoot)
```
JuArce: I should not use defer in this case

aggregator/internal/pkg/server.go
```diff
@@ -54,6 +54,7 @@ func (agg *Aggregator) ProcessOperatorSignedTaskResponseV2(signedTaskResponse *t
 		"SenderAddress", "0x"+hex.EncodeToString(signedTaskResponse.SenderAddress[:]),
 		"BatchIdentifierHash", "0x"+hex.EncodeToString(signedTaskResponse.BatchIdentifierHash[:]),
 		"operatorId", hex.EncodeToString(signedTaskResponse.OperatorId[:]))
+	defer agg.telemetry.LogOperatorResponse(signedTaskResponse.BatchMerkleRoot, signedTaskResponse.OperatorId)
```
JuArce: I should not use defer in this case

telemetry_api/lib/telemetry_api/traces.ex
```diff
@@ -21,22 +21,148 @@ defmodule TelemetryApi.Traces do
...
+          "Operator ID: #{operator_id}",
+          [
+            {:merkle_root, merkle_root},
+            {:operator_id, operator_id}
+          ]
+        )
+
+        ctx = Ctx.get_current()
+
+        TraceStore.store_trace(
```
JuArce: Is it needed update the context on every event?

JuArce (changes requested):

aggregator/internal/pkg/aggregator.go
avilagaston9: Done!

aggregator/internal/pkg/server.go
avilagaston9: Done!

telemetry_api/lib/telemetry_api/traces.ex
avilagaston9: It's not, you are right! Updated!

aggregator/internal/pkg/aggregator.go
```diff
@@ -250,17 +268,15 @@ func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsA
 	}
 
 	agg.taskMutex.Lock()
```
JulianVentura, line 271: Aren't we missing a log `Logger.Info("- Locked Resources: task data")` here? I also think that both log messages (lock and unlock) should have the same resource name, so it should say something like `Logger.Info("- Unlocked Resources: task data")`

aggregator/internal/pkg/aggregator.go
```diff
@@ -216,9 +223,20 @@ func (agg *Aggregator) Start(ctx context.Context) error {
...
+	batchData := agg.batchDataByIdentifierHash[batchIdentifierHash]
+	taskCreatedBlock := agg.batchCreatedBlockByIdx[blsAggServiceResp.TaskIndex]
+	agg.taskMutex.Unlock()
+	agg.AggregatorConfig.BaseConfig.Logger.Info("- Unlocked Resources: Fetching task data")
+
+	// Finish task trace once the task is processed (either successfully or not)
+	defer agg.telemetry.FinishTrace(batchData.BatchMerkleRoot)
+
 	if blsAggServiceResp.Err != nil {
+		agg.telemetry.LogTaskError(batchData.BatchMerkleRoot, blsAggServiceResp.Err)
```
JuArce, line 238: There is missing log error if the transaction fails in ethereum (L309)

aggregator/internal/pkg/telemetry.go
```diff
@@ -0,0 +1,133 @@
...
+		TaskError:  taskError.Error(),
+	}
+	if err := t.sendTelemetryMessage("/api/taskError", body); err != nil {
+		t.logger.Error("[Telemetry] Error in LogTaskError", "error", err)
+	}
+}
+
+func (t *Telemetry) FinishTrace(batchMerkleRoot [32]byte) {
+	go func() {
+		time.Sleep(10 * time.Second)
```
JulianVentura, line 97: Why is this sleep here?

JuArce (changes requested):

aggregator/internal/pkg/telemetry.go
avilagaston9: This is to wait for all operator responses, even after the quorum is reached. I will add a comment for clarification!

JuArce: Because the BLS aggregation service returns immediately the quorum is reached and it is not necessary when all operators have responded the task. So, if we finish the trace immediately, we will mark working operators as not responding, that is wrong. 
So, we close the trace with a delay to store most operators that responded to the task

aggregator/internal/pkg/aggregator.go
avilagaston9: You are right! We were missing it. Thank you!

aggregator/internal/pkg/aggregator.go
avilagaston9: Done!

JuArce (approved):

PR #1106, opened 2024-09-27
refactor(batcher): add user state & other code quality refactors

entropidelic:
This PR refactors the batcher codebase, hopefully making it more readable.

A `UserState` struct has been created to handle the cached state of each user who sent some proof to the batch, with the information about its nonce, the number of proofs in the batch, and the minimum fee of all the proofs that it has submitted and are currently in the batch.

The `handle_message` function has been refactored to be more clear in its steps:
1. Verify integrity of the message (checking correct serialization, verify signature, verify proof length and that it is valid, etc)
2. Perform validations related to the user state (check the user has enough balance for paying all proofs in the batch, check nonce is correct, etc)
3. Add the new proof to the batch, updating the user state too.

Since the batch queue and the user state are updated at the same time, this should get rid of the vulnerability highlited by Least Authority in issue #1081.

## Testing
Test that everything works as expected, try sending proofs with both paying and nonpaying addresses. Everything should work as expected.

**NOTE**: Since there were a lot of changes, it could be useful for reviewing to check the whole `handle_message` function and try to understand each step, which should be clearer now, and validate that it makes sense. Give it special attention since is where the most changes have been done


batcher/aligned-batcher/src/types/user_state.rs
```diff
@@ -0,0 +1,29 @@
+use ethers::types::U256;
+
+pub(crate) struct UserState {
+    pub nonce: Option<U256>,
+    /// The minimum fee of a pending proof for a user.
+    /// This should always be the fee of the biggest pending nonce by the user.
+    /// This is used to check if a user is submitting a proof with a higher nonce and higher fee,
```
Oppen, line 7: If this is a minimum fee, wouldn't the invalid scenario be a lower fee?

batcher/aligned-batcher/src/lib.rs
```diff
@@ -667,61 +519,48 @@ impl Batcher {
     /// Returns true if the message was replaced in the batch, false otherwise
     async fn handle_replacement_message(
         &self,
-        mut batch_state: tokio::sync::MutexGuard<'_, BatchState>,
+        mut batch_state_lock: MutexGuard<'_, BatchState>,
```
Oppen, line 522: This lock needs to be a `tokio::sync` one, as it is held across an `await` point. Otherwise, you'll deadlock.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -753,43 +615,81 @@ impl Batcher {
     /// Adds verification data to the current batch queue.
     async fn add_to_batch(
         &self,
-        mut batch_state: tokio::sync::MutexGuard<'_, BatchState>,
+        mut batch_state_lock: MutexGuard<'_, BatchState>,
```
Oppen, line 665: I believe receiving the mutable reference (`&mut BatchState`) would work and is simpler.

batcher/aligned-batcher/src/types/user_state.rs
entropidelic: the wording in this part is a bit confusing, but the invalid scenario is when a user submits a proof with some max fee N, and then submits a proof with a max fee M, with M > N.

Since the BatchPriorityQueue orders entries by max fee, the one with the higher max fee will be processed first, but this conflicts with the nonce ordering.
So, we restrict messages that have higher nonce to have a lower max fee than the one with a lower nonce

batcher/aligned-batcher/src/lib.rs
Oppen: That said, it might actually be a great idea to have access to the handle here, since we're sending the error messages here: we should drop the lock before sending the error, otherwise we'll be waiting on the network with the lock taken, not handling other messages in the meantime.

Oppen: Alternatively, it may be cleaner to leave the response to one of the higher layers, and only return errors in the core logic.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -455,206 +341,172 @@ impl Batcher {
...
+    // Checks user has sufficient balance for paying all its the proofs in the current batch.
+    async fn check_min_balance(&self, addr: &Address, user_proofs_in_batch: usize) -> bool {
+        let Some(user_balance) = self.get_user_balance(addr).await else {
+            return false;
+        };
+        let min_balance = U256::from(user_proofs_in_batch) * U256::from(MIN_FEE_PER_PROOF);
+        if user_balance < min_balance {
+            return false;
+        }
         true
```
IAvecilla: I think you can do `user_balance >= min_balance`

batcher/aligned-batcher/src/lib.rs
```diff
@@ -667,61 +519,48 @@ impl Batcher {
...
 
+        // The replacement entry is built from the old entry and validated for then to be replaced
+        let mut replacement_entry = entry.clone();
         replacement_entry.signature = signature;
         replacement_entry.verification_data_commitment =
             nonced_verification_data.verification_data.clone().into();
         replacement_entry.nonced_verification_data = nonced_verification_data;
 
-        // close old sink and replace with new one
+        // Close old sink in old entry replace it with new one
```
IAvecilla:
```suggestion
        // Close old sink in old entry and replace it with the new one
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -731,16 +570,39 @@ impl Batcher {
...
+        batch_state_lock.batch_queue.push(
+            replacement_entry.clone(),
+            BatchQueueEntryPriority::new(replacement_max_fee, nonce),
+        );
+
+        let updated_min_fee_in_batch = batch_state_lock.get_user_min_fee_in_batch(&addr);
+        let Some(_) = batch_state_lock.update_user_min_fee(&addr, updated_min_fee_in_batch) else {
+            warn!("User state for address {addr:?} was not present in batcher user states, but it should be");
+            return;
+        };
```
IAvecilla: Since you don't need the inner value you can just do 
`if batch_state_lock.update_user_min_fee(&addr, updated_min_fee_in_batch).is_none {...}`

batcher/aligned-batcher/src/lib.rs
```diff
@@ -753,43 +615,81 @@ impl Batcher {
...
+            .await
+        else {
+            error!("User state of address {proof_submitter_addr} was not found when trying to update user state. This user state should have been present");
+            return Err(BatcherError::AddressNotFoundInUserStates(
+                proof_submitter_addr,
+            ));
+        };
+
+        // User state is updated
+        let Some(_) = batch_state_lock.update_user_state(
```
IAvecilla: Same as above, since you are not using the inner value maybe you can just use `is_none` function

batcher/aligned-batcher/src/lib.rs
```diff
@@ -832,29 +731,44 @@ impl Batcher {
...
+        let (resulting_batch_queue, finalized_batch) =
+            batch_queue::try_build_batch(batch_queue_copy, gas_price, self.max_batch_size)
+                .inspect_err(|e| {
+                    *batch_posting = false;
+                    match e {
+                        // We can't post a batch since users are not willing to pay the needed fee, wait for more proofs
+                        BatcherError::BatchCostTooHigh => {
+                            info!("No working batch found. Waiting for more proofs")
+                        }
+                        // FIXME: We should refactor this code and instead of returning None, return an error.
```
IAvecilla, line 796: I think this comment is staled

batcher/aligned-batcher/src/lib.rs
```diff
@@ -832,29 +731,44 @@ impl Batcher {
...
+                        e => error!("Unexpected error: {:?}", e),
+                    }
+                })
+                .ok()?;
+
+        batch_state_lock.batch_queue = resulting_batch_queue;
+        let updated_user_proof_count_and_min_fee =
+            batch_state_lock.get_user_proofs_in_batch_and_min_fee();
+
+        let user_addresses: Vec<Address> = batch_state_lock.user_states.keys().cloned().collect();
```
Oppen, line 807: I believe you're only iterating the addresses once. The `collect` is unnecessary.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -455,206 +341,177 @@ impl Batcher {
...
+        // we replace the old entry with the new from the replacement message.
+        if expected_nonce > msg_nonce {
+            info!("Possible replacement message received: Expected nonce {expected_nonce:?} - message nonce: {msg_nonce:?}");
+            self.handle_replacement_message(
+                batch_state_lock,
+                nonced_verification_data,
+                ws_conn_sink.clone(),
+                client_msg.signature,
+                addr,
+            )
```
Oppen, line 504: Let's make sure to also drop the lock inside this function before responding to the network.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -455,206 +341,177 @@ impl Batcher {
...
+            .add_to_batch(
+                batch_state_lock,
+                nonced_verification_data,
+                ws_conn_sink.clone(),
+                client_msg.signature,
+                addr,
+            )
+            .await
+        {
+            error!("Error while adding entry to batch: {e:?}");
```
Oppen, line 538: Missing the drop.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -455,206 +341,177 @@ impl Batcher {
...
-        self.add_to_batch(
-            batch_state,
-            nonced_verification_data,
-            ws_conn_sink.clone(),
-            signature,
-            addr,
-        )
-        .await;
+        info!("Verification data message handled");
+        send_message(ws_conn_sink, ValidityResponseMessage::Valid).await;
```
Oppen, line 544: We should probably drop early here too.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -667,61 +524,48 @@ impl Batcher {
...
+        ws_conn_sink: WsMessageSink,
         signature: Signature,
         addr: Address,
-        expected_user_nonce: U256,
-    ) -> bool {
+    ) {
         let replacement_max_fee = nonced_verification_data.max_fee;
         let nonce = nonced_verification_data.nonce;
+        let Some(entry) = batch_state_lock.get_entry(addr, nonce) else {
+            warn!("Invalid nonce for address {addr}. Queue entry with nonce {nonce} not found");
```
Oppen, line 573: We should drop the lock before answering.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -667,61 +524,48 @@ impl Batcher {
...
-                        entry.nonced_verification_data.max_fee, replacement_max_fee
-                    );
-                    send_message(
-                        ws_conn_sink.clone(),
-                        ValidityResponseMessage::InvalidReplacementMessage,
-                    )
-                    .await;
+        let original_max_fee = entry.nonced_verification_data.max_fee;
+        if original_max_fee > replacement_max_fee {
+            warn!("Invalid replacement message for address {addr}, had fee {original_max_fee:?} < {replacement_max_fee:?}");
```
Oppen, line 581: Ditto.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -731,16 +575,39 @@ impl Batcher {
         }
 
         replacement_entry.messaging_sink = Some(ws_conn_sink.clone());
-        if let Some(msg) = batch_state.validate_and_increment_max_fee(replacement_entry) {
+        if !batch_state_lock.replacement_entry_is_valid(&replacement_entry) {
             warn!("Invalid max fee");
-            send_message(ws_conn_sink.clone(), msg).await;
-            return false;
+            send_message(
```
Oppen, line 619: Ditto

batcher/aligned-batcher/src/lib.rs
entropidelic: here the lock is passed to the `add_to_batch` function so that function takes ownership of it

batcher/aligned-batcher/src/types/batch_state.rs
```diff
@@ -0,0 +1,159 @@
...
+    ) -> Option<(U256, U256, usize)> {
+        let updated_nonce = self.update_user_nonce(addr, new_nonce);
+        let updated_min_fee = self.update_user_min_fee(addr, new_min_fee);
+        let updated_proof_count = self.update_user_proof_count(addr, new_proof_count);
+
+        match (updated_nonce, updated_min_fee, updated_proof_count) {
+            (Some(_), Some(_), Some(_)) => Some((new_nonce, new_min_fee, new_proof_count)),
+            _ => None,
+        }
+    }
```
IAvecilla: Same as above, if you're not using the inner value maybe you can check those variables with `is_some()`.

IAvecilla (changes requested):

Makefile
```diff
@@ -271,7 +271,7 @@ batcher_send_risc0_burst:
 		--proof ../../scripts/test_files/risc_zero/fibonacci_proof_generator/risc_zero_fibonacci.proof \
         --vm_program ../../scripts/test_files/risc_zero/fibonacci_proof_generator/fibonacci_id.bin \
         --public_input ../../scripts/test_files/risc_zero/fibonacci_proof_generator/risc_zero_fibonacci.pub \
-        --repetitions $(BURST_SIZE) \
+        --repetitions 500 \
```
uri-99: I think you didn't mean this change

Makefile
```diff
@@ -361,7 +361,8 @@ batcher_send_halo2_ipa_task_burst_5: batcher/target/release/aligned
 		--proof ../../scripts/test_files/halo2_ipa/proof.bin \
 		--public_input ../../scripts/test_files/halo2_ipa/pub_input.bin \
 		--vk ../../scripts/test_files/halo2_ipa/params.bin \
-		--repetitions 5 \
+		--repetitions 500 \
```
uri-99: here also using BURST_SIZE may be more comfortable

Makefile
```diff
@@ -361,7 +361,8 @@ batcher_send_halo2_ipa_task_burst_5: batcher/target/release/aligned
...
+		--private_key 0xdbda1821b80551c9d65939329250298aa3472ba22feea921c0cf5d620ea67b97 \
```
uri-99: Was this intentional? hardcoding a private key like this seems prone to errors when maintaining. If needed, i would consider adding a var on top and using its value

batcher/aligned-batcher/src/lib.rs
```diff
@@ -433,11 +318,12 @@ impl Batcher {
                 return Ok(());
             }
         };
+        let msg_nonce = client_msg.verification_data.nonce;
+        info!("Received message with nonce: {msg_nonce:?}",);
```
uri-99: Of all the things that can be printed here, msg_nonce seems quite odd. Does this log add any value? Is there a better thing to log?

batcher/aligned-batcher/src/lib.rs
```diff
@@ -455,206 +341,177 @@ impl Batcher {
...
-                let nonced_verification_data = client_msg.verification_data;
-                if nonced_verification_data.verification_data.proof.len() > self.max_proof_size {
-                    error!("Proof size exceeds the maximum allowed size.");
-                    send_message(ws_conn_sink.clone(), ValidityResponseMessage::ProofTooLarge)
-                        .await;
-                    return Ok(());
-                }
+        let nonced_verification_data = client_msg.verification_data.clone();
+        if nonced_verification_data.verification_data.proof.len() > self.max_proof_size {
+            error!("Proof size exceeds the maximum allowed size.");
```
uri-99: if the `proof.len` is being checked after a `.clone()` , could it be the batcher clones a n sized proof before checking its size?

batcher/aligned-batcher/src/lib.rs
```diff
@@ -731,16 +577,41 @@ impl Batcher {
         }
 
         replacement_entry.messaging_sink = Some(ws_conn_sink.clone());
-        if let Some(msg) = batch_state.validate_and_increment_max_fee(replacement_entry) {
+        if !batch_state_lock.replacement_entry_is_valid(&replacement_entry) {
+            std::mem::drop(batch_state_lock);
             warn!("Invalid max fee");
```
uri-99: "Invalid Replacement Message" sounds more appropriate, since it is not necessarily only because of max fee

batcher/aligned-batcher/src/lib.rs
```diff
@@ -455,206 +341,183 @@ impl Batcher {
...
+        // *        Perform validations over user state         *
+        // * ---------------------------------------------------*
+
+        // For now on until the message is fully processed, the batch state is locked
+        // This is needed because we need to query the user state to make validations and
+        // finally add the proof to the batch queue.
+
+        // At this point, we will have a user state for sure, since we have inserted it
+        // if not already present. It should be safe to call `unwrap()` here.
+        let batch_state_lock = self.batch_state.lock().await;
```
entropidelic, line 461: **NOTE**: From here until the whole message is processed, the batch state is locked. This means that the `batch_queue` **and** the `users_states` are both locked. I tried to decouple both but faced a lot of deadlocks. This might be a skill issue from my side, but didn't want to take so long on this PR and over complexify things. This could be explored and tackled in another PR that just focuses on that.
For the time being, this just replicates the behavior we had until this moment, since we were already locking all the batch state:
https://github.com/yetanotherco/aligned_layer/blob/d6c2e34ecce7b91afe6e148d9f216ffec41007b2/batcher/aligned-batcher/src/lib.rs#L503

Some checks were moved to be done _before_ the locking, so that the time the lock is held is minimized

batcher/aligned-batcher/src/lib.rs
entropidelic: solved in 9f7e6957e93a70c6b7954a1933f5dcd7c3f4f3b6

batcher/aligned-batcher/src/lib.rs
entropidelic: solved in 68bb54e1c1e52c0b5ac707f7dcac0d4888fea7f7

batcher/aligned-batcher/src/lib.rs
entropidelic: solved in ddb9fd1e5672d344ee842be7ec84419bb54f4543

batcher/aligned-batcher/src/lib.rs
entropidelic: it still holds, we are returning an `Option` when it would be better to return a `Result`

uri-99 (changes requested):

Makefile
entropidelic: my bad, pushed these by mistake

Makefile
entropidelic: my bad, pushed these by mistake

Makefile
entropidelic: my bad, pushed these by mistake

Makefile
entropidelic: solved in 37a9a0ba03d86a5804a27f272fb84190501d43ba

Makefile
entropidelic: solved in 37a9a0ba03d86a5804a27f272fb84190501d43ba

batcher/aligned-batcher/src/lib.rs
entropidelic: solved in 3f0e36277913b9a52859f36b2fe4209879f823e6

batcher/aligned-batcher/src/lib.rs
entropidelic: solved in 3f0e36277913b9a52859f36b2fe4209879f823e6

batcher/aligned-batcher/src/lib.rs
entropidelic: changed to `debug` in 3f0e36277913b9a52859f36b2fe4209879f823e6

batcher/aligned-batcher/src/lib.rs
entropidelic: fixed in bfccf27fe7393a6ac283b50af418aef5348ff5f9

batcher/aligned-batcher/src/lib.rs
entropidelic: I left it as we talked yesterday

batcher/aligned-batcher/src/lib.rs
entropidelic: the `add_to_batch` function takes ownership of the lock so it is dropped inside of it

IAvecilla (approved):

uri-99 (approved):

batcher/aligned-batcher/src/lib.rs
```diff
@@ -473,208 +357,188 @@ impl Batcher {
...
-
-                                return Ok(());
-                            }
-                        };
-
-                        batch_state.user_nonces.insert(addr, user_nonce);
-                        user_nonce
-                    }
-                };
+        if self.user_balance_is_unlocked(&addr).await {
```
Oppen, line 399: I believe this needs to be protected by the state lock to avoid TOCTTOU where a user unlocks its funds after this check is done.

batcher/aligned-batcher/src/lib.rs
Oppen: Oh, right! We're no longer locked here then.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1159,204 +1113,140 @@ impl Batcher {
...
-
-                        vacant.insert(nonce)
-                    }
-                };
-
-                info!("non paying nonce: {:?}", nonpaying_nonce);
+        let batch_state_lock = self.batch_state.lock().await;
+        let Some(non_paying_nonce) = batch_state_lock.get_user_nonce(&replacement_addr).await
+        else {
+            error!("Nonce for non-paying address {replacement_addr:?} not found in cache.");
```
Oppen, line 1169: Release the lock before responding.

batcher/aligned-batcher/src/lib.rs
entropidelic: forgot to answer here, but same as in the other cases, the `handle_replacemente_message` function takes ownership of the lock and drops it

batcher/aligned-batcher/src/lib.rs
entropidelic: solved in 95897f1863fa9a188f4af93cf37e7489f7fbaa4b

batcher/aligned-batcher/src/lib.rs
Oppen: This kind of comment is better placed in the code itself.

batcher/aligned-batcher/src/lib.rs
Oppen: @entropidelic PING: remember to comment on why this is safe to do without the state lock.

Oppen (approved):

batcher/aligned-batcher/src/lib.rs
entropidelic: added in 166509ba09940075c4c6d9a6a8a11990b5513c4b

PR #1110, opened 2024-09-27
feat: migrate cbor lib and add Fuzzer for cbor serialization

### LINKED ISSUE #902: Marshall/Unmarshall conflict

### LINKED ISSUE #1093: feat: migrate cbor codec to fxamacker/cbor/v2 to avoid crashes
ugorji/go/codec doesn't seem to sanitize its input, leading to OOM crashes on forged input.
fxamacker/cbor/v2 seems to handle that just fine.

IAvecilla:
This:
closes 
#1093
https://github.com/yetanotherco/aligned_layer/issues/902

To test, run the following commands:
-  `make operator_marshall_unmarshall_fuzz_macos` this tests that both decoding and encoding functions don't panic with random inputs without checking for correctness.
- `make operator_valid_marshall_fuzz_macos` this tests the encoding function correctness checking for correct results only dealing with valid inputs.


uri-99 (approved):
Tested on Mac and Linux

MarcosNicolau (approved):
Tested on my mac.

MauroToscano (approved):

PR #1114, opened 2024-09-30
refactor(contracts): remove merkle root and signature checks on payment service contract

### LINKED ISSUE #1119: Remove checks in Solidity Payment Contracts, measure costs
Most expensive check should be the merkle tree.

We should measure costs without it
Then check the costs without any signed data

MarcosNicolau:
**Changes**
Remove the merkle root and signature checks to save on gas in the payment service contract.

**Gas changes**

Here is the gas diff between validating the merkle root and signature and not doing it:

| Proofs | Without any checks  | With merkle root and signature checks |  Diff|
|----------|-----------------|-----------------|-----|
| 10   | 136,752.00       | 251,792.00       |+- 115,040.00|
| 35   | 169,530.00       | 505,330.00       |+-335,800.00|
| 100  | 254,765.00      | 1,153,964.00      |+- 899,199.00|
| 328  | 561,762.00       | 3,754,187.00      |+- 3,192,425.00|

For a more complete report see the google sheet [here](https://docs.google.com/spreadsheets/d/1Sw1COtCdJ50HkdwseCXcXeN095KOhKRGY6Y7vqkW19M/edit?usp=sharing).

The report can be reproduced in the following way:
1. Replace the current code from [here](https://github.com/yetanotherco/aligned_layer/blob/014bd4d7d64be442c24a8235fda4700bd225a69f/batcher/aligned-batcher/src/lib.rs#L1084C13-L1087C14) with the following code to log the gas report from that batch:
```rust
Ok(receipt) => {
    info!("Batch verification task created on Aligned contract");
    // putting an error for easier debugging!
    error!(
        "Batch statistics:\n\t- Proofs: {} \n\t- Gas used: {}",
        num_proofs_in_batch,
        receipt.gas_used.unwrap()
    );
    Ok(())
}
```
3. Run the following proofs from this pr to get the results without checks and from this commit `014bd4d7d64be442c24a8235fda4700bd225a69f` to run with the checks:
    - `BURST_SIZE=10 make batcher_send_sp1_burst`
    - `make batcher_send_burst_groth16`: Here you'll have to wait as it sends proofs non-stop.
    - `BURST_SIZE=100 make batcher_send_risc0_burst`
    - `BURST_SIZE=1000 make batcher_send_risc0_burst`: Here the max proofs per batch will be around `300`-`380`, but it is difficult to get the same number in both executions.
  
Results may vary between executions but should be around the same on average.

Closes #1119 

Edit: the table was update to the commit where all the checks are removed, not only the merkle and signature.


uri-99 (changes requested):
We can still remove the usage of `nonce` and `maxFee` from the contracts to make it cheaper, `ProofSubmitterData` can be only the address

MarcosNicolau:
> We can still remove the usage of nonce and maxFee from the contracts to make it cheaper, ProofSubmitterData can be only the address

I addressed your comments. I also updated the gas usage results.

contracts/src/core/BatcherPaymentServiceStorage.sol
uri-99, line 1: if `ProofSubmitterData` is going to be only 1 address, it should be quite cheaper to just pass the address. It will save reading/writing inside the struct.

uri-99 (changes requested):

contracts/src/core/BatcherPaymentServiceStorage.sol
MarcosNicolau: Updated!

contracts/src/core/BatcherPaymentService.sol
```diff
@@ -44,11 +44,11 @@ contract BatcherPaymentService is
     error InvalidSignature(); // 8baa579f
     error InvalidNonce(uint256 expected, uint256 actual); // 06427aeb
     error InvalidMaxFee(uint256 maxFee, uint256 actualFee); // f59adf4a
-    error SignerInsufficientBalance(
+    error SubmissionInsufficientBalance(
         address signer,
```
uri-99: it is no longer a signer

contracts/src/core/BatcherPaymentService.sol
uri-99, line 1: remove unused errors
`NoLeavesSubmitted`
`NotEnoughLeaves`
etc.

refactor to appropriate new names:
`NoProofSubmitterSignatures`
etc.

uri-99, line 1: remove EIP712 related things, since it is for signatures

uri-99, line 1: remove user_nonces

uri-99, line 1: remove `noncedVerificationDataTypeHash` things

uri-99 (changes requested):

uri-99 (approved):

contracts/src/core/BatcherPaymentServiceStorage.sol
uri-99, line 1: update GAP size

uri-99 (changes requested):

uri-99 (changes requested):
needs changes on docs

entropidelic:
@uri-99 I updated the docs in [docs/2_architecture/components/2_payment_service_contract.md](https://github.com/yetanotherco/aligned_layer/pull/1114/files#diff-79cfd0b19de5a783364b561762e20f58bedb3fda8f92235523f6036f7d3ae336)

entropidelic (approved):

uri-99 (approved):

PR #1000, opened 2024-09-18
refactor(tracker): rewrite operator tracker in elixir

JuArce:
# Description
Rewriting of the operator tracker in elixir

# Tasks

- [x] CRUD for operators
- [x] Dockerfile for database
- [x] Validate ECDSA signature from operators
- [x] Add README
- [x] Add targets to makefile
- [x] Check if operator is registered in RegistryCoordinator
- [x] Update operator version if already exist in database 
- [x] Fetch operators from Ethereum
- [ ] Error management (addressed in #1075 )

# How to test

1. Start the database
```
make telemetry_run_db
```

## Run aligned

1. Run anvil
```
make anvil_start_with_block_time
```

2. Run aggregator
```
make aggregator_start
```

3. Register operator
```
export OPERATOR_ADDRESS=0x70997970C51812dc3A010C7d01b50e0d17dc79C8
make operator_full_registration CONFIG_FILE=config-files/config-operator-1.yaml
```

> [!NOTE]  
> There isn't necessary to run the batcher, since we are not sending proofs.

## Run tracker

Once you registered the operator, the tracker will fetch registered operators on start.

1. Run
```
make telemetry_create_env
make telemetry_start
```

2. Go to [tracker](http://localhost:4001/api/operators)

You should see the following response

```json
[
    {
        "id": "0x12027b4b27251854a8c58ea5df60bc0dbb57f97d474df0ba8bd33842f20a0e9e",
        "name": "Aligned Layer Operator",
        "version": null, <- Operator version is not set
        "address": "0x70997970C51812DC3A010C7D01B50E0D17DC79C8",
        "stake": "100000000000000000"
    }
]
```

## Send Operator version

1. Build the operator
```
make build_operator
```

2. Run the operator
```
./operator/build/aligned-operator start --config ./config-files/config-operator-1.yaml
```

3. Go to [tracker](http://localhost:4001/api/operators) again.
You should see the entry updated with the operator version

```json
[
    {
        "id": "0x12027b4b27251854a8c58ea5df60bc0dbb57f97d474df0ba8bd33842f20a0e9e",
        "name": "Aligned Layer Operator",
        "version": "v0.5.2", <- Operator version is correctly set
        "address": "0x70997970C51812DC3A010C7D01B50E0D17DC79C8",
        "stake": "100000000000000000"
    }
]
```

PR #1136, opened 2024-10-01
feat(telemetry): Add url data to operator updates on telemetry service

JulianVentura:
### Description

This PR adds new data to be updated for each operator on the telemetry service, which includes:
* `eth_rpc_url`
* `eth_rpc_url_fallback`
* `eth_ws_url`
* `eth_ws_url_fallback`
* `version` (this one was already present)

The implementation is also updated from the operator's side.

### Testing

In order to test, the following steps were taken:

1. Start anvil and the aggregator with:
 - `make aggregator_start`
2. Build and register operator with:
 - `make build_operator`
 - `make operator_full_registration CONFIG_FILE=config-files/config-operator-1.yaml`
3. Start telemetry service with:
 - `make telemetry_clean_db`
 - `make telemetry_run_db`
 - `make telemetry_start`
 4. Go to `http://localhost:4001/api/operators` and you should see the operator with its data fields. Some of them, like the version and RPC related should be set as null, since the operator is not up yet.
 5. Start the operator with `./operator/build/aligned-operator start --config ./config-files/config-operator-1.yaml`
 6. Go to `http://localhost:4001/api/operators` and you should see the operator, now with all fields with a certain value, including version and RPC related.

It is also highly advisable to test previous functionality as it's described on [this already merged PR](https://github.com/yetanotherco/aligned_layer/pull/1077)


operator/cmd/actions/utils.go
```diff
@@ -0,0 +1,17 @@
...
+  "regexp" 
+  "errors"
+)
+
+func baseUrlOnly(url string) (string, error)  {
+  // Removes the protocol and api key part from any url formated like so:
+  // "<protocol>://<base_url>/<api_key>"
+  regex := regexp.MustCompile(`^[a-z]+://([^/]+)`)
+  match := regex.FindStringSubmatch(url)
+  if len(match) > 1 {
```
JuArce: Would be better to invert the condition, so the function return is the normal behavior and the if return is an exceptional case

operator/cmd/actions/start.go
```diff
@@ -51,11 +40,32 @@ func operatorMain(ctx *cli.Context) error {
...
+    return err 
+  }
+  ethWsUrl, err := baseUrlOnly(operatorConfig.BaseConfig.EthWsUrl)
+  if err != nil {
+    return err 
+  }
+  ethWsUrlFallback, err := baseUrlOnly(operatorConfig.BaseConfig.EthWsUrlFallback)
+  if err != nil {
+    return err 
+  }
```
avilagaston9: I would suggest fixing this indentation

operator/cmd/actions/start.go
```diff
@@ -51,11 +40,32 @@ func operatorMain(ctx *cli.Context) error {
...
 
 	body := map[string]interface{}{
 		"version":   ctx.App.Version,
 		"signature": signature,
+    "eth_rpc_url": ethRpcUrl,
+    "eth_rpc_url_fallback": ethRpcUrlFallback,
+    "eth_ws_url": ethWsUrl,
+    "eth_ws_url_fallback": ethWsUrlFallback,
```
avilagaston9: Same here

JulianVentura: Done

operator/cmd/actions/start.go
JulianVentura: Done

telemetry_api/lib/telemetry_api/operators.ex
```diff
@@ -98,46 +98,23 @@ defmodule TelemetryApi.Operators do
...
 
-      iex> update_operator_version(%{field: bad_value})
+      iex> update_operator(%{field: bad_value})
       {:error, string}
 
   """
-  def update_operator_version(%{"version" => version, "signature" => signature}) do
-    with {:ok, address} <- SignatureVerifier.recover_address(version, signature) do
+  def update_operator(attrs) do
+    with {:ok, address} <- SignatureVerifier.recover_address(attrs["version"], attrs["signature"]) do
```
avilagaston9: I guess we should keep the old version of this. As suggested [here](https://github.com/yetanotherco/aligned_layer/pull/1000#discussion_r1784944450).

telemetry_api/lib/telemetry_api_web/controllers/operator_controller.ex
```diff
@@ -2,32 +2,37 @@ defmodule TelemetryApiWeb.OperatorController do
...
     operators = Operators.list_operators()
     render(conn, :index, operators: operators)
   end
 
-  def create(conn, operator_params) do
-    case Operators.update_operator_version(operator_params) do
-      {:ok, %Operator{} = operator} ->
-        conn
+  def create(conn, params) do
+    params = Map.take(params, @create_params)
```
avilagaston9: Same as above

operator/cmd/actions/utils_test.go
```diff
@@ -0,0 +1,46 @@
...
+func TestBaseUrlOnlyHappyPath(t *testing.T) {
+  // Format "<protocol>://<base_url>/<api_key>"
+
+  urls := [...][2]string{
+    {"http://localhost:8545/asdfoij2a7831has89%342jddav98j2748", "localhost:8545"},
+    {"ws://test.com/23r2f98hkjva0udhvi1j%342jddav98j2748", "test.com"},
+    {"http://localhost:8545", "localhost:8545"},
+  }
+
+  for _, pair := range urls {
```
Oppen: There seems to be inconsistent indentation here.

telemetry_api/lib/telemetry_api_web/controllers/operator_controller.ex
```diff
@@ -2,32 +2,37 @@ defmodule TelemetryApiWeb.OperatorController do
...
-    |> send_resp(:bad_request, Jason.encode!(%{error: message}))
-  end
+  @create_params [
+    "version",
+    "signature",
+    "eth_rpc_url",
+    "eth_rpc_url_fallback",
+    "eth_ws_url",
+    "eth_ws_url_fallback"
+  ]
```
avilagaston9: There should be a different way to handle this, but maybe it's something for a different PR

JulianVentura: Done

telemetry_api/lib/telemetry_api_web/controllers/operator_controller.ex
JulianVentura: Done

telemetry_api/lib/telemetry_api/operators.ex
JulianVentura: I made some changes, have a look and let me know if you think I should change something else!

operator/cmd/actions/utils_test.go
JulianVentura: I think it's ok but github renders it weird for some reason.

JulianVentura: It should be correctly indented now

telemetry_api/lib/telemetry_api_web/controllers/operator_controller.ex
```diff
@@ -11,8 +12,8 @@ defmodule TelemetryApiWeb.OperatorController do
     render(conn, :index, operators: operators)
   end
 
-  def create_or_update(conn, operator_params) do
-    with {:ok, %Operator{} = operator} <- Operators.update_operator_version(operator_params) do
+  def create_or_update(conn, %{"version" => version, "signature" => signature} = attrs) do
+    with {:ok, %Operator{} = operator} <- Operators.update_operator(version, signature, attrs) do
```
avilagaston9, line 15: Maybe we can send the `attrs` directly here:
```Elixir
Operators.update_operator(attrs)
```

JulianVentura: We could, but since both `version` and `signature` are required arguments I like it better to have them explicitly as parameters. I know we could do something like:
`def update_operator(%{"version" => version, "signature" => signature} = attrs)` 
 but I prefer it the other way.

operator/cmd/actions/utils_test.go
Oppen: It still looks the same. Did you run `go fmt` on this directory?
What it seems to have happened is that some lines use spaces and other lines use tabs. In Go, it's customary to use tabs, and that's what `go fmt` enforces. Some editors default to four spaces.

Oppen: In this case it seems it used two spaces.

MarcosNicolau (commented):
Tested on my machine, and it worked fine. One thing, in the diff I see you changed the `eigenlayer-middleware`, is that on purpose?

JulianVentura:
> Tested on my machine, and it worked fine. One thing, in the diff I see you changed the `eigenlayer-middleware`, is that on purpose?

No, I haven't touched anything related to that. There was a merge from staging branch, so maybe it was an auto merge error? `eigenlayer-middleware` is a submodule, I checked and the staging branch points to the same commit as this one, so Idk what caused that diff.

JulianVentura:
> > Tested on my machine, and it worked fine. One thing, in the diff I see you changed the `eigenlayer-middleware`, is that on purpose?
> 
> No, I haven't touched anything related to that. There was a merge from staging branch, so maybe it was an auto merge error? `eigenlayer-middleware` is a submodule, I checked and the staging branch points to the same commit as this one, so Idk what caused that diff.

My bad. The submodule commit was in fact different from the one in staging branch. Good catch!

MarcosNicolau (approved):

operator/cmd/actions/utils.go
```diff
@@ -0,0 +1,17 @@
...
+func baseUrlOnly(url string) (string, error) {
+	// Removes the protocol and api key part from any url formated like so:
+	// "<protocol>://<base_url>/<api_key>"
+	regex := regexp.MustCompile(`^[a-z]+://([^/]+)`)
+	match := regex.FindStringSubmatch(url)
+	if len(match) > 1 {
+		return match[1], nil
+	}
+	return "", errors.New("Url did not match the expected format <protocol>://<base_url>/<api_key>")
+}
```
entropidelic: I think this would be better placed in the operator root folder instead of in the actions one

entropidelic:
~~everything worked correctly, but I am thinking that it does not make much sense to register the http and websocket providers URL separately. Since the protocol and the API key are stripped from the URL, they will always be the same, or am I missing something?~~
nvm, talked with @JuArce and the current behaviour is correct

telemetry_api/lib/telemetry_api/operators.ex
```diff
@@ -142,24 +137,6 @@ defmodule TelemetryApi.Operators do
...
-
-      iex> update_operator(operator, %{field: bad_value})
-      {:error, %Ecto.Changeset{}}
-
-  """
-  def update_operator(%Operator{} = operator, attrs) do
-    operator
-    |> Operator.changeset(attrs)
-    |> Repo.update()
-  end
```
avilagaston9: I'm currently using this one here: https://github.com/yetanotherco/aligned_layer/pull/1186/files#diff-84f700b2608821ebb00e5350ef544a8c4e235cee1e465ba75f071beb9d7101b1R50-R54

We may have conflicts.

telemetry_api/lib/telemetry_api/operators.ex
```diff
@@ -116,21 +116,16 @@ defmodule TelemetryApi.Operators do
...
-      iex> update_operator_version(%{field: value})
+      iex> update_operator(%{field: value})
       {:ok, %Ecto.Changeset{}}
 
       iex> update_operator_version(%{field: bad_value})
       {:error, "Some status", "Some message"}
 
   """
-  def update_operator_version(%{"version" => version, "signature" => signature}) do
+  def update_operator(version, signature, changes) do
```
avilagaston9, line 126: Maybe you can update this one to use the deleted one

operator/cmd/actions/utils.go
avilagaston9: Done!

operator/pkg/utils.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"errors"
+	"regexp"
+)
+
+func BaseUrlOnly(url string) (string, error) {
+	// Removes the protocol and api key part from any url formated like so:
+	// "<protocol>://<base_url>/<api_key>"
+	regex := regexp.MustCompile(`^[a-z]+://([^/]+)`)
+	match := regex.FindStringSubmatch(url)
```
uri-99: fomat can't have more `/` ?
i:e: `https://myservice.com/<holesky-sepolia-mainnet>/ApiKey`
I think if a service has an extra `/` for any reason it may break the regex matching

operator/pkg/utils_test.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+func TestBaseUrlOnlyFailureCases(t *testing.T) {
+
+	urls := [...]string{
+		"localhost:8545/asdfoij2a7831has89%342jddav98j2748",
+		"this-is-all-wrong",
+	}
+
```
uri-99, line 39: we could add a failure case as mentioned above, using more `/`

uri-99 (approved):

uri-99 (changes requested):
Meant to request changes because of my previous comments. Everything else works

operator/pkg/utils_test.go
avilagaston9: Done!

operator/pkg/utils.go
avilagaston9: I have already handled it!

uri-99 (approved):

operator/cmd/actions/start.go
```diff
@@ -26,19 +27,7 @@ var StartCommand = &cli.Command{
...
-	if err != nil {
-		return err
-	}
-
-	operator, err := operator.NewOperatorFromConfig(*operatorConfig)
-	if err != nil {
-		return err
-	}
-
+func updateTelemetryService(operator *operator.Operator, ctx *cli.Context, operatorConfig *config.OperatorConfig) error {
```
JuArce:
```suggestion
func sendTelemetryData(operator *operator.Operator, ctx *cli.Context, operatorConfig *config.OperatorConfig) error {
```

JuArce: This could be an operator method instead of receive an operator

operator/cmd/actions/start.go
```diff
@@ -68,15 +78,35 @@ func operatorMain(ctx *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	operator, err := operator.NewOperatorFromConfig(*operatorConfig)
+	if err != nil {
+		return err
+	}
+
+	err = updateTelemetryService(operator, ctx, operatorConfig)
```
JuArce:
```suggestion
	err = operator.sendTelemetryData(ctx, operatorConfig)
```

JuArce (changes requested):

operator/cmd/actions/start.go
entropidelic: +1

operator/cmd/actions/start.go
entropidelic: I think even the `operatorConfig` argument can be omitted if we define `sendTelemetryData` as a method of Operator

JuArce (approved):

PR #1142, opened 2024-10-01
feat: add fund_payment_service and get_balance_in_payment_sevice on sdk

### LINKED ISSUE #1101: feat: add pay_batcher() on SDK
There is a useful function called `pay_batcher` ,  it is already developed in the ZKQuiz Example. It would be useful to have this function integrated in the SDK. It should also take Network as parameter

uri-99:
# This PR
This PR adds `fund_payment_service` and `get_balance_in_payment_sevice` functionalities to the SDK.

It also migrates the CLI functionalities so they use these new functions

## To Test

0) 
```
```

1) run anvil

2)
To query the current balance of the wallet we will be using, you can run:
```
cast call 0x7969c5eD335650692Bc04293B07F5BF2e7A673C0 "user_balances(address)(uint256)" f39fd6e51aad88f6f4ce6ab8827279cfffb92266
```
it should be 0

3) 
```
aligned deposit-to-batcher --amount 0.1ether --keystore_path config-files/devnet/keys/anvil.ecdsa.key.json
```

4)
```
aligned get-user-balance --user_addr 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
```
Should give the same value as running
```
cast call 0x7969c5eD335650692Bc04293B07F5BF2e7A673C0 "user_balances(address)(uint256)" f39fd6e51aad88f6f4ce6ab8827279cfffb92266
```


JuArce (approved):

MarcosNicolau (approved):
Code looks good. Everything went fine on my tests 👍

JuArce (approved):

PR #1153, opened 2024-10-02
refactor: zkquiz setup and testing locally/devnet

MarcosNicolau:
**Changes**
This pr refactors zkquiz example so that testing it locally becomes easier.

**Testing**
To test this, you can follow [this guide](https://github.com/yetanotherco/aligned_layer/blob/refactor-zk-quiz-testing/examples/zkquiz/README.md#testing-locally).


examples/zkquiz/quiz/script/src/main.rs
```diff
@@ -34,6 +30,12 @@ struct Args {
         default_value = "https://ethereum-holesky-rpc.publicnode.com"
     )]
     rpc_url: String,
+    #[arg(short, long, default_value = "wss://batcher.alignedlayer.com")]
+    batcher_url: String,
+    #[arg(short, long, default_value = "holesky")]
+    network: Network,
+    #[arg(short, long, default_value = "17000")]
+    chain_id: u64,
```
uri-99: No, chainID must be retreived from RPC

examples/zkquiz/quiz/script/src/main.rs
```diff
@@ -50,7 +52,7 @@ async fn main() {
 
     let wallet = LocalWallet::decrypt_keystore(args.keystore_path, &keystore_password)
         .expect("Failed to decrypt keystore")
-        .with_chain_id(17000u64);
+        .with_chain_id(args.chain_id);
```
uri-99: ++

examples/zkquiz/quiz/script/src/main.rs
```diff
@@ -196,8 +198,9 @@ fn read_answer() -> char {
 async fn deposit_to_batcher(
     from: Address,
     signer: Arc<SignerMiddleware<Provider<Http>, LocalWallet>>,
+    network: Network,
 ) -> anyhow::Result<()> {
-    let addr = get_payment_service_address(NETWORK);
+    let addr = get_payment_service_address(network);
```
uri-99: This function should not exist, deposit_to_aligned from the SDK should be used

uri-99 (changes requested):
Also, currently batches must be at least size 2. If you only send this proof, the batch will not be finalized by the batcher. If it is a local devnet example, you could send this proof twice to make the batch finalize, or clarify that the user must in parallel send another proof (such as `make batcher_send_risc0_burst`)

examples/zkquiz/quiz/script/src/main.rs
MarcosNicolau: Updated!

uri-99 (approved):

examples/zkquiz/README.md
```diff
@@ -52,3 +52,34 @@ This will:
...
+    CONTRACT_ADDRESS=<VERIFIER_CONTRACT_ADDRESS> make answer_quiz_local
+    ```
+
+> [!NOTE]
+> Make sure you send another proof alongside the ZKQuiz proof, as the batcher will not create a batch with only 1 proof.
+> 
+> For example you could do:
+> ```
+> make batcher_send_risc0_burst
+> ```
```
entropidelic: fortunately the PR that fixed the functionality of merkle trees with only one leaf was merged yesterday, so this note shouldn't be necessary. Make sure to test it works correctly without the need to send more proofs

entropidelic:
left a small comment on the readme, but aside from that it looks good!

examples/zkquiz/README.md
uri-99: removed comment

uri-99 (changes requested):
sending only 1 proof doesn't work

uri-99 (approved):
fixed

entropidelic (approved):

PR #1162, opened 2024-10-03
feat(testing): Task Sender / Stress Tester.

### LINKED ISSUE #1015: create rust program with the SDK to mimic infinite proof submission scripts

### LINKED ISSUE #1113: Write a stress test for concurrent connections
We need to make sure we can handle a reasonable number of concurrent clients.
We also need to make sure we gracefully reject excess load to avoid DDoS attacks.

PatStiles:
Closes #1113 and Closes #1015 . 

Adds a task sender based on the previous make script for stress testing the network. It spawns infinite tokio tasks submitting groth16 proofs to aligned across n threads. 

## To Test
- Run a local testnet
- Start the task sender via
```
make batcher_send_infinite_burst
```

```
make batcher_send_infinite_stream
```

```
make batcher_connections NUM_SENDERS=<# connections>
```


scripts/test_files/gnark_groth16_bn254_infinite_script/pkg/generate_proof.go
```diff
@@ -26,7 +26,7 @@ func (circuit *InequalityCircuit) Define(api frontend.API) error {
 }
 
 func GenerateIneqProof(x int) {
-	outputDir := "scripts/test_files/gnark_groth16_bn254_infinite_script/infinite_proofs/"
+	outputDir := "../../scripts/test_files/gnark_groth16_bn254_infinite_script/infinite_proofs/"
```
Oppen, line 29: Maybe this should just be a parameter.

batcher/aligned-task-sender/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+                    e.to_string()
+                );
+                return Ok(());
+            }
+        }
+    };
+
+    let chain_id = get_chain_id(&args.eth_rpc_url).await.map_err(|e| {
+                    SubmitError::GenericError(format!(
+                        "Failed to retrieve chain id, verify `eth_rpc_url` is cor/rect or local testnet is running on \"http://localhost:8545\": {}", e
```
Oppen:
```suggestion
                        "Failed to retrieve chain id, verify `eth_rpc_url` is correct or local testnet is running on \"http://localhost:8545\": {}", e
```

batcher/aligned-task-sender/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+                    std::thread::sleep(Duration::from_secs(burst_time as u64));
+                }
+            })
+        })
+        .collect::<Vec<_>>();
+
+    for t in threads {
+        if let Err(e) = t.join() {
+            error!("Thread panicked: {:?}", e);
+            return Ok(());
```
Oppen: Suggestion: don't return, just report the panic and wait for the other tasks.

batcher/aligned-task-sender/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+                        .choose_multiple(&mut thread_rng(), burst_size)
+                        .cloned()
+                        .collect();
+                    info!(
+                        "Sending {:?} Proofs to {:?} Aligned Batcher",
+                        burst_size, network
+                    );
+                    let nonce = nonce.fetch_add(burst_size as u64, Ordering::Relaxed);
+                    let batcher_url = batcher_url.clone();
+                    let wallet = wallet.clone();
```
Oppen, line 242: I wonder if each thread could use a unique wallet. Using the same one might force serialization in the batcher if it represents the same operator.

PatStiles: For this we would need to programmatically create and fund accounts on anvil for each wallet. This is possible using via two methods:
- The task sender initiates anvil with `num_senders` accounts pre-funded and extract the private keys and addresses.
- Each thread generates an address and private key pair then calls `anvil_setBalance` https://book.getfoundry.sh/reference/anvil/ .

MarcosNicolau:
Closing this one in favour of #1189

PR #1184, opened 2024-10-04
fix: operator commands

### LINKED ISSUE #1182: fix: make operator_register_and_start on Linux
This fix needs:
`OS := /bin/bash` on Makefile for linux
`LD_LIBRARY_PATH` flag when running the operator

MarcosNicolau:
**Changes**
Fixes operator commands on linux.

**Test**
To test these commands now work, run the following commands on a linux machine:

- `make aggregator_start`
- `make build_operator`
- `make operator_register_and_start`
- `make batcher_start_local`
- `make batcher_send_burst_groth16`

You should be able to run this without any issues.

Closes #1182

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 .PHONY: help tests
 
+SHELL := /bin/bash
 OS := $(shell uname -s)
 
 CONFIG_FILE?=config-files/config.yaml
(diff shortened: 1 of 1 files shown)


uri-99 (approved):

Makefile
```diff
@@ -1,5 +1,6 @@
 .PHONY: help tests
 
+SHELL := /bin/bash
```
Oppen, line 3: Be careful here. This is being set globally for all OSes and Bash in Mac is __ancient__ by default. Some scripts may not work as intended.

Oppen: The proper fix, in any case, is to not use `#!/bin/sh` in our scripts but use `#!/bin/bash` instead. Or, avoid bashisms where not needed.

MarcosNicolau: Yeah, probably the best solution would be to refactor our scripts avoiding bash syntax. But I don't see that as a viable solution given the number of scripts we currently have. I' ll mark this conversation resolved as setting #!/bin/bash would have the same result.

Oppen (approved):
As discussed offline, the issue with using `#!/bin/bash` would be the same for MacOS as with using `SHELL := /bin/bash`, so we'll keep the latter.

PR #1189, opened 2024-10-07
feat(testing): Task Sender / Stress Tester

### LINKED ISSUE #1015: create rust program with the SDK to mimic infinite proof submission scripts

### LINKED ISSUE #1113: Write a stress test for concurrent connections

PatStiles:
> [!NOTE]
> This PR is pointed to 1314 branch because it needs the fix included in said branch.
> [PR of that branch is here](https://github.com/yetanotherco/aligned_layer/pull/1325)

## Changes
Adds a cli program to:
- Generate, fund wallets, and deposit to aligned payment service
- Send proofs with a burst of `n` infinitely from the generated wallets.
- Open `n` connection and hang infinitely.

### devnet
- run devnet, with a new function that start anvil with many prefunded accounts:
```
make anvil_start_with_block_time_with_more_prefunded
```

- Run Batcher
```
make batcher_start_local
```

- Use the prefunded accounts to fund the BatcherPaymentService
Note:
This will fund from every private key stored in the file `batcher/aligned-task-sender/wallets/devnet`
You can take any amount of wallets you want from the file `batcher/aligned-task-sender/wallets/devnet-backup`
```
make task_sender_fund_wallets_devnet
```

- Generate groth16 proofs. These will be taken at random when sending them later on, so merkle_roots are not easily collided.
This will create a `proofs` folder in `scripts/test_files/task_sender/proofs` by default.
```
NUMBER_OF_PROOFS=15 make task_sender_generate_groth16_proofs 
```

- Send infinite proofs
Before running this command, you need to make sure you have first generated wallets and proofs. This command reads the default paths mentioned before.
  - `BURST_SIZE` is how many proofs will each sender send.
  - `BURST_TIME_SECS` is how much time will each sender wait between each send_burst.
```
BURST_SIZE=2 BURST_TIME_SECS=5 make task_sender_infinite_proofs_devnet
```

- Test connections
This will start `NUM_SENDERS` connections and hang them infinitely without sending any proof
```
NUM_SENDERS=123 make task_sender_test_connections_devnet
```

### Testnet

- Generate and fund wallets
This will generate wallets on holesky, send them `AMOUNT_TO_DEPOSIT` funds from `FUNDING_WALLET_PRIVATE_KEY`, then send `AMOUNT_TO_DEPOSIT_TO_ALIGNED` funds from the newly created wallet to the BatcherPaymentService, to allow submission of proofs from this wallet.
```
FUNDING_WALLET_PRIVATE_KEY=xxx NUM_WALLETS=10 AMOUNT_TO_DEPOSIT=0.01 AMOUNT_TO_DEPOSIT_TO_ALIGNED=0.0099 make task_sender_generate_and_fund_wallets_holesky_stage
```

- Generate groth16 proofs. These will be taken at random when sending them later on, so merkle_roots are not easily collided.
This will create a `proofs` folder in `scripts/test_files/task_sender/proofs` by default.
```
NUMBER_OF_PROOFS=15 make task_sender_generate_groth16_proofs 
```

- Send infinite proofs
Before running this command, you need to make sure you have first generated wallets and proofs. This command reads the default paths mentioned before.
  - `BURST_SIZE` is how many proofs will each sender send.
  - `BURST_TIME_SECS` is how much time will each sender wait between each send_burst.
```
BURST_SIZE=2 BURST_TIME_SECS=2 make task_sender_infinite_proofs_holesky_stage
```

- Test connections
This will start `NUM_SENDERS` connections and hang them infinitely without sending any proof
```
NUM_SENDERS=123 make task_sender_test_connections_holesky_stage
```

Closes #1113 and Closes #1015.  Reopens #1162 due to merge conflicts.


MarcosNicolau (commented):
It works. But it doesn't support many open connections. I think the idea is to open millions of concurrent requests. For that, we just need to open the socket connections, it is not necessary to send proofs which might be the bottleneck.

batcher/aligned-task-sender/src/structs.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+#[derive(Parser, Clone, Debug, ValueEnum)]
+pub enum ProofType {
+    Groth16,
+}
+
+#[derive(Parser, Debug)]
+#[command(version, about, long_about = None)]
+pub struct CleanProofsArgs {}
```
MarcosNicolau: I have to remove this one

MauroToscano (changes requested):
Wait for SDK

uri-99:
Also missing build sender in CI

Makefile
```diff
@@ -441,6 +450,82 @@ batcher_send_burst_groth16: batcher/target/release/aligned
...
+task_sender_fund_wallets_devnet:
+	@cd batcher/aligned-task-sender && \
+	cargo run --release -- generate-and-fund-wallets \
+	--eth-rpc-url http://localhost:8545 \
+	--network devnet \
+	--amount-to-deposit 1 \
+	--amount-to-deposit-to-aligned 0.9999 \
+	--private-keys-filepath $(CURDIR)/batcher/aligned-task-sender/wallets/devnet
+
+task_sender_infinite_proofs_devnet:
```
JulianVentura: Can we call this `task_sender_send_infinite_proofs_devnet`?

Makefile
```diff
@@ -441,6 +450,82 @@ batcher_send_burst_groth16: batcher/target/release/aligned
...
+	cargo run --release -- generate-and-fund-wallets \
+	--eth-rpc-url https://ethereum-holesky-rpc.publicnode.com \
+	--network holesky-stage \
+	--funding-wallet-private-key $(FUNDING_WALLET_PRIVATE_KEY) \
+	--number-wallets $(NUM_WALLETS) \
+	--amount-to-deposit $(AMOUNT_TO_DEPOSIT) \
+	--amount-to-deposit-to-aligned $(AMOUNT_TO_DEPOSIT_TO_ALIGNED) \
+	--private-keys-filepath $(CURDIR)/batcher/aligned-task-sender/wallets/holesky-stage
+
+task_sender_infinite_proofs_holesky_stage:
```
JulianVentura: Same

batcher/aligned-task-sender/README.md
```diff
@@ -0,0 +1,95 @@
...
+Then run the following make target, with `NUM_WALLETS` being the amount of wallets you want to deposit funds to aligned payment service, up to 1000.
+```bash
+NUM_WALLETS=<N> make task_sender_generate_and_fund_wallets_devnet
+```
+
+## SendInfiniteProofs
+
+This command sends `BURST_SIZE` proofs from each private key in `PATH_TO_PRIVATE_KEYS_FILE` every `BURST_TIME_SECS` seconds.
+
+To vary the amount of senders, it is recommended to have a backup with all private keys, and add/remove keys from the file being used.
```
JulianVentura, line 57: nit: Instead of modifying the keys file in order to change the amount of senders, we could do so with a new parameter. If the specified amount is larger than the senders found in the file, we log a warn and continue with all senders. If not, we sample the file at random.

batcher/aligned-task-sender/src/commands.rs
```diff
@@ -0,0 +1,453 @@
...
+            // Load the wallet
+            let private_key_str = line.unwrap();
+            let wallet = Wallet::from_str(private_key_str.trim())
+                .expect("Invalid private key")
+                .with_chain_id(chain_id.as_u64());
+
+            // Send funds to aligned from the wallet
+            let funded_wallet_signer =
+                SignerMiddleware::new(eth_rpc_provider.clone(), wallet.clone());
+            tokio::time::sleep(Duration::from_millis(50)).await; // To avoid overloading the RPC
```
JulianVentura, line 99: nit: This may be a bottleneck, we could try to improve it in the future if it takes too long:
* We could send 3 transactions and then wait for ~1 min
* We could use different RPC nodes

batcher/aligned-task-sender/src/commands.rs
```diff
@@ -0,0 +1,453 @@
...
+                {
+                    error!("Could not deposit to aligned, err: {:?}", err);
+                    return;
+                }
+                info!("Successfully deposited to aligned for wallet {}", i);
+            }));
+        }
+
+        // Wait for last tx to be sent
+        if let Some(handle) = last_handle {
```
JulianVentura: nit: This works, but we should await all handles, not only the last one

JulianVentura (commented):
Code looks good to me and it worked on devnet, I'll give you the approve after testing on staging.
Left you some nit comments, you don't have to address them

JulianVentura (approved):
Tested on staging, work as expected!
Note: I noticed that while sending proofs, after each iteration, the ws connections between the stress tester and the batcher are shown as broken. I think that this is a minor detail since it works fine.

batcher/aligned-task-sender/README.md
MauroToscano: This is good enough for now

batcher/aligned-task-sender/src/commands.rs
MauroToscano: Unless it actually becomes a bottleneck, it's not a bottleneck, but if we try something and it stops here we should check

uri-99: this is not really a problem, funding 1000 wallets takes less than a minute. And it is necesarry the small await for not overloading the RPC (devnet, so we cant have many, since it is local).

Also, the suggested fix does 3 transactions every 1 minute. this one does more than 1000 transactions per minute

Makefile
```diff
@@ -132,6 +132,10 @@ anvil_start_with_block_time:
 	@echo "Starting Anvil..."
 	anvil --load-state contracts/scripts/anvil/state/alignedlayer-deployed-anvil-state.json --block-time 7
 
+anvil_start_with_block_time_with_more_prefunded:
+	@echo "Starting Anvil..."
+	anvil --load-state contracts/scripts/anvil/state/alignedlayer-deployed-anvil-state.json --block-time 7 -a 2000
```
Oppen: This looks like what we really want is to parameterize how much funding we want, while giving a sane default. Otherwise, we may later need a target `anvil_start_with_block_time_with_even_more_prefunded` and it'd be a never-ending story.

Oppen:
> Wait for SDK

@MauroToscano is the needed work in the SDK done? Maybe we can remove the change request?

Makefile
uri-99: more prefunded is more prefunded accounts! not quantity of funds, i will add the `accounts` for more clarity

batcher/aligned-task-sender/src/commands.rs
uri-99: done

Oppen (approved):

MauroToscano (approved):

PR #1196, opened 2024-10-07
fix: add panic catch on operator calling FFI

### LINKED ISSUE #1174: Add try catch on verifiers calls

JulianVentura:
### Motivation

We have some calls to external Rust functions on operator Golang code. If any of those functions panics, that error is propagated to the golang application and it fails completly. We want to catch those errors so this doesn't happen.

### Changes

* Add golang panic handling with `recover()` on `VerifySp1Proof`, `VerifyRiscZeroReceipt` and `VerifyMerkleTreeBatch`.
* Add a function wrapper with a `catch_unwind` from Rust side for `verify_sp1_proof_ffi`, `verify_risc_zero_receipt_ffi` and `verify_merkle_tree_batch_ffi`.
* Modify `verify_sp1_proof_ffi`, `verify_risc_zero_receipt_ffi` and `verify_merkle_tree_batch_ffi` to return `i32` instead of `bool` to inform about errors.

### How to test

In order to test properly, we first have to take the `staging` branch as baseline and add a `panic!()` in the following FFI Rust functions (one at a time):

* `verify_merkle_tree_batch_ffi` on `merkle_tree/lib/src/lib.rs`
* `verify_sp1_proof_ffi` on `sp1/lib/src/lib.rs`
* `verify_risc_zero_receipt_ffi`  on `risc_zero/lib/src/lib.rs`

This way, we will make the golang operator crash on a call to any of these FFIs, stoping execution immediately.

_Note: Explicit instructions on how to run the operator are provided in the last section of this PR. The operator must be rebuilt after modifying the FFI functions_

1. To test the call to `verify_merkle_tree_batch_ffi` we have to run the hole system and start sending transactions with:
      - `make batcher_send_infinite_sp1`
2. To test the call to `verify_sp1_proof_ffi` we have to run the hole system and start sending transactions with:
      - `make batcher_send_infinite_sp1`
3. To test the call to `verify_risc_zero_receipt_ffi` we have to run the hole system and start sending transactions with: 
      - `make batcher_send_risc0_burst BURST_SIZE=5`

Once you have tested the crashing behaviour, you can now switch to `fix/operator-ffi-panic-catch` branch and repeat the process, but this time the panics will have to be added to the `inner` versions of the provided FFI functions, which are:

* `inner_verify_merkle_tree_batch_ffi` on `merkle_tree/lib/src/lib.rs`
* `inner_verify_sp1_proof_ffi` on `sp1/lib/src/lib.rs`
* `inner_verify_risc_zero_receipt_ffi`  on `risc_zero/lib/src/lib.rs`

These should also be tested individually.

After executing you should see some errors being logged in the operator, providing information about the rust code panics. If the operator hasn't crashed, then everything is working properly.

### How to run the operator

1. Execute anvil: 
2. Launch aggregator: 
  - `make aggregator_start`
3. Launch batcher:
  - `make batcher_start_local`
5. Launch operator: 
  - `make build_operator`
  - `make operator_register_and_start CONFIG_FILE=config-files/config-operator-1.yaml`

Closes #1174


Makefile
```diff
@@ -298,6 +298,18 @@ batcher_send_risc0_burst:
 		--rpc_url $(RPC_URL) \
 		--network $(NETWORK)
 
+batcher_send_risc0_big_burst:
```
MarcosNicolau: You can use the current target `batcher_send_risc0_burst` and provide a `BURST_SIZE=100`

MarcosNicolau (changes requested):
I tested on macos and go still panics. Have you rebuilt the ffis after every change? 

I might have tested it wrongly, but would you please check If it works on your machine if it does, I'll make sure to try again.

JuArce:
Would be nice a review from @Oppen

PatStiles:
This sucessfully caught panics for the merkle tree verification and sp1 verification but failed when I tested with risc0.

Makefile
JulianVentura: Done!

JulianVentura:
> I tested on macos and go still panics. Have you rebuilt the ffis after every change?
> 
> I might have tested it wrongly, but would you please check If it works on your machine if it does, I'll make sure to try again.

Could you check again, please?

MarcosNicolau (approved):
It is working for me now! 🚀

operator/merkle_tree_old/lib/libmerkle_tree.a
Oppen, line 1: This seems to have been committed by accident.

operator/merkle_tree_old/lib/libmerkle_tree.dylib
Oppen, line 1: This as well.

Oppen (approved):
Works for me and the code looks good. Just remove the leftover binaries.

PatStiles (approved):
Caught all panics on a local testnet.

uri-99 (approved):

PR #1198, opened 2024-10-08
fix(explorer): fetch_batch_data_pointer in explorer reads the entire response body without any limitation

### LINKED ISSUE #1016: fix(explorer): `fetch_batch_data_pointer` in `explorer` reads the entire response body without any limitation.
## Overview:
The [`fetch_batch_data_pointer`](https://github.com/yetanotherco/aligned_layer/blob/staging/explorer/lib/explorer_web/live/utils.ex#L182) function in the [`explorer`](explorer/lib/explorer_web/live/utils.ex) is vulnerable to an OOM attack.

The function reads the entire response body without any limitation, which can lead to an OOM attack.

## Detailed Behavior
During the processing of a new batch log, the explorer uses a mutex to prevent multiple indexation of the same batch. It periodically checks for new batch logs and processes them. If a batcher takes more time than the interval between two checks, the explorer will lose the other batch logs since it will provoke an OOM and crash the explorer.

## Mitigation
Set a timeout for the request to prevent an OOM attack.

avilagaston9:
**Motivation**

The function `fetch_batch_data_pointer` in the explorer reads the entire response body without any
limitations, which can lead to an OOM attack.

**Description**

Uses a `max_batch_size` limit when reading request bodies.

**How to Test**

See #1202 

Closes #1016

### DIFF
--- a/explorer/lib/explorer_web/live/utils.ex
+++ b/explorer/lib/explorer_web/live/utils.ex
@@ -172,7 +172,7 @@ defmodule Utils do
     |> Enum.reverse()
   end
 
-  def calculate_proof_hashes({:ok, deserialized_batch}) do
+  def calculate_proof_hashes(deserialized_batch) do
     deserialized_batch
     |> Enum.map(fn s3_object ->
       :crypto.hash(:sha3_256, s3_object["proof"])
--- a/explorer/lib/explorer_web/live/utils.ex
+++ b/explorer/lib/explorer_web/live/utils.ex
@@ -183,8 +183,8 @@ defmodule Utils do
   end
 
   defp stream_handler({:headers, headers}, acc) do
-    {_, batch_size} = List.keyfind(headers, "content-length", 0, {nil, 0})
-    check_batch_size(batch_size, acc)
+    {_, batch_size} = List.keyfind(headers, "content-length", 0, {nil, "0"})
+    check_batch_size(String.to_integer(batch_size), acc)
   end
 
   defp stream_handler({:status, 200}, acc), do: {:cont, acc}
--- a/explorer/lib/explorer_web/live/utils.ex
+++ b/explorer/lib/explorer_web/live/utils.ex
@@ -183,7 +183,7 @@ defmodule Utils do
   end
 
   defp stream_handler({:headers, headers}, acc) do
-    {_, batch_size} = List.keyfind(headers, "content-length", 0)
+    {_, batch_size} = List.keyfind(headers, "content-length", 0, {nil, 0})
     check_batch_size(batch_size, acc)
   end
 
--- a/explorer/lib/explorer_web/live/utils.ex
+++ b/explorer/lib/explorer_web/live/utils.ex
@@ -177,6 +177,7 @@ defmodule Utils do
 
   def calculate_proof_hashes({:error, reason}) do
     Logger.error("Error calculating proof hashes: #{inspect(reason)}")
+    # This ensures we avoid attempting to fetch the invalid data again.
     ["invalid"]
   end
 
--- a/explorer/lib/explorer_web/live/utils.ex
+++ b/explorer/lib/explorer_web/live/utils.ex
@@ -132,7 +132,7 @@ end
 # Backend utils
 defmodule Utils do
   require Logger
-  @max_batch_size String.to_integer(System.get_env("POOL_SIZE") || "268435456")
+  @max_batch_size String.to_integer(System.get_env("MAX_BATCH_SIZE") || "268435456")
 
   def string_to_bytes32(hex_string) do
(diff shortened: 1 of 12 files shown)


uri-99 (approved):

uri-99 (changes requested):
WIP

explorer/lib/explorer_web/live/utils.ex
```diff
@@ -132,6 +132,7 @@ end
 # Backend utils
 defmodule Utils do
   require Logger
+  @max_batch_size String.to_integer(System.get_env("MAX_BATCH_SIZE") || "268435456")
```
uri-99:
```suggestion
  @max_batch_size (case System.fetch_env("MAX_BATCH_SIZE") do
    {:ok, ""} -> 268_435_456 #empty env var
    {:ok, value} -> String.to_integer(value)
    _ -> 268_435_456 # error
  end)
```

explorer/lib/explorer_web/live/utils.ex
```diff
@@ -245,9 +262,17 @@ defmodule Utils do
...
+          batch_content = batch.data_pointer |> Utils.fetch_batch_data_pointer()
+          case batch_content do
+            {:ok, batch_content} ->
+              batch_content
+              |> Utils.calculate_proof_hashes()
+
+            {:error, reason} ->
+              Logger.error("Error fetching batch content: #{inspect(reason)}")
+              # Returning something ensures we avoid attempting to fetch the invalid data again.
+              ["invalid batch"]
```
uri-99: This string is converted to hex in frontend, better handle invalids in a cleaner way

uri-99 (changes requested):
It would be best to have another state for the batch.
For this a new ecto migrate is needed, adding an `is_valid` bool to the batches table.
Also the schemas should be updated, and the value should be set to false if the batch is not valid.

Also in the frontend, invalid proofs should be shown a red `invalid` status.

explorer/lib/explorer_web/live/utils.ex
```diff
@@ -167,21 +172,36 @@ defmodule Utils do
...
+    do: {:halt, {:error, {:http_error, status_code}}}
+
+  defp stream_handler({:data, chunk}, {_acc_body, size})
+       when size + byte_size(chunk) > @max_batch_size do
+    {:halt, {:error, {:http, :body_too_large}}}
+  end
+
+  defp stream_handler({:data, chunk}, {acc_body, size}) do
+    new_size = size + byte_size(chunk)
+    {:cont, {acc_body <> chunk, new_size}}
```
entropidelic: instead of separating the function in these two cases, I think it is a bit more understandable to condense these in one function and check the condition `size + byte_size(chunk) > @max_batch_size` inside it

entropidelic: also, for consistency I think it is better to rename `{acc_body, size}` to `{acc_body, acc_size}`

explorer/lib/explorer_web/live/utils.ex
```diff
@@ -167,21 +172,36 @@ defmodule Utils do
...
     deserialized_batch
     |> Enum.map(fn s3_object ->
       :crypto.hash(:sha3_256, s3_object["proof"])
     end)
   end
 
-  def calculate_proof_hashes({:error, reason}) do
-    Logger.error("Error calculating proof hashes: #{inspect(reason)}")
-    []
+  defp stream_handler({:headers, _headers}, acc), do: {:cont, acc}
```
entropidelic: we could add a check here to validate that the content lenght in the request header is not greater than the maximum batch size, like we have in the operator side, in order to avoid downloading unecessary data https://github.com/yetanotherco/aligned_layer/blob/477de1d5801b93507427eb7cb8e3504ac55dc617/operator/pkg/s3.go#L73

JuArce (commented):
Same comments as Marian

explorer/lib/explorer_web/live/utils.ex
avilagaston9: Done!

explorer/lib/explorer_web/live/utils.ex
avilagaston9: Nice! Added.

explorer/lib/explorer_web/live/utils.ex
avilagaston9: Changed to `[<<0>>]` to show `0x0` in the frontend for now. I'm solving this in a different PR in case we need to merge this one.

uri-99 (approved):

MarcosNicolau (approved):
Tested on macos.

entropidelic (approved):

PR #1211, opened 2024-10-10
fix: don't start operator if `last_processed_batch_filepath` isn't provided

### LINKED ISSUE #1209: Don't start operator if it hasn't provide `last_processed_batch_filepath` in config file
Currently there is no check if the `last_processed_batch_filepath` field has been provided in the config file. Since this is a new mandatory field, we need to make sure we do a fatal log to ensure all operators set this field.

MarcosNicolau:
**Changes**
If `last_processed_batch_filepath` is not provided on the operator config, then don't start the program. This is to make sure operator owners don't forget to pass this new field. Alternatively, we could create a default file but I am not sure about how that could go.

**Testing**
Remove the `last_processed_batch_filepath` from the operator config and you should see an err indicating the missing field.

Closes #1209


JuArce (approved):

PatStiles:
Works on my machine

PatStiles (approved):

uri-99 (changes requested):
It should also fail if the folder of the file is non-existent. This is because Operator will fail to create the file if the folder given does not exist.

MarcosNicolau:
> It should also fail if the folder of the file is non-existent. This is because Operator will fail to create the file if the folder given does not exist.

Updated!

uri-99 (changes requested):
new constraint doesnt allow creation of file. It should be able to create a file, but only on an existing directory.

uri-99 (approved):

PR #1227, opened 2024-10-11
feat(PoC): retry functions

entropidelic:
> [!NOTE]
> This PR is just to showcase a PoC of the standardized retry functions and is not intended to be merged

For both the Rust and Go codebases, a proof of concept has been made for the standardization of the `retry` functions used across the entire project.

The `retry` functions in each case receives some function to be retried. In our concrete case, these functions will be related to message sending between different entities such as `Batcher -> Ethereum`, for example.

The function can return two distinct type of errors: `Recoverable` and `NonRecoverable`. For the `Recoverable` case, `retry` will keep retrying until some of the stop condition is met. For the `NonRecoverable` case, the `retry` will not keep trying.

The parameters in both Rust and Go versions have been set to have the same behaviour.

## Test
You will see a dummy function that prints to console being called periodically with an exponential backoff spacing between calls. There is some randomness for simulating a `NonRecoverable` error case, for which the function will return immediately without retrying again

### Rust version
From the Aligned root folder,
```bash
cd rust_retry
```

**tokio_retry lib**
```bash
cd rust_tokio_retry
cargo run
```

**backon lib**
```bash
cd rust_retry_backon
cargo run
```

### Go version
From the Aligned root folder,

```bash
cd go_retry_poc
```

**cenkalti/backoff lib**
```shell
go run ./cenkalti/main.go
```

**avast/retry-go lib**
```shell
go run ./avast/main.go
```


go_retry_poc/main.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	// * ---------------------------------------------------------------------------------------- *
+	// *                         EXPONENTIAL BACKOFF CONFIGURATION                                *
+	// * ---------------------------------------------------------------------------------------- *
+	// For the exponential backoff formula `backoff(n) = a * b^n`, we set the following config:
+	// *    a = 2000ms
+	// * 	b = 2
+	// *    0 <= n <= 2
+	// There is no randomization factor
```
Oppen: We should probably use a single formula, with the constants for that formula, and then derive the parameters for the particular implementation. Something like:
```
Formula: a * b ^ n, with a = 1000ms, b = 2, 1 <= n <= 3
For Rust: n already starts at 1, so we simply pass the values.
For Go: n actually starts at 0 in the library, so we perform the first multiplication on the a parameter so `a * b * b^n-1 = a * b^n`, thus a = 2000ms, b = 2, 0 <= n <= 2.
```
With better phrasing and style than I used, of course. Visually, it should be like the original, but more explicit about the behavior.

Oppen:
As a side-note: we should probably report a bug or fork the Rust library. I've been thinking and its behavior is incorrect: if you state a parameter is the initial wait time for the first retry, you should honor that rather than multiply it on the first wait behind the users' back. Only after the first retry should it start adding time to the wait.

PR #1236, opened 2024-10-14
refactor: avoid lazy init for batcher metrics

IAvecilla:
Initially, we implemented a minimal solution to get the job done, but we ended up using `LazyStatic` without a clear purpose. We initialize all the metric variables anyway, as we need them from the start. These changes ensure that the metrics are available right from the beginning, and they are now incorporated as a new component of the batcher. Additionally, the port used to expose metrics is now part of the batcher's configuration.

### DIFF
--- a/batcher/aligned-batcher/src/main.rs
+++ b/batcher/aligned-batcher/src/main.rs
@@ -52,7 +52,7 @@ async fn main() -> Result<(), BatcherError> {
         }
     });
 
-    batcher.metrics.batcher_started.inc();
+    batcher.metrics.inc_batcher_restart();
 
     batcher.listen_connections(&addr).await?;
 
--- a/batcher/aligned-batcher/src/main.rs
+++ b/batcher/aligned-batcher/src/main.rs
@@ -5,9 +5,7 @@ use std::sync::Arc;
 use clap::Parser;
 use env_logger::Env;
 
-use aligned_batcher::metrics;
 use aligned_batcher::{types::errors::BatcherError, Batcher};
-use warp::Filter;
 
 /// Batcher main flow:
 /// There are two main tasks spawned: `listen_connections` and `listen_new_blocks`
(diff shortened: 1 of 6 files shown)


Oppen (approved):

Oppen (commented):
Will test later today. Looks good.

Oppen (approved):
Seems to work.

batcher/aligned-batcher/src/main.rs
```diff
@@ -63,7 +52,7 @@ async fn main() -> Result<(), BatcherError> {
         }
     });
 
-    metrics::batcher_started();
+    batcher.metrics.batcher_started.inc();
```
uri-99: This should call the function that has a short timer before increasing the value

uri-99 (changes requested):

PatStiles (approved):

batcher/aligned-batcher/src/metrics.rs
```diff
@@ -1,67 +1,74 @@
...
+        if let Err(e) = encoder.encode(&registry.gather(), &mut buffer) {
+            eprintln!("could not encode prometheus metrics: {}", e);
+        };
+        let res = match String::from_utf8(buffer.clone()) {
+            Ok(v) => v,
+            Err(e) => {
+                eprintln!("prometheus metrics could not be from_utf8'd: {}", e);
+                String::default()
+            }
+        };
```
entropidelic: very pedantic nit :), learned this one from Pablo
```suggestion
    let res = String::from_utf8(buffer.clone())
        .inspect_err(|e| eprintln!("prometheus metrics could not be from_utf8'd: {e}"))
        .unwrap_or_default();
```
```

entropidelic: also I think the error message is missing some word

IAvecilla: Nice! I also fixed the comment [78dfcbe](https://github.com/yetanotherco/aligned_layer/pull/1236/commits/78dfcbe5c08913963b71ed5171312450993bb1e7)

MarcosNicolau (approved):
Works on my machine!

batcher/aligned-batcher/src/main.rs
IAvecilla: Yeah, I added the sleep back but with 2 seconds, just to don't sleep for that long. We may need to change this in the future I don't think is the better solution, I created a new [issue](https://github.com/yetanotherco/aligned_layer/issues/1259) for future reference.

uri-99 (approved):

entropidelic (approved):

PR #1255, opened 2024-10-14
feat: broken sockets metric in batcher

### LINKED ISSUE #1241: Measure broken websockets/errors  with Prometheus when sending responses in Batcher

MarcosNicolau:
**Changes**
Adds metrics for the broken sockets when sending responses in the batcher.

**Testing**
Set up a local testnet as usual, and:
- Run metrics: `make run_metrics`
- Start sending proofs and stop the execution.
- Go to `localhost:3000` and you should see a new module: `Broken sockets` 

Closes #1241


batcher/aligned-batcher/src/connection.rs
```diff
@@ -40,12 +43,15 @@ pub(crate) async fn send_batch_inclusion_data_responses(
...
 
         match sending_result {
-            Err(Error::AlreadyClosed) => (),
-            Err(e) => error!("Error while sending batch inclusion data response: {}", e),
-            Ok(_) => (),
+            Err(Error::ConnectionClosed) | Err(Error::AlreadyClosed) | Err(Error::Io(_)) => {
+                metrics::BROKEN_SOCKETS_LATEST_BATCH.inc();
+                error!("Error while sending the batch response, socket connection was closed!");
+            }
+            Err(e) => {
```
MarcosNicolau, line 50: I believe we should also take into account these cases in the metrics, but I am not sure.

batcher/aligned-batcher/src/connection.rs
```diff
@@ -14,41 +14,50 @@ use tokio_tungstenite::{
...
-        match sending_result {
-            Err(Error::AlreadyClosed) => (),
-            Err(e) => error!("Error while sending batch inclusion data response: {}", e),
-            Ok(_) => (),
+            match sending_result {
+                Err(Error::ConnectionClosed) | Err(Error::AlreadyClosed) | Err(Error::Io(_)) => {
+                    self.metrics.broken_sockets_latest_batch.inc();
+                    error!("Error while sending the batch response, socket connection was closed!");
+                }
+                Err(e) => {
```
MarcosNicolau: I believe we should also take into account these cases in the metrics, but I am not sure...

batcher/aligned-batcher/src/connection.rs
```diff
@@ -14,41 +14,50 @@ use tokio_tungstenite::{
...
 
-pub(crate) async fn send_batch_inclusion_data_responses(
-    finalized_batch: Vec<BatchQueueEntry>,
-    batch_merkle_tree: &MerkleTree<VerificationCommitmentBatch>,
-) -> Result<(), BatcherError> {
-    for (vd_batch_idx, entry) in finalized_batch.iter().enumerate() {
-        let batch_inclusion_data = BatchInclusionData::new(vd_batch_idx, batch_merkle_tree);
-        let response = ResponseMessage::BatchInclusionData(batch_inclusion_data);
+impl Batcher {
+    pub(crate) async fn send_batch_inclusion_data_responses(
```
avilagaston9: I would move this function to lib.rs

entropidelic: I agree, it is better to have all the batcher methods in one place

batcher/aligned-batcher/src/connection.rs
```diff
@@ -14,41 +14,50 @@ use tokio_tungstenite::{
...
+                .await
+                .send(Message::binary(serialized_response))
+                .await;
 
-        match sending_result {
-            Err(Error::AlreadyClosed) => (),
-            Err(e) => error!("Error while sending batch inclusion data response: {}", e),
-            Ok(_) => (),
+            match sending_result {
+                Err(Error::ConnectionClosed) | Err(Error::AlreadyClosed) | Err(Error::Io(_)) => {
```
entropidelic: I think the `AlreadyClosed` error shouldn't be counted as an broken connection. When a user has two proofs in a batch sent from the same connection, the two `ws_sinks` will be a reference to the same ws socket. So when one is closed, the other one will raise a `AlreadyClosed` error when trying to send the message, but this should be an ok behaviour

batcher/aligned-batcher/src/connection.rs
entropidelic: maybe we could take them into account in the future, but they should appear in another metric. We are ok for now with these

grafana/provisioning/dashboards/aligned/aggregator_batcher.json
```diff
@@ -1241,6 +1341,6 @@
   "timezone": "browser",
   "title": "Aggregator Data",
   "uid": "aggregator",
-  "version": 10,
+  "version": 1,
   "weekStart": ""
-}
+}
```
IAvecilla: nit: Add EOL here

batcher/aligned-batcher/src/connection.rs
MarcosNicolau: Ok!

batcher/aligned-batcher/src/connection.rs
MarcosNicolau: Updated!

batcher/aligned-batcher/src/connection.rs
MarcosNicolau: Updated!

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1238,4 +1241,41 @@ impl Batcher {
...
+            let sending_result = ws_sink
+                .write()
+                .await
+                .send(Message::binary(serialized_response))
+                .await;
+
+            match sending_result {
+                Err(Error::ConnectionClosed) | Err(Error::Io(_)) => {
+                    self.metrics.broken_sockets_on_latest_batch.inc();
+                    error!("Error while sending the batch response, socket connection was closed!");
```
IAvecilla: I think this error message might be misleading, we are checking for other errors than the closed connection, maybe is worth to separate the cases or change the message to log the actual error.

MarcosNicolau: I agree! Just for the record, `IO` errs means a general error with the connection. This happens whenever you try to read or write from/to a socket that has been abnormally disconnected. The `ConnectionClosed` means that the underlying connection was gracefully closed. I'll add a comment about that in the code.

grafana/provisioning/dashboards/aligned/aggregator_batcher.json
MarcosNicolau: Updated!

batcher/aligned-batcher/src/lib.rs
MarcosNicolau: Updated!

entropidelic (approved):

batcher/aligned-batcher/src/metrics.rs
```diff
@@ -13,6 +13,7 @@ pub struct BatcherMetrics {
     pub reverted_batches: IntCounter,
     pub batcher_started: IntCounter,
     pub gas_price_used_on_latest_batch: IntGauge,
+    pub broken_ws_connections: IntGauge,
```
uri-99: This should be a counter, not a gauge

uri-99: Alternatevly if we want to count how many were resetted per batch (not the current implementation but maybe the intended functionality), we could use a [CounterVec](https://docs.rs/prometheus/latest/prometheus/type.CounterVec.html)

uri-99 (changes requested):

batcher/aligned-batcher/src/metrics.rs
MarcosNicolau: I changed it to a `IntCounter`, we might consider `CounterVec` in the future.

MarcosNicolau: Actually, the counter is ok, since we do not have a measure against a reference anymore, we are just incrementing.

uri-99 (approved):

PR #1257, opened 2024-10-15
feat: add gas price to prometheus

### LINKED ISSUE #1244: Prometheus of Gas Price of every block

avilagaston9:
**Motivation**

This PR creates a new Grafana dashboard with a panel that shows the gas_price of every block.

**Description**

- Creates a new Grafana dashboard.
- Fetches the `gas_price` from the net into the Elixir tracker every 12 seconds.
- Enables the `/metrics` endpoint with the new metric.
- Adds the new target to the Prometheus config.

Closes #1244 

## To Test
```
```
```
make telemetry_create_env
make telemetry_start
```
```
make run_metrics
```
Access to the new `Ethereum` dashboard at `http://localhost:3000`.
And then you can run transactions and visualize gas costs variations. 
You can also set the `RPC_URL` env variable on `telemetry_api/.env` to a holesky RPC and visualize real time testnet gas costs.

image


telemetry_api/lib/telemetry_api/periodically.ex
```diff
@@ -24,7 +25,9 @@ defmodule TelemetryApi.Periodic.OperatorFetcher do
   end
 
   def send_work() do
+    one_second = 1000
     :timer.send_interval(@wait_time_ms, :fetch_operators)
+    :timer.send_interval(one_second * 12, :gas_price) # every 12 seconds, once per block
```
uri-99: I would recommend once per 10 seconds, for redundancy, as not every block is exactly 12 seconds.

```suggestion
    :timer.send_interval(one_second * 10, :gas_price) # every 10 seconds, once per block + some margin
```

uri-99 (approved):

JuArce (changes requested):
You are trying to merge it to testnet. It may be wrong

prometheus/prometheus.yaml
```diff
@@ -24,3 +24,10 @@ scrape_configs:
       - targets: [ "host.docker.internal:9093" ]
         labels:
           bot: "batcher"
+  
+  - job_name: "aligned-tracker"
+    scrape_interval: 1s
+    static_configs:
+      - targets: [ "host.docker.internal:4001" ]
```
JuArce, line 31: I think it is a bad idea to scrap to the same port where the API es served. It should another one

JuArce (changes requested):

MarcosNicolau (approved):
Code LGTM, tested on my machine!

telemetry_api/lib/telemetry_api/periodically.ex
```diff
@@ -30,6 +31,8 @@ defmodule TelemetryApi.Periodic.OperatorFetcher do
   end
 
   def send_work() do
+    one_second = 1000
+    :timer.send_interval(one_second * 10, :gas_price) # every 10 seconds, once per block + some margin
```
entropidelic, line 35: maybe the sampling rate can be defined as a constant somewhere

entropidelic:
lgtm, maybe we should start using some logging library instead of the `IO.inspect`s but not really important right now

entropidelic (approved):

JuArce (approved):

PR #1260, opened 2024-10-15
feat: remove disconnected sockets from batch

### LINKED ISSUE #1242: Avoid making a batch with broken websockets, add prometheus stats of these broken websockets

MarcosNicolau:
**Changes**
Following pr #1255, this pr removes from the finalized batch the proofs whose clients connections is down. It also adds metrics for the number of dropped proofs per finalized batch. 

**Test**
First, run a local testnet as usual, then:
- Run metrics: `make run_metrics`
- Send a bunch of proofs and cancel some of them quickly
- Go to `localhost:300` and check the `Dismissed sockets` metric.
- Check that the proofs that you have canceled have not been included in the batch: ``

You can also test that effectively all proofs from the batch queue have been removed by sending proofs with `make batcher_send_burst_groth16` and in parallel sending other proofs and canceling them. You should see that newer proofs fail because of the nonce, as the cached user is removed.

Closes #1242

### DIFF
--- a/batcher/aligned-batcher/src/connection.rs
+++ b/batcher/aligned-batcher/src/connection.rs
@@ -10,7 +10,10 @@ use log::{error, info};
 use serde::Serialize;
 use tokio::{net::TcpStream, sync::RwLock};
 use tokio_tungstenite::{
-    tungstenite::{Error, Message},
+    tungstenite::{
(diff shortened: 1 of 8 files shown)


MarcosNicolau:
Closing this one as we decided not to include this logic and instead let the user recover its proof.

PR #1261, opened 2024-10-15
feat: bump gas price 10% for batcher

### LINKED ISSUE #1252: Bump base Fee 10% for Batcher

uri-99:
# This PR
- Bumps 10% the gas price calculated by the batcher, to avoid transactions hanged in the mempool.
- Refactors names of constants, as multiplier/divider is unnecesarily complex, while a simple multiplier/100 is enough
- Refactored batcher, now uses constants defined in SDK properly, instead of multiple definitions of this variable.

## To Test
This changes the SDK and the Batcher.
So, test normal flow


avilagaston9 (approved):
LGTM!

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1002,15 +995,17 @@ impl Batcher {
...
-            * U256::from(DEFAULT_AGGREGATOR_FEE_MULTIPLIER))
-            / U256::from(DEFAULT_AGGREGATOR_FEE_DIVIDER);
+            * U256::from(DEFAULT_AGGREGATOR_FEE_PERCENTAGE_MULTIPLIER))
+            / U256::from(PERCENTAGE_DIVIDER);
         let respond_to_task_fee_limit = (fee_for_aggregator
-            * U256::from(RESPOND_TO_TASK_FEE_LIMIT_MULTIPLIER))
-            / U256::from(RESPOND_TO_TASK_FEE_LIMIT_DIVIDER);
+            * U256::from(RESPOND_TO_TASK_FEE_LIMIT_PERCENTAGE_MULTIPLIER))
+            / U256::from(PERCENTAGE_DIVIDER);
+        let final_gas_price = gas_price * U256::from(GAS_PRICE_PERCENTAGE_MULTIPLIER)
```
JuArce: `final_gas_price` does not tell anything about the modification applied to the gas price

Also we should discuss where should be done this kind of actions because, the incremented gas price is being propagated across the function calls, but `fee_params.gas_price` does not say anything about a previous increment of 10%

JuArce: IMO, gas price increment of 10% should be done here

```rust 
//eth/mod.rs

pub async fn try_create_new_task(
...
) ... {  
...
.gas_price(fee_params.gas_price); <- here
...
}
```

JuArce: One more thing, with the new system of retries, this can change too

JuArce (changes requested):

entropidelic (approved):

JuArce (approved):

PR #1262, opened 2024-10-15
feat: disable verifiers v2

IAvecilla:
> [!NOTE]  
> This PR is derived from [this one](https://github.com/yetanotherco/aligned_layer/pull/1090).  
> Although this branch will be merged instead of the original, this approach simplifies reviewing the new feature.

This PR introduces the logic to request the disabled verifiers' bitmap for every block. If the batcher already has proofs stored and a verifier is disabled before grouping the proofs into a batch, the batcher iterates over the batch queue and filters out the proofs that used the disabled verifier.

### Unit Tests
There are new unit tests for the filtering function, which removes invalid proofs from the queue. To run these tests:
- Navigate to the batcher directory:  
  `cd batcher`
- Run the tests:  
  `cargo test`

### Full Flow Testing

1. Start the Aligned components (anvil, aggregator, and operator).
2. Open the batcher's configuration file at `config-files/config-batcher.yaml` and adjust the following parameters:
   ```
   block_interval: 100000
   batch_size_interval: 4
   ```
   > This ensures the batch won't be submitted while sending valid proofs, but the batch remains open, allowing you to disable the verifier later.
   
3. Run the batcher:  
   `make batcher_start_local`
   
4. Send two SP1 proofs:  
   `make batcher_send_sp1_task`
   
5. Disable the SP1 verifier:  
   `make verifier_disable_devnet VERIFIER_ID=3`
   
6. Verify that the clients stop and encounter an `InvalidProof` error. The batcher's queue should now be empty.

- Follow the same steps but modify step 4 by sending a Risc0 proof between the SP1 proofs. In this case, all the proofs will be invalidated anyways, since the Risc0 proof has a higher nonce and we discard everything from the same sender after the first invalid but it should have the proper error explaining this.
- If you do the same but the Risc0 proof was sent in the first place then all the SP1 will be invalidated but the Risc0 one will keep being added to the batch and it will persist in the batcher queue.


batcher/aligned-batcher/src/lib.rs
```diff
@@ -1017,13 +1030,27 @@ impl Batcher {
...
+        let gas_price_future = self.get_gas_price();
+        let disabled_verifiers_future = self.disabled_verifiers();
+
+        let (gas_price, disable_verifiers) =
+            tokio::join!(gas_price_future, disabled_verifiers_future);
+        let gas_price = gas_price.ok_or(BatcherError::GasPriceError)?;
+        let new_disable_verifiers =
+            disable_verifiers.map_err(|e| BatcherError::DisabledVerifiersError(e.to_string()))?;
+
+        let mut disabled_verifiers = self.disabled_verifiers.lock().await;
```
```suggestion
        let mut disabled_verifiers_lock = self.disabled_verifiers.lock().await;
```
just to make the lock a bit more explicit

batcher/aligned-batcher/src/lib.rs
```diff
@@ -48,6 +48,9 @@ pub mod sp1;
 pub mod types;
 mod zk_utils;
 
+#[cfg(test)]
+mod testonly;
```
entropidelic, line 57: maybe we could name it just `testing`? sounds a little cleaner to me

uri-99: +1

uri-99 (commented):
Apart from entropidelic's comments, code looks good. I need to test a small thing and will approve it

uri-99 (approved):

uri-99:
Setting PR as Draft because a simpler version (flush the whole queue) was implemented in the original version of this PR, [here](https://github.com/yetanotherco/aligned_layer/pull/1090)

uri-99:
Closing as simpler version was implemented, see previous comment

PR #1273, opened 2024-10-16
refactor(batcher): retry function

entropidelic:
**Motivation**

This PR implements the `retry_function()`, which receives a function to be retried using exponential backoff.

**Description**

Modifies the batcher to use `retry_function()` in the following calls:

- `user_balance_is_unlocked`
- `get_user_nonce_from_ethereum`
- `get_user_balance` 
- `get_gas_price`
- `s3::upload_object`
- `listen_new_blocks()`

Stops using our forked version of `ethers-rs` since is not longer needed.

**Observations**

The calls within the `Batcher::new()` function are not modified since in case of failure the program will panic (changes can be implemented if needed).
Also `create_new_task` was left for a different PR.

All modified calls use the following default parameters:

- `DEFAULT_MIN_DELAY  = 500`
- `DEFAULT_MAX_TIMES = 5`
- `DEFAULT_FACTOR = 2.0`

# How to Test

```
make batcher_start_local
```

Now try killing the container running the s3 service. You should see 5 warnings separated by exponential time:
```
[2024-10-17T21:14:09Z WARN  aligned_batcher] Error uploading batch to s3 dispatch failure
```

- There are also some test in `batcher/aligned-batcher/src/retry.rs`

- You can also add a `sleep()` before the function you want to test in the batcher, and then kill Anvil before the call happens.


batcher/aligned-batcher/src/lib.rs
```diff
@@ -268,7 +268,48 @@ impl Batcher {
...
+        stream_fallback: &mut SubscriptionStream<'_, Ws, Block<TxHash>>,
+    ) -> Result<Block<H256>, RetryError<()>> {
+        tokio::select! {
+            Some(block) = stream.next() => Ok(block),
+            Some(block) = stream_fallback.next() => Ok(block),
+            else => {
+                warn!("Error fetching next block.");
+                Err(RetryError::Transient)
+            }
+        }
```
entropidelic: the functionality to get new blocks works by a websocket subscription event, so we don't really need to make it retry. In the future it may be useful to add polling of new event blocks for better reliability than only the websocket events, and it will make sense to have a retry there, but for now let's just leave it like it is

batcher/aligned-batcher/Cargo.toml
```diff
@@ -19,10 +19,7 @@ bytes = "1.7.1"
 hex = "0.4.3"
 dotenvy = "0.15.0"
 anyhow = "1.0.83"
-ethers = { tag = "v2.0.15-fix-reconnections", features = [
-    "ws",
-    "rustls",
-], git = "https://github.com/yetanotherco/ethers-rs.git" }
+ethers = { version = "2.0", features = ["ws", "rustls"] }
```
Oppen, line 22: Why did we change this?

batcher/aligned-batcher/src/lib.rs
```diff
@@ -261,17 +262,35 @@ impl Batcher {
...
     pub async fn listen_new_blocks(self: Arc<Self>) -> Result<(), BatcherError> {
-        let mut stream = self
-            .eth_ws_provider
-            .subscribe_blocks()
-            .await
-            .map_err(|e| BatcherError::EthereumSubscriptionError(e.to_string()))?;
+        retry_function(
+            || {
+                let app = self.clone();
+                async move { app.listen_new_blocks_retryable().await }
```
Oppen, line 261: Does this work? Isn't this kind of like having an actual async lambda?

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1272,55 +1296,102 @@ impl Batcher {
...
+        s3_bucket_name: &str,
+    ) -> Result<(), RetryError<String>> {
+        s3::upload_object(&s3_client, s3_bucket_name, batch_bytes.to_vec(), file_name)
             .await
-            .inspect_err(|e| warn!("Failed to get gas price: {e:?}"))
-            .ok()
+            .map_err(|e| {
+                warn!("Error uploading batch to s3 {e}");
+                RetryError::Transient(e.to_string())
+            })?;
```
Oppen, line 1387: I wonder whether all errors are actually transient for S3.

Oppen (approved):
Basic test seems to work. Question: how do we determine which operations need retry to look for completeness?

batcher/aligned-batcher/src/connection.rs
```diff
@@ -66,3 +58,36 @@ pub(crate) async fn send_message<T: Serialize>(ws_conn_sink: WsMessageSink, mess
...
+    serialized_response: Vec<u8>,
+) -> Result<(), RetryError<Error>> {
+    let sending_result = ws_sink
+        .write()
+        .await
+        .send(Message::binary(serialized_response))
+        .await;
+
+    match sending_result {
+        Err(Error::AlreadyClosed) => Err(RetryError::Permanent(Error::AlreadyClosed)),
```
MarcosNicolau: We might want some other errs here such as: `Err(Error::ConnectionClosed) | Err(Error::AlreadyClosed) | Err(Error::Io(_))`.

MarcosNicolau: Because `AlreadyClosed` will occur when the connection has been gracefully closed. If the client has `Forcefully` closed the connection, then we'll get an `IO` err, and so the retry would be infinite. Maybe, we can close the connection

batcher/aligned-batcher/src/lib.rs
```diff
@@ -261,17 +263,48 @@ impl Batcher {
...
-            .eth_ws_provider_fallback
-            .subscribe_blocks()
-            .await
-            .map_err(|e| BatcherError::EthereumSubscriptionError(e.to_string()))?;
+    pub async fn listen_new_blocks_retryable(
+        self: Arc<Self>,
+    ) -> Result<(), RetryError<BatcherError>> {
+        let eth_ws_provider = Provider::connect(&self.eth_ws_url).await.map_err(|e| {
+            warn!("Failed to get ethereum websocket provider");
+            RetryError::Transient(BatcherError::EthereumSubscriptionError(e.to_string()))
```
MarcosNicolau, line 276: Returning the err here will never use the fallback url.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -261,17 +263,48 @@ impl Batcher {
...
+            Provider::connect(&self.eth_ws_url_fallback)
+                .await
+                .map_err(|e| {
+                    warn!("Failed to get fallback ethereum websocket provider");
+                    RetryError::Transient(BatcherError::EthereumSubscriptionError(e.to_string()))
+                })?;
+
+        let mut stream = eth_ws_provider.subscribe_blocks().await.map_err(|e| {
+            warn!("Error subscribing to blocks.");
+            RetryError::Transient(BatcherError::EthereumSubscriptionError(e.to_string()))
```
MarcosNicolau, line 289: Ditto

batcher/aligned-batcher/src/lib.rs
MarcosNicolau: I tested it and it worked for me.

batcher/aligned-batcher/src/connection.rs
MarcosNicolau: Or maybe don't if we have a low `MaxRetry`

batcher/aligned-batcher/src/lib.rs
```diff
@@ -261,17 +263,48 @@ impl Batcher {
...
-            .await
-            .map_err(|e| BatcherError::EthereumSubscriptionError(e.to_string()))?;
+        retry_function(
+            || {
+                let app = self.clone();
+                async move { app.listen_new_blocks_retryable().await }
+            },
+            DEFAULT_MIN_DELAY,
+            DEFAULT_FACTOR,
+            DEFAULT_MAX_TIMES,
```
MarcosNicolau: `DEFAULT_MAX_TIMES` is constant across the whole execution of the batcher. It would be good to restart the max times upon a correct response. That or increase the value.

MarcosNicolau: See if this applies to other points.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -261,17 +263,48 @@ impl Batcher {
...
-        let mut stream_fallback = self
-            .eth_ws_provider_fallback
-            .subscribe_blocks()
-            .await
-            .map_err(|e| BatcherError::EthereumSubscriptionError(e.to_string()))?;
+    pub async fn listen_new_blocks_retryable(
+        self: Arc<Self>,
+    ) -> Result<(), RetryError<BatcherError>> {
+        let eth_ws_provider = Provider::connect(&self.eth_ws_url).await.map_err(|e| {
+            warn!("Failed to get ethereum websocket provider");
```
```suggestion
            warn!("Failed to instantiate Ethereum websocket provider");
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -261,17 +263,48 @@ impl Batcher {
...
+        let eth_ws_provider = Provider::connect(&self.eth_ws_url).await.map_err(|e| {
+            warn!("Failed to instantiate Ethereum websocket provider");
+            RetryError::Transient(BatcherError::EthereumSubscriptionError(e.to_string()))
+        })?;
+
+        let eth_ws_provider_fallback =
+            Provider::connect(&self.eth_ws_url_fallback)
+                .await
+                .map_err(|e| {
+                    warn!("Failed to get fallback ethereum websocket provider");
```
```suggestion
                    warn!("Failed to instantiate fallback Ethereum websocket provider");
```

batcher/aligned-batcher/src/connection.rs
```diff
@@ -33,17 +35,7 @@ pub(crate) async fn send_batch_inclusion_data_responses(
...
-            .await
-            .send(Message::binary(serialized_response))
-            .await;
-
-        match sending_result {
-            Err(Error::AlreadyClosed) => (),
-            Err(e) => error!("Error while sending batch inclusion data response: {}", e),
-            Ok(_) => (),
-        }
+        send_response(ws_sink, serialized_response).await;
```
MarcosNicolau: I would return a `Result` in `send_response` and show the log accordingly if it was successfully sent or not.

batcher/aligned-batcher/src/lib.rs
MarcosNicolau: I guess it doesn't matter if we have a `MaxRetry` right?

MarcosNicolau (commented):
Other than the minor comments, everything seems to work fine, these are the cases I tested:
- The one in the comment: stopping `s3`, see that it retries, starting it back again and it succeeded.
- Restarting anvil: stopped anvil and checked that it retries to get new blocks and subscribe to the socket. This had one problem, it would retry up to `MaxRetry` without restarting the value once it responded.
- Disconnected client: see that it retries to send a response.

Also, I would increase the `MaxRetry` I think 2 is very low.

batcher/aligned-batcher/src/retry.rs
```diff
@@ -0,0 +1,286 @@
...
+        let balance =
+            get_user_balance_retryable(&payment_service, &payment_service, &dummy_user_addr)
+                .await
+                .unwrap();
+
+        assert_eq!(balance, U256::zero());
+    }
+
+    #[tokio::test]
+    async fn test_user_balance_is_unlocked_retryable_kill_anvil() {
```
entropidelic:
```suggestion
    async fn test_user_balance_is_unlocked_retryable() {
```

batcher/aligned-batcher/src/retry.rs
```diff
@@ -0,0 +1,286 @@
...
+            &dummy_user_addr,
+        )
+        .await
+        .unwrap();
+
+        assert_eq!(unlocked, false);
+    }
+
+    #[tokio::test]
+    async fn test_get_user_nonce_retryable_kill_anvil() {
```
entropidelic:
```suggestion
    async fn test_get_user_nonce_retryable() {
```

batcher/aligned-batcher/src/lib.rs
MarcosNicolau: Github does not show the line, but here is a `?`.

batcher/aligned-batcher/src/connection.rs
entropidelic: good catch @MarcosNicolau, the `AlreadyClosed` error shouldn't be considered an error. It will raise in normal situations, so it should just be ignored:
```rust
Err(Error::AlreadyClosed) => ()
```
In other words, it should be an `Ok()` case.
The same applies for the `ConnectionClosed` error.

I think all other cases should be Permanent for the moment. You can see more details in https://docs.rs/tungstenite/latest/tungstenite/error/enum.Error.html

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1272,55 +1310,102 @@ impl Batcher {
...
+                    &self.payment_service,
+                    &self.payment_service_fallback,
+                    addr,
+                )
+            },
+            DEFAULT_MIN_DELAY,
+            DEFAULT_FACTOR,
+            DEFAULT_MAX_TIMES,
+        )
+        .await
```
```suggestion
        let Ok(unlocked) = retry_function(
            || {
                user_balance_is_unlocked_retryable(
                    &self.payment_service,
                    &self.payment_service_fallback,
                    addr,
                )
            },
            DEFAULT_MIN_DELAY,
            DEFAULT_FACTOR,
            DEFAULT_MAX_TIMES,
        )
        .await else {
             warn!("Could not get user locking state.");
              return false;
        }
```

batcher/aligned-batcher/src/retry.rs
```diff
@@ -0,0 +1,286 @@
...
+            dummy_user_addr,
+        )
+        .await
+        .unwrap();
+
+        assert_eq!(nonce, U256::zero());
+    }
+
+    #[tokio::test]
+    async fn test_get_gas_price_retryable_kill_anvil() {
```
entropidelic:
```suggestion
    async fn test_get_gas_price_retryable() {
```

batcher/aligned-batcher/src/lib.rs
avilagaston9: I think that by using a large number of retries we avoid this issue

batcher/aligned-batcher/src/connection.rs
avilagaston9: Since all errors should be Permanent, I removed the retry logic from the `send_response()` method.

batcher/aligned-batcher/src/lib.rs
avilagaston9: We want both providers to be working before starting the listen loop. We will create an issue to revisit `listen_new_blocks` anyway.

batcher/aligned-batcher/src/connection.rs
MarcosNicolau: Perfect!

batcher/aligned-batcher/src/lib.rs
MarcosNicolau: Ok!

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1272,55 +1313,100 @@ impl Batcher {
...
+        else {
+            warn!("Could not get user locking state.");
+            return false;
+        };
+        unlocked
     }
 
-    /// Gets the current gas price from Ethereum.
+    /// Gets the current gas price from Ethereum using exponential backoff.
     /// Returns `None` if the gas price couldn't be returned
```
```suggestion
    /// Returns `None` if the gas price couldn't be returned
```
```suggestion
```

MarcosNicolau (approved):

batcher/aligned-batcher/src/connection.rs
MarcosNicolau: Discussed offline.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1272,55 +1313,99 @@ impl Batcher {
...
+                    addr,
+                )
+            },
+            DEFAULT_MIN_DELAY,
+            DEFAULT_FACTOR,
+            DEFAULT_MAX_TIMES,
+        )
+        .await
+        else {
+            warn!("Could not get user locking state.");
```
uri-99, line 1332: return err would be more appropriate

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1272,55 +1313,99 @@ impl Batcher {
...
+                    self.s3_client.clone(),
+                    &self.s3_bucket_name,
+                )
+            },
+            DEFAULT_MIN_DELAY,
+            DEFAULT_FACTOR,
+            DEFAULT_MAX_TIMES,
+        )
+        .await
+        .map_err(|_| BatcherError::BatchUploadError("Error uploading batch to s3".to_string()))
```
uri-99: `.map_err(|_|` is not a good technical debt to have. is there a way to forward this error? maybe batchUploadError could take `e` as parameter.

batcher/aligned-batcher/src/retry.rs
```diff
@@ -0,0 +1,317 @@
...
+use backon::Retryable;
+use ethers::prelude::*;
+use log::warn;
+use std::{future::Future, time::Duration};
+
+use crate::eth::payment_service::BatcherPaymentService;
+
+pub const DEFAULT_MIN_DELAY: u64 = 2000;
+pub const DEFAULT_MAX_TIMES: usize = 3;
+pub const DEFAULT_FACTOR: f32 = 2.0;
```
uri-99: should these be defined in `constants.rs` ?

batcher/aligned-batcher/src/retry.rs
```diff
@@ -0,0 +1,317 @@
...
+
+#[derive(Debug)]
+pub enum RetryError<E> {
+    Transient(E),
+    Permanent(E),
```
uri-99: Permanent is not really used anywhere

batcher/aligned-batcher/src/lib.rs
Oppen: It depends on whether the delay matters. Failing fast might be a good thing sometimes.

batcher/aligned-batcher/src/retry.rs
```diff
@@ -0,0 +1,317 @@
...
+        .user_balances(*addr)
+        .call()
+        .await
+        .map_err(|e| {
+            warn!("Failed to get balance for address {:?}. Error: {e}", addr);
+            RetryError::Transient(e.to_string())
+        })
+}
+
+pub async fn get_user_nonce_from_ethereum_retryable(
```
uri-99: Adding these functions logics here seems odd.
Maybe the outer function could be in this file, which contains the call to the core function with the retries; and the core function that contains the logic could go back to `lib.rs`

batcher/aligned-batcher/src/lib.rs
uri-99: If having `MaxRetry` invalidates Permanent errors, there is no use case for Permanent errors.
We are missing the differentiation of Transient and Permanent errors in this PR, but it could be added in a later issue

batcher/aligned-batcher/src/retry.rs
```diff
@@ -0,0 +1,317 @@
...
+        let anvil = Anvil::new()
+            .port(port)
+            .arg("--load-state")
+            .arg("../../contracts/scripts/anvil/state/alignedlayer-deployed-anvil-state.json")
+            .spawn();
+
+        let eth_rpc_provider = eth::get_provider(format!("http://localhost:{}", port))
+            .expect("Failed to get provider");
+
+        let payment_service_addr = String::from("0x7969c5eD335650692Bc04293B07F5BF2e7A673C0");
```
uri-99: _cries in hardcoded_

uri-99: Is reading the deployment output file an overkill? How difficult would it be?
They won't change frequently, but the test will fail when they do.

batcher/aligned-batcher/src/retry.rs
```diff
@@ -0,0 +1,317 @@
...
+            },
+            payment_service_addr,
+        )
+        .await
+        .expect("Failed to get Batcher Payment Service contract");
+        (anvil, payment_service)
+    }
+
+    #[tokio::test]
+    async fn test_get_user_balance_retryable() {
```
uri-99: Shouldn't we test the actual retry function? Instead of manually calling the `_retryable` , calling the function that has the retries in it.

uri-99 (changes requested):

batcher/aligned-batcher/src/lib.rs
Oppen: I'm not opposed at all to adding it later. I was mostly wondering if all errors were recoverable (it can be the case for some protocols I guess) and this was intentional or an oversight.

batcher/aligned-batcher/src/retry.rs
Oppen: These should probably have more explicit names. `DEFAULT_FACTOR` on use context doesn't let me know what factor we're talking about. `DEFAULT_BACKOFF_FACTOR`, `DEFAULT_MAX_RETRIES` and `DEFAULT_MIN_RETRY_DELAY` are what I would suggest.

batcher/aligned-batcher/src/retry.rs
Oppen: I agree, this doesn't seem to be the right place for the business logic. The mechanism belongs here.

Oppen: I'm inclined to guess they're here for the tests. You can just import the functions inside `mod test`, while placing them near their call sites.

batcher/aligned-batcher/src/retry.rs
Oppen: How much parsing do they require? The file themselves may be `include_str!`'d to avoid dealing with files.

batcher/aligned-batcher/src/lib.rs
avilagaston9: Done!

batcher/aligned-batcher/src/retry.rs
avilagaston9: Done!

batcher/aligned-batcher/src/retry.rs
avilagaston9: We use `Permanent` errors in the PR that adds the retry logic to the `create_new_task` function, but we need to revisit `Transient` vs. `Permanent` errors in the future anyway.

batcher/aligned-batcher/src/retry.rs
avilagaston9: Moved the retryable functions to `retry::batcher_retry` module

batcher/aligned-batcher/src/retry.rs
avilagaston9: Discussed offline. We need to revisit how to test the retry functionality properly but in a different PR.

batcher/aligned-batcher/src/lib.rs
avilagaston9: Discussed offline. In this PR, we aimed not to modify the original flow of the batcher, only the inner functions.

uri-99: https://github.com/yetanotherco/aligned_layer/issues/1312

batcher/aligned-batcher/src/retry.rs
uri-99: wip renaming according to Oppen's suggestion

batcher/aligned-batcher/src/retry.rs
avilagaston9: Modified using `ContractDeploymentOutput::new()`, which is how we handle this in the batcher.

batcher/aligned-batcher/src/retry.rs
avilagaston9: Done!

uri-99 (approved):

batcher/aligned-batcher/src/lib.rs
```diff
@@ -220,12 +218,20 @@ impl Batcher {
         }
         .expect("Failed to get disabled verifiers");
 
+        let eth_http_provider =
```
JuArce: It is being created three https providers for the normal and three for the fallback. I think creating just one and reusing it is enough

batcher/aligned-batcher/src/retry/batcher_retryables.rs
```diff
@@ -0,0 +1,270 @@
...
+            let _anvil;
+            (_anvil, payment_service) = setup_anvil(8545u16).await;
+
+            let balance =
+                get_user_balance_retryable(&payment_service, &payment_service, &dummy_user_addr)
+                    .await
+                    .unwrap();
+
+            assert_eq!(balance, U256::zero());
+            // Kill anvil
```
JuArce:
```suggestion
            // Anvil is killed when the scope is left
```

batcher/aligned-batcher/src/retry/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+        match self {
+            RetryError::Transient(e) => e,
+            RetryError::Permanent(e) => e,
+        }
+    }
+}
+
+impl<E: std::fmt::Display> std::error::Error for RetryError<E> where E: std::fmt::Debug {}
+
+pub async fn retry_function<FutureFn, Fut, T, E>(
```
JuArce, line 34: Since the implementation of exp backoff is different if the fn is async or blocking, I'd specify which type we are covering (async).
Also I'd add refference to the library https://docs.rs/backon/latest/backon/#retry-an-async-function

JuArce (changes requested):
This provider is using main rpc instead of fallback https://github.com/yetanotherco/aligned_layer/pull/1273/files#diff-fb54942eba81c272160ef10dfb2972fd0a6de56d45ed87c98ca1a65b3b2dd29bR126-R127

batcher/aligned-batcher/src/lib.rs
avilagaston9: Done!

batcher/aligned-batcher/src/retry/batcher_retryables.rs
avilagaston9: Updated!

batcher/aligned-batcher/src/retry/mod.rs
avilagaston9: Added!

JuArce (approved):

PR #1286, opened 2024-10-17
feat: aggregator bump fee if transaction was not included

### LINKED ISSUE #311: ( aggregator ) Bump fees
- [ ] Test bump fee mechanism with any bump
- [ ] Bump fee 50% after 3 blocks/30s, bump to the limit of the Agg after 5

MarcosNicolau:
Before merging this pr, pr #1304 should be merged first, as it depends on the retry logic.

### Changes
Bumps the `gasPrice` in the aggregator when sending a responded task:
- After a timeout period the aggregator will send a new `ReponseToTask` with a higher bumped gas price.
- Configurations fields have been added to easily change the behavior: 
   - `gas_base_bump_percentage`: how much to bump the gas price. Configured to `10%`
   - `gas_bump_incremental_percentage`: an extra percentage to bump every retry `i*5`. Configured to `2%`
   - `time_to_wait_before_bump`: the time to wait for the receipt once we sent the tx. Configured to `36s`
- Traces for tasks:  a new endpoint in the telemetry to log when the gas price is bumped
Screenshot 2024-11-11 at 11 34 02 AM

- Time series in grafana: shows the number of times the gas was bumped when sending task response across time.
image

---

**Considerations:**

For every retry, the gas price is bumped a `10%` + `2% * i`, where `i` is the iteration number. `i` is only incremented when the receipt timeout has passed. That is, if the rpc node fails the fee `i` won't be incremented. This will run infinitely until the transaction gets accepted **unless the `checkRespondToTaskFeeLimit` fails**. In that case, we stop retrying and leave the tx in the mempool. Issue #1389 has been created to cover this case.

Another note is that `waitForTransactionReceipt` has been refactored. Now it does not have a `maxTries` anymore (set to `0`) and instead receives a new parameter `waitTimeout` which defines the retries based on a timeout. Also, the `maxInterval` has been modified to `2` seconds because we can't reliably measure the specific time the tx will be included in a block. Setting a higher value will imply doing fewer retries across the waitTimeout so we might lose the receipt.

### Test

Setup a local testnet as usual and introduce the following lines in the `avs_writer.go` file:

```go
125:   if i < 2 {
126:	    i++
127:	    return nil, fmt.Errorf("error")
130:   }
```

You should check through the aggregator logs, telemetry traces, and grafana charts, that the gas price was indeed bumped.


core/utils/eth_client_utils.go
```diff
@@ -42,3 +48,51 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
 	}
 	return quorumThresholdPercentages
 }
+
+// Very basic algorithm to calculate the gasPrice bump based on the currentGasPrice and retry iteration.
+// It adds a i/10 percentage to the current prices, where i represents the iteration.
+func CalculateGasPriceBumpBasedOnRetry(currentGasPrice *big.Int, iteration int) *big.Int {
```
MarcosNicolau: This is a very basic implementation to have something. We might one to change it for something more appropriate 🤷.

core/chainio/avs_writer.go
```diff
@@ -70,10 +72,11 @@ func NewAvsWriterFromConfig(baseConfig *config.BaseConfig, ecdsaConfig *config.E
 	}, nil
 }
 
-func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMerkleRoot [32]byte, senderAddress [20]byte, nonSignerStakesAndSignature servicemanager.IBLSSignatureCheckerNonSignerStakesAndSignature) (*common.Hash, error) {
+func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMerkleRoot [32]byte, senderAddress [20]byte, nonSignerStakesAndSignature servicemanager.IBLSSignatureCheckerNonSignerStakesAndSignature) (*types.Receipt, error) {
```
entropidelic: for what I understand, this function has two parts:

1. One where a simulation of the transaction is made
2. One part with the logic that is going to be retried if the transaction is not included.

I think it is easier to follow the code if we just encapsulate all the logic in 2 in an anonymous function and just call `Retry`. It may be a little more monolithic but it is not really a problem

core/connection.go
```diff
@@ -0,0 +1,81 @@
...
+	return e.Inner
+}
+func (e PermanentError) Is(err error) bool {
+	_, ok := err.(PermanentError)
+	return ok
+}
+
+// Same as Retry only that the functionToRetry can return a value upon correct execution
+func RetryWithData[T any](functionToRetry func() (*T, error), minDelay uint64, factor float64, maxTries uint64) (*T, error) {
+	i := 0
```
entropidelic: this `i` variable is stale

core/connection.go
```diff
@@ -0,0 +1,81 @@
...
+	f := func() (*T, error) {
+		val, err := functionToRetry()
+		i++
```
entropidelic: ditto

core/connection.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Retries a given function in an exponential backoff manner.
+// It will retry calling the function while it returns an error, until the max retries.
+// If maxTries == 0 then the retry function will run indefinitely until success
+// from the configuration are reached, or until a `PermanentError` is returned.
+// The function to be retried should return `PermanentError` when the condition for stop retrying
+// is met.
+func Retry(functionToRetry func() error, minDelay uint64, factor float64, maxTries uint64) error {
+	i := 0
```
entropidelic: ditto

core/connection.go
```diff
@@ -0,0 +1,81 @@
...
+	f := func() error {
+		err := functionToRetry()
+		i++
```
entropidelic: ditto

core/utils/eth_client_utils.go
```diff
@@ -42,3 +48,46 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
...
+	factor := (new(big.Int).Add(big.NewInt(100), new(big.Int).Mul(big.NewInt(int64(iteration)), big.NewInt(10))))
+	gasPrice := new(big.Int).Mul(currentGasPrice, factor)
+	gasPrice = gasPrice.Div(gasPrice, big.NewInt(100))
```
entropidelic: very small nit :), but it could make the math a little clearer. We can make two variables with names like `factorNumerator` and `factorDenominator`, and then the gas price will be
```
gasPrice := currentGasPrice * factorNumerator / factorDenominator
```

core/chainio/avs_writer.go
MarcosNicolau: Updated!

core/utils/eth_client_utils.go
MarcosNicolau: Updated! Though `big.Int` math is horrible to work with, you can't use the normal operators.

core/chainio/avs_writer.go
```diff
@@ -87,21 +90,52 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
-		tx, err = w.AvsContractBindings.ServiceManagerFallback.RespondToTaskV2(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
+	txOpts.Nonce = txNonce
+
+	// Sends a transaction and waits for the receipt for three blocks, if not received
+	// it will try again bumping the gas price based on `CalculateGasPriceBumpBasedOnRetry`
+	// This process happens indefinitely until we sendTransaction does not return err.
+	i := 0
+	sendTransaction := func() (*types.Receipt, error) {
+		i++
+		gasPrice := utils.CalculateGasPriceBumpBasedOnRetry(tx.GasPrice(), i)
```
MarcosNicolau: Maybe here we could re-run the simulation to get an updated `gasPrice`?

MarcosNicolau: Done.

core/chainio/avs_writer.go
```diff
@@ -87,21 +98,61 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
+		tx, err = w.AvsContractBindings.ServiceManager.RespondToTaskV2(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
+		if err != nil {
+			// Retry with fallback
+			tx, err = w.AvsContractBindings.ServiceManagerFallback.RespondToTaskV2(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
+			if err != nil {
+				return nil, connection.PermanentError{Inner: err}
+			}
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*36)
```
entropidelic: I think thie timeout constant we are using should be a little more visible, also a comment explaining the criteria for choosing that number would be useful ("the average time of 3 blocks", etc)

core/connection_test.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func TestRetry(t *testing.T) {
+	function := func() error {
+		_, err := DummyFunction(43)
+		return err
+	}
+	err := connection.Retry(function, 1000, 2, 3)
+	if err != nil {
+		t.Errorf("Retry error!: %s", err)
```
entropidelic: very small nit but let's leave this tests a little better :) I did this ones only for quick testing of the original `Retry` function. 
Maybe we can make three distinct test cases, one for `x < 42`, one for `x == 42` and one for `x > 42`, it should be really quick to do

core/chainio/avs_writer.go
MarcosNicolau: Updated!

core/connection_test.go
MarcosNicolau: Updated!

entropidelic (approved):

core/chainio/avs_writer.go
```diff
@@ -70,10 +78,15 @@ func NewAvsWriterFromConfig(baseConfig *config.BaseConfig, ecdsaConfig *config.E
 	}, nil
 }
 
-func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMerkleRoot [32]byte, senderAddress [20]byte, nonSignerStakesAndSignature servicemanager.IBLSSignatureCheckerNonSignerStakesAndSignature) (*common.Hash, error) {
+// Sends AggregatedResponse and waits for the receipt for three blocks, if not received
+// it will try again bumping the last tx gas price based on `CalculateGasPriceBump`
+// This process happens indefinitely until the transaction is included.
+// Note: If the rpc endpoints fail, the retry will stop, as it will infinitely try
```
uri-99: `// Note: If the rpc endpoints fail, the retry will stop, as it will infinitely try`
It will infinitely try to what? Try to reconnect to the RPC?
After it has reconnected does it try to send the aggregated response again?

core/chainio/avs_writer.go
```diff
@@ -87,21 +100,61 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
 	txOpts.NoSend = false
-	txOpts.GasLimit = tx.Gas() * 110 / 100 // Add 10% to the gas limit
-	tx, err = w.AvsContractBindings.ServiceManager.RespondToTaskV2(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
-	if err != nil {
-		// Retry with fallback
-		tx, err = w.AvsContractBindings.ServiceManagerFallback.RespondToTaskV2(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
+	txOpts.Nonce = txNonce
+
+	i := 0
+	sendTransaction := func() (*types.Receipt, error) {
```
uri-99: i would try make this a separate function, it is quite big for an anonymous one

core/chainio/avs_writer.go
```diff
@@ -87,21 +100,61 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
+		if receipt != nil {
+			return receipt, nil
+		}
+		// if we are here, this means we have reached the timeout (after three blocks it hasn't been included)
+		// so we try again by bumping the fee to make sure its included
 		if err != nil {
 			return nil, err
 		}
+		return nil, fmt.Errorf("transaction failed")
 	}
```
uri-99, line 143: isn't this missing a `()` at the end? to close the anonymous function

core/utils/eth_client_utils.go
```diff
@@ -42,3 +47,15 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
 	}
 	return quorumThresholdPercentages
 }
+
+// Very basic algorithm to calculate the gasPrice bump based on the currentGasPrice a constant percentage and the retry number.
```
uri-99:
```suggestion
// Simple algorithm to calculate the gasPrice bump based on: the currentGasPrice, a constant percentage, and the retry number.
```

core/utils/eth_client_utils.go
```diff
@@ -42,3 +47,15 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
...
+// It adds a the percentage to the current gas price and a 5% * i, where i is the iteration number. That is:
```
uri-99: Write the formula directly is better and more clear.
Im not sure which is the intended formula, but for example:
```
// new_gas_price = current_gas_price + current_gas_price * constant_percentage_increase + current_gas_price * variable_percentage_increase * iteration
```

core/utils/eth_client_utils.go
```diff
@@ -42,3 +47,15 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
...
+func CalculateGasPriceBumpBasedOnRetry(currentGasPrice *big.Int, percentage int, i int) *big.Int {
+	retryPercentage := new(big.Int).Mul(big.NewInt(5), big.NewInt(int64(i)))
```
uri-99: Move magic number `5` to a better defined constant

core/utils/eth_client_utils.go
```diff
@@ -42,3 +47,15 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
...
+	percentageBump := new(big.Int).Add(big.NewInt(int64(percentage)), retryPercentage)
+	bumpAmount := new(big.Int).Mul(currentGasPrice, percentageBump)
+	bumpAmount = new(big.Int).Div(bumpAmount, big.NewInt(100))
+	bumpedGasPrice := new(big.Int).Add(currentGasPrice, bumpAmount)
```
uri-99: Rename the variables. `percentage` is not a good variable name here, there are many percentages moving around, it is not easy to read and follow.

It is a simple formula, it should be easy to read to verify the formula is correctly implemented

core/chainio/avs_writer.go
```diff
@@ -87,21 +100,61 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
+
+	i := 0
+	sendTransaction := func() (*types.Receipt, error) {
+		if i > 0 {
+			onRetry()
+		}
+		i++
+
+		// get gas price
+		gasPrice, err := w.Client.SuggestGasPrice(context.Background())
```
uri-99: I was not aware of this `SuggestGasPrice` function, it looks pretty useful. Did you test it returns accurate values?

core/chainio/avs_writer.go
```diff
@@ -87,21 +100,61 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
+				return nil, fmt.Errorf("transaction simulation failed: %v", err)
+			}
+		}
+		txOpts.GasPrice = utils.CalculateGasPriceBumpBasedOnRetry(gasPrice, GasBumpPercentage, i)
+
+		w.logger.Infof("Sending ResponseToTask transaction with a gas price of %v", txOpts.GasPrice)
+		err = w.checkRespondToTaskFeeLimit(tx, txOpts, batchIdentifierHash, senderAddress)
+
+		if err != nil {
+			return nil, connection.PermanentError{Inner: err}
```
uri-99: This error is not permanent, it can be fixed by a favourable gas price change

core/chainio/avs_writer.go
```diff
@@ -87,21 +100,61 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
+		}
+
+		tx, err = w.AvsContractBindings.ServiceManager.RespondToTaskV2(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
+		if err != nil {
+			// Retry with fallback
+			tx, err = w.AvsContractBindings.ServiceManagerFallback.RespondToTaskV2(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
+			if err != nil {
+				return nil, connection.PermanentError{Inner: err}
```
uri-99: Why is this error permanent? what are the possible errors?
an RPC connection error could be fixed with a retry if the network suffered a temporal downtime

core/chainio/avs_writer.go
```diff
@@ -15,7 +15,15 @@ import (
...
 	servicemanager "github.com/yetanotherco/aligned_layer/contracts/bindings/AlignedLayerServiceManager"
+	connection "github.com/yetanotherco/aligned_layer/core"
 	"github.com/yetanotherco/aligned_layer/core/config"
+	"github.com/yetanotherco/aligned_layer/core/utils"
+)
+
+const (
+	GasBumpPercentage int = 20
+	// wait as much as 3 blocks time for the receipt
+	SendAggregateResponseReceiptTimeout time.Duration = time.Second * 36
```
uri-99: these variables should be read from agg config file

core/chainio/avs_writer.go
```diff
@@ -87,21 +100,61 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
+		if err != nil {
+			// Retry with fallback
+			tx, err = w.AvsContractBindings.ServiceManagerFallback.RespondToTaskV2(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
+			if err != nil {
+				return nil, connection.PermanentError{Inner: err}
+			}
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), SendAggregateResponseReceiptTimeout)
+		defer cancel()
```
uri-99: What is this cancelling?

core/utils/eth_client_utils.go
```diff
@@ -17,12 +18,16 @@ const sleepTime = 5 * time.Second
 
 func WaitForTransactionReceipt(client eth.InstrumentedClient, ctx context.Context, txHash gethcommon.Hash) (*types.Receipt, error) {
 	for i := 0; i < maxRetries; i++ {
-		receipt, err := client.TransactionReceipt(ctx, txHash)
-		if err != nil {
-			time.Sleep(sleepTime)
-		} else {
+		receipt, _ := client.TransactionReceipt(ctx, txHash)
```
uri-99: why remove the usage of `TransactionReceipt` err return?

core/utils/eth_client_utils.go
```diff
@@ -17,12 +18,16 @@ const sleepTime = 5 * time.Second
...
+		if receipt != nil {
 			return receipt, nil
 		}
+		// if context has timed out, return
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		} else {
+			time.Sleep(sleepTime)
+		}
```
uri-99:
```suggestion
		}
		time.Sleep(sleepTime)
```

core/connection.go
```diff
@@ -0,0 +1,77 @@
+package connection
+
+import (
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+)
+
+type PermanentError struct {
```
uri-99: is there no TemporalError?

uri-99 (changes requested):
Still missing review of `core/connection.go` and `core/connection_test.go`

core/chainio/avs_writer.go
MauroToscano: Is it a timeout ? Or something like a BlocksToWaitBeforeBumping ?

core/utils/eth_client_utils.go
JuArce: I agree with Uri. The function is hard to understand

JuArce: Also there is a 5% hardcoded and it also receives a percentage parameter. It is not clear

core/chainio/avs_writer.go
```diff
@@ -87,21 +100,61 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
+			gasPrice, err = w.ClientFallback.SuggestGasPrice(context.Background())
+			if err != nil {
+				return nil, connection.PermanentError{Inner: err}
+			}
+		}
+		txOpts.GasPrice = utils.CalculateGasPriceBumpBasedOnRetry(gasPrice, GasBumpPercentage, i)
+
+		w.logger.Infof("Sending ResponseToTask transaction with a gas price of %v", txOpts.GasPrice)
+		err = w.checkRespondToTaskFeeLimit(tx, txOpts, batchIdentifierHash, senderAddress)
+
```
JuArce, line 112:
```suggestion
```

core/connection.go
```diff
@@ -0,0 +1,77 @@
...
+func (e PermanentError) Unwrap() error {
+	return e.Inner
+}
+func (e PermanentError) Is(err error) bool {
+	_, ok := err.(PermanentError)
+	return ok
+}
+
+// Same as Retry only that the functionToRetry can return a value upon correct execution
+func RetryWithData[T any](functionToRetry func() (*T, error), minDelay uint64, factor float64, maxTries uint64) (*T, error) {
```
JuArce: Add a comment with the meaning of sending maxTries=0

core/connection.go
```diff
@@ -0,0 +1,77 @@
...
+	f := func() (*T, error) {
+		val, err := functionToRetry()
+		if perm, ok := err.(PermanentError); err != nil && ok {
```
JuArce: Type assertion on errors fails on wrapped errors

core/chainio/avs_writer.go
```diff
@@ -87,21 +100,61 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
+	txNonce := new(big.Int).SetUint64(tx.Nonce())
 	txOpts.NoSend = false
-	txOpts.GasLimit = tx.Gas() * 110 / 100 // Add 10% to the gas limit
-	tx, err = w.AvsContractBindings.ServiceManager.RespondToTaskV2(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
-	if err != nil {
-		// Retry with fallback
-		tx, err = w.AvsContractBindings.ServiceManagerFallback.RespondToTaskV2(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
+	txOpts.Nonce = txNonce
+
+	i := 0
```
JuArce, line 96: Is this goroutine/thread safe?

JuArce (changes requested):
How to test this PR is not too clear
Also, it should add a note in the description that it requires a deploy of new metrics/dashboards

core/chainio/avs_writer.go
MarcosNicolau: Ups, I forgot to end that sentence...

core/chainio/avs_writer.go
MarcosNicolau: AFAIK, there isn't a goroutine here. But if you mean incrementing the `i` declared on the outside scope of the callback, yes it is, see [here](https://gobyexample.com/closures).

core/chainio/avs_writer.go
MarcosNicolau: I just liked that it was all in one place since we already have two layers of `SendResponseTransacion`, adding another would make the whole flow uncomfortable to read IMO. But I will move it to another function if you think that's better.

core/chainio/avs_writer.go
MarcosNicolau: Actually, it is just a call to `eth_gasPrice`.

core/chainio/avs_writer.go
MarcosNicolau: RPC errors are marked as permanent to not fall into an infinite retry. Waiting for the receipt and not getting is always a recoverable error as we are bumping the fee, at some point our tx should be accepted. That is why this function is wrapped in another retry that has a max which would account for rpc errors.

core/chainio/avs_writer.go
MarcosNicolau: Reference: https://pkg.go.dev/context

> Even though ctx will be expired, it is good practice to call its cancellation function in any case. Failure to do so may keep the context and its parent alive longer than necessary.

core/chainio/avs_writer.go
MarcosNicolau: No, because it is not being immediately called, it is passed as a callback to the retry function.

core/connection.go
MarcosNicolau: It is not necessary as errors without `Permanent` will automatically be retryable. This is just a nice wrapper to the `backoff` library we are using.

core/utils/eth_client_utils.go
MarcosNicolau: This err means that: the there was an err with the RPC server (if it is down for example) or the receipt was not found (not yet mined). Either way, we want to retry. So we don't need to perform any extra check over the err that is why we ignore it.

grafana/provisioning/dashboards/aligned/aggregator_batcher.json
```diff
@@ -1323,6 +1423,6 @@
   "timezone": "browser",
   "title": "Aggregator Data",
   "uid": "aggregator",
-  "version": 6,
+  "version": 15,
   "weekStart": ""
-}
+}
```
MarcosNicolau: Missing EOL here

core/chainio/avs_writer.go
MarcosNicolau: Wouldn't that mean adding too much complex logic into something that will hardly ever happen? Why don't we just let it fail and let the outside function retry again, given that it also retrievable with a max retry?

core/chainio/avs_writer.go
MarcosNicolau: See [here](https://github.com/yetanotherco/aligned_layer/blob/f0e58cf1545c566d0b4e61af1167561656bdf8b5/aggregator/internal/pkg/aggregator.go#L272-L286).

core/chainio/avs_writer.go
MarcosNicolau: Exactly it is a timeout that we'll wait to get the receipt. If we don't find it after that, we will bump the fee. I'll make sure to move to a constant!

core/connection.go
MarcosNicolau: The `ok` doesn't handle that already? If a type assertion fails then go sets `ok` to false. 

reference: [https://go.dev/tour/methods/15](https://go.dev/tour/methods/15)

MarcosNicolau: In fact, `err != nil` is redundant as `ok` already ensures that err is non-nil.

MarcosNicolau:
@JuArce I left some more explicit commands on how to run the test for this pr!

core/chainio/avs_writer.go
```diff
@@ -70,10 +81,18 @@ func NewAvsWriterFromConfig(baseConfig *config.BaseConfig, ecdsaConfig *config.E
 	}, nil
 }
 
-func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMerkleRoot [32]byte, senderAddress [20]byte, nonSignerStakesAndSignature servicemanager.IBLSSignatureCheckerNonSignerStakesAndSignature) (*common.Hash, error) {
+// Sends AggregatedResponse and waits for the receipt for three blocks, if not received
+// it will try again bumping the last tx gas price based on `CalculateGasPriceBump`
+// This process happens indefinitely until the transaction is included.
+//
+// Note: If the rpc endpoints fail, the retry mechanism stops, returning a permanent error.
```
uri-99: If returned the permanent error, the aggregated response is lost? this would not be good

core/chainio/avs_writer.go
uri-99: Oh, i understand. It would be tidy to add a comment of this.
Also now it makes sense to [ignore this comment](https://github.com/yetanotherco/aligned_layer/pull/1286/files#r1816960961)

uri-99: also change the variable name to be more clear it is a function and not a value

core/chainio/avs_writer.go
uri-99: i understand now this function is being build and not created. see my other comment about suggestions on this

core/chainio/avs_writer.go
```diff
@@ -87,21 +106,60 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
+		}
+		i++
+
+		// get gas price
+		gasPrice, err := w.Client.SuggestGasPrice(context.Background())
+		if err != nil {
+			// Retry with fallback
+			gasPrice, err = w.ClientFallback.SuggestGasPrice(context.Background())
+			if err != nil {
+				return nil, connection.PermanentError{Inner: err}
```
uri-99: do permanent errors revert the sending of the task response?

And it is not really a permanent error, as some more retries may solve if the error is both rpcs are temporarily down.

core/utils/eth_client_utils.go
```diff
@@ -42,3 +46,23 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
 	}
 	return quorumThresholdPercentages
 }
+
+// Simple algorithm to calculate the gasPrice bump based on:
+// the currentGasPrice, a base bump percentage, a retry percentage, and the retry count.
+// Formula: currentGasPrice + (currentGasPrice * (baseBumpPercentage + retryCount * incrementalRetryPercentage) / 100)
```
uri-99, line 54: much better 👍

core/utils/eth_client_utils.go
```diff
@@ -42,3 +46,23 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
...
+
+	// Total bump percentage: base bump + incremental retry percentage
+	totalBumpPercentage := new(big.Int).Add(big.NewInt(int64(baseBumpPercentage)), incrementalRetryPercentage)
+
+	// Calculate the bump amount: currentGasPrice * totalBumpPercentage / 100
+	bumpAmount := new(big.Int).Mul(currentGasPrice, totalBumpPercentage)
+	bumpAmount = new(big.Int).Div(bumpAmount, big.NewInt(100))
+
+	// Final bumped gas price: currentGasPrice + bumpAmount
+	bumpedGasPrice := new(big.Int).Add(currentGasPrice, bumpAmount)
```
uri-99: yes this is now understandable. Good improvement 👍

core/utils/eth_client_utils.go
uri-99: yes but if we have an err here, we can trigger the retry now and avoid the following code.

Mainly because im not sure if err can be != nil and receipt != nil at the same time.

if you have an err, and some extra info in receipt, you will be returning the extra error info, an invalid receipt

core/utils/eth_client_utils.go
```diff
@@ -17,12 +18,15 @@ const sleepTime = 5 * time.Second
...
-			time.Sleep(sleepTime)
-		} else {
+		receipt, _ := client.TransactionReceipt(ctx, txHash)
+		if receipt != nil {
 			return receipt, nil
 		}
+		// if context has timed out, return
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
```
uri-99: Missing a comment about how the context timeout on the other function triggers this error.
I see the comment you put on line 25, i would make it more verbose.
This is because the maxRetries with the RetryTimeout gives the idea they are more relevant

core/chainio/avs_writer.go
uri-99: As my pr review comment says, it would be cleaner to remove the outside function retries, so the loop is executed in only one place, and not 2 anidated loops that add no real value

core/chainio/avs_writer.go
uri-99: So as we talked, we should have the "normal error" (a simple return err) , and errors that don't bump the fee; instead of permanent errors which give place to confusion (more so when the outside loop is removed)

core/connection.go
uri-99: ditto. What would be a good name for an error that doesn't require gas bump?

core/connection.go
```diff
@@ -0,0 +1,76 @@
...
+}
+func (e PermanentError) Is(err error) bool {
+	_, ok := err.(PermanentError)
+	return ok
+}
+
+// Retries a given function in an exponential backoff manner and returns a value upon correct execution.
+// It will retry calling the function while it returns a non permanent error, until the max retries.
+// If maxTries == 0 then the retry function will run indefinitely until success or until a `PermanentError` is returned.
+func RetryWithData[T any](functionToRetry func() (*T, error), minDelay uint64, factor float64, maxTries uint64) (*T, error) {
```
uri-99: is it called "withdata" because it returns data? or data is related to some sort of parameter? a more clear name would be better

uri-99 (changes requested):
I would remove the retries made in here, and have the underlying sendAggregatedResponse handle the retries.
This would need to be paired with the PermanentErrors to be a bit different. They should not abort the cycle, but retry it without increasing the bump %.

core/chainio/avs_writer.go
```diff
@@ -70,30 +73,77 @@ func NewAvsWriterFromConfig(baseConfig *config.BaseConfig, ecdsaConfig *config.E
...
+		}
+		bumpedGasPrice := utils.CalculateGasPriceBumpBasedOnRetry(gasPrice, gasBumpPercentage, gasBumpIncrementalPercentage, i)
+		// new bumped gas price must be higher than the last one (this should hardly ever happen though)
+		if gasPrice.Cmp(txOpts.GasPrice) > 0 {
+			txOpts.GasPrice = bumpedGasPrice
+		} else {
+			// bump the last price 10% to replace it.
+			bumpAmount := new(big.Int).Mul(txOpts.GasPrice, big.NewInt(10))
+			bumpAmount = new(big.Int).Div(bumpAmount, big.NewInt(100))
+			txOpts.GasPrice = new(big.Int).Mul(txOpts.GasPrice, bumpAmount)
```
JulianVentura: This should be an `Add` instead of a `Mul`

core/chainio/avs_writer.go
```diff
@@ -70,30 +73,77 @@ func NewAvsWriterFromConfig(baseConfig *config.BaseConfig, ecdsaConfig *config.E
...
+
 	tx, err := w.RespondToTaskV2Retryable(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
 	if err != nil {
 		return nil, err
 	}
 
 	err = w.checkRespondToTaskFeeLimit(tx, txOpts, batchIdentifierHash, senderAddress)
 	if err != nil {
-		return nil, err
+		return nil, retry.PermanentError{Inner: err}
```
uri-99: https://github.com/yetanotherco/aligned_layer/issues/1389

core/chainio/avs_writer.go
uri-99: you can call `CalculateGasPriceBumpBasedOnRetry` with 0 as params for this 10% bump

core/chainio/avs_writer.go
```diff
@@ -70,30 +73,77 @@ func NewAvsWriterFromConfig(baseConfig *config.BaseConfig, ecdsaConfig *config.E
...
+		if gasPrice.Cmp(txOpts.GasPrice) > 0 {
+			txOpts.GasPrice = bumpedGasPrice
+		} else {
+			// bump the last price 10% to replace it.
+			bumpAmount := new(big.Int).Mul(txOpts.GasPrice, big.NewInt(10))
+			bumpAmount = new(big.Int).Div(bumpAmount, big.NewInt(100))
+			txOpts.GasPrice = new(big.Int).Mul(txOpts.GasPrice, bumpAmount)
+		}
+		if i > 0 {
+			onRetry(txOpts.GasPrice)
```
uri-99: i would change to `gasBumped()` and add a comment here it is for telemetrics

core/chainio/avs_writer.go
```diff
@@ -70,30 +73,77 @@ func NewAvsWriterFromConfig(baseConfig *config.BaseConfig, ecdsaConfig *config.E
...
+		}
+
+		err = w.checkRespondToTaskFeeLimit(tx, txOpts, batchIdentifierHash, senderAddress)
+		if err != nil {
+			// We bump the fee so much that the transaction cost is more expensive than the batcher fee limit
+			return nil, retry.PermanentError{Inner: err}
```
uri-99, line 119: https://github.com/yetanotherco/aligned_layer/issues/1389

core/chainio/avs_writer.go
```diff
@@ -70,30 +73,77 @@ func NewAvsWriterFromConfig(baseConfig *config.BaseConfig, ecdsaConfig *config.E
...
+		if err != nil {
+			return nil, err
+		}
+		return nil, fmt.Errorf("transaction failed")
 	}
 
-	txHash := tx.Hash()
-
-	return &txHash, nil
+	return retry.RetryWithData(respondToTaskV2Func, 1000, 2, 0, 60, retry.MaxElapsedTime)
```
uri-99: `RetryWithData` function name is misleading, sounds like `Data` is actually input and not output

uri-99 (changes requested):

uri-99 (changes requested):
Also blocking until base branch is merged before this one

core/chainio/avs_writer.go
```diff
@@ -70,30 +73,77 @@ func NewAvsWriterFromConfig(baseConfig *config.BaseConfig, ecdsaConfig *config.E
...
-	txOpts.GasLimit = tx.Gas() * 110 / 100 // Add 10% to the gas limit
-	tx, err = w.RespondToTaskV2Retryable(&txOpts, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
-	if err != nil {
-		return nil, err
+	i := 0
+
+	respondToTaskV2Func := func() (*types.Receipt, error) {
+		// We bump only when the timeout for waiting for the receipt has passed
+		// not when an rpc failed
+		gasPrice, err := w.ClientFallback.SuggestGasPrice(context.Background())
```
JulianVentura: Is there a reason here to use the fallback?

JulianVentura (commented):
LGTM, I'll test later.

core/chainio/avs_writer.go
MarcosNicolau: I agree, but this sounds like something that should be done in #1304. I don't mind refactoring it thou.

core/chainio/avs_writer.go
MarcosNicolau: Solved in [daa8f40](https://github.com/yetanotherco/aligned_layer/pull/1286/commits/daa8f400f371a67b43120a692f9161e7edab3907).

JulianVentura (approved):
Looks good and works on my machine

core/utils/eth_client_utils.go
```diff
@@ -10,11 +12,14 @@ import (
...
-func WaitForTransactionReceipt(client eth.InstrumentedClient, ctx context.Context, txHash gethcommon.Hash) (*types.Receipt, error) {
+// WaitForTransactionReceiptRetryable repeatedly attempts to fetch the transaction receipt for a given transaction hash.
+// If the receipt is not found, the function will retry with exponential backoff until the specified `waitTimeout` duration is reached.
+// If the receipt is still unavailable after `waitTimeout`, it will return an error.
+func WaitForTransactionReceiptRetryable(client eth.InstrumentedClient, ctx context.Context, txHash gethcommon.Hash, waitTimeout time.Duration) (*types.Receipt, error) {
 	receipt_func := func() (*types.Receipt, error) {
 		return client.TransactionReceipt(ctx, txHash)
 	}
-	return retry.RetryWithData(receipt_func, retry.MinDelay, retry.RetryFactor, retry.NumRetries, retry.MaxInterval, retry.MaxElapsedTime)
+	return retry.RetryWithData(receipt_func, retry.MinDelay, retry.RetryFactor, 0, retry.MaxInterval, waitTimeout)
```
avilagaston9: We should adjust these parameters. With the current configuration, we are retrying at intervals of `1 sec - 2 secs - 4 secs - 8 secs - 16 secs` and then stop retrying when the `waitTimeout` is reached. If we start retrying at the beginning of a block, we won't be able to retry in the third block due to the exponential backoff. I suggest using a `MaxInterval` of 2 secs (to retry every 2 seconds) and increment the `waitTimeout` a bit to ensure we can retry in the third block as well.

MarcosNicolau: You are right, good one!

uri-99:
There is something odd with the bumper, i added the lines of code in the description to make it fail, it bumped correctly but then skyrocketed on because it had no delay?
image

core/utils/eth_client_utils.go
MarcosNicolau: Solved in [475125d](https://github.com/yetanotherco/aligned_layer/pull/1286/commits/475125db9e982b3e220a055c0ffe29b3d76dc8db).

MarcosNicolau:
> There is something odd with the bumper, i added the lines of code in the description to make it fail, it bumped correctly but then skyrocketed on because it had no delay?

I tried to reproduce the behavior but I couldn't.  Would you try again with the latest changes pls.

uri-99 (approved):

core/chainio/avs_writer.go
```diff
@@ -70,7 +72,10 @@ func NewAvsWriterFromConfig(baseConfig *config.BaseConfig, ecdsaConfig *config.E
 	}, nil
 }
 
-func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMerkleRoot [32]byte, senderAddress [20]byte, nonSignerStakesAndSignature servicemanager.IBLSSignatureCheckerNonSignerStakesAndSignature) (*common.Hash, error) {
+// Sends AggregatedResponse and waits for the receipt for three blocks, if not received
+// it will try again bumping the last tx gas price based on `CalculateGasPriceBump`
+// This process happens indefinitely until the transaction is included.
```
avilagaston9, line 77: Maybe for a separate PR, but we need to add some numbers to graph the retry logic. See: https://github.com/yetanotherco/aligned_layer/blob/b03a625a2fc755459797d013297de83793521a48/batcher/aligned-batcher/src/lib.rs#L1422

core/utils/eth_client_utils.go
```diff
@@ -32,3 +40,38 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
...
+	bumpAmount := new(big.Int).Mul(currentGasPrice, totalBumpPercentage)
+	bumpAmount = new(big.Int).Div(bumpAmount, big.NewInt(100))
+
+	// Final bumped gas price: currentGasPrice + bumpAmount
+	bumpedGasPrice := new(big.Int).Add(currentGasPrice, bumpAmount)
+
+	return bumpedGasPrice
+}
+
+func GetGasPriceRetryable(client eth.InstrumentedClient, ctx context.Context) (*big.Int, error) {
```
JuArce: This function uses the same client twice. Shouldn't use the fallback client on the second call?

JuArce: Also, ctx is not being used

JuArce:
```suggestion
/*
GetGasPriceRetryable
Get the gas price from the client with retry logic.
- All errors are considered Transient Errors
- Retry times: 1 sec, 2 sec, 4 sec
*/
func GetGasPriceRetryable(client eth.InstrumentedClient, ctx context.Context) (*big.Int, error) {
```

aggregator/pkg/aggregator.go
```diff
@@ -300,7 +301,11 @@ func (agg *Aggregator) sendAggregatedResponse(batchIdentifierHash [32]byte, batc
 		"senderAddress", hex.EncodeToString(senderAddress[:]),
 		"batchIdentifierHash", hex.EncodeToString(batchIdentifierHash[:]))
 
-	txHash, err := agg.avsWriter.SendAggregatedResponse(batchIdentifierHash, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature)
+	onGasPriceBumped := func(bumpedGasPrice *big.Int) {
```
JuArce, line 305:
```suggestion
	// This function is a callback that is called when the gas price is bumped on the avsWriter.SendAggregatedResponse
	onGasPriceBumped := func(bumpedGasPrice *big.Int) {
```

aggregator/pkg/aggregator.go
```diff
@@ -300,7 +301,11 @@ func (agg *Aggregator) sendAggregatedResponse(batchIdentifierHash [32]byte, batc
...
+		agg.metrics.IncBumpedGasPriceForAggregatedResponse()
+		agg.telemetry.BumpedTaskGasPrice(batchMerkleRoot, bumpedGasPrice.String())
+	}
+	receipt, err := agg.avsWriter.SendAggregatedResponse(batchIdentifierHash, batchMerkleRoot, senderAddress, nonSignerStakesAndSignature, agg.AggregatorConfig.Aggregator.GasBaseBumpPercentage, agg.AggregatorConfig.Aggregator.GasBumpIncrementalPercentage, agg.AggregatorConfig.Aggregator.TimeToWaitBeforeBump, onGasPriceBumped)
```
JuArce:
```suggestion
	receipt, err := agg.avsWriter.SendAggregatedResponse(
		batchIdentifierHash,
		batchMerkleRoot,
		senderAddress,
		nonSignerStakesAndSignature,
		agg.AggregatorConfig.Aggregator.GasBaseBumpPercentage,
		agg.AggregatorConfig.Aggregator.GasBumpIncrementalPercentage,
		agg.AggregatorConfig.Aggregator.TimeToWaitBeforeBump,
		onGasPriceBumped,
	)
```

JuArce (changes requested):

core/utils/eth_client_utils.go
MarcosNicolau: yeah, good catch

MarcosNicolau: I addressed your comments at [5b1219a](https://github.com/yetanotherco/aligned_layer/pull/1286/commits/5b1219a79927bf1c66394d9eed4a52cff2259ddf).

core/utils/eth_client_utils.go
```diff
@@ -32,3 +40,38 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
 	}
 	return quorumThresholdPercentages
 }
+
+// Simple algorithm to calculate the gasPrice bump based on:
+// the currentGasPrice, a base bump percentage, a retry percentage, and the retry count.
+// Formula: currentGasPrice + (currentGasPrice * (baseBumpPercentage + retryCount * incrementalRetryPercentage) / 100)
+func CalculateGasPriceBumpBasedOnRetry(currentGasPrice *big.Int, baseBumpPercentage uint, retryAttemptPercentage uint, retryCount int) *big.Int {
+	// Incremental percentage increase for each retry attempt (i*5%)
```
avilagaston9:
```suggestion
	// Incremental percentage increase for each retry attempt (i*2%)
```

core/chainio/avs_writer.go
```diff
@@ -83,17 +88,61 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
+			return nil, err
+		}
+
+		bumpedGasPrice := utils.CalculateGasPriceBumpBasedOnRetry(gasPrice, gasBumpPercentage, gasBumpIncrementalPercentage, i)
+		// new bumped gas price must be higher than the last one (this should hardly ever happen though)
+		if bumpedGasPrice.Cmp(txOpts.GasPrice) > 0 {
+			txOpts.GasPrice = bumpedGasPrice
+		} else {
+			// bump the last tx gas price a little by `gasBumpIncrementalPercentage` to replace it.
+			txOpts.GasPrice = utils.CalculateGasPriceBumpBasedOnRetry(txOpts.GasPrice, gasBumpIncrementalPercentage, 0, 0)
```
avilagaston9, line 110: We should add comments explaining how we are using this function. We are using different bump percentages depending on which price that we are bumping.

avilagaston9 (approved):
LGTM! Left some nit comments

JuArce (approved):

PR #1304, opened 2024-10-22
refactor(aggregator): Aggregator retries

PatStiles:
# Adds retry functions to the Aggregator

## Description

Add retries to rpc calls and connections within the Aggregator:
- Refactors `RetryWithData` to add a panic catch and use generic types.
- Creates wrapper functions for code that await fallible connections using retry functions.
- Adds tests using Anvil as a local testnet that check the returning error case for a call that successfully connects and executes (but does return the correct value), then kills anvil and asserts the failing transient error is returned.

### Notes:
- Exhaustively finding and distinguishing all Permanent and Transient errors still needs to be done. Issue: https://github.com/yetanotherco/aligned_layer/issues/1390
- Test for `InitializeNewTaskRetryable` and `ProcessNewSignatureRetryable` and commented out. This is due to an issue with the eigen-sdk where it panics if it loses connection. There is an issue here: https://github.com/yetanotherco/aligned_layer/issues/1361
- The unit tests for the retries require coordination are somewhat flaky and could be approved in a separate pr by separating the retryable logic into a separate function and testing it independent of retry library function: https://github.com/yetanotherco/aligned_layer/pull/1392

### To Test

- Run a local network and verify things work
- Run the unit test `cd core && go test -v`

## Type of change

- [x] Refactor

## Checklist
- [x] Unit tests added


aggregator/pkg/aggregator.go
```diff
@@ -269,17 +270,13 @@ func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsA
...
-		time.Sleep(2 * time.Second)
+	_, err = agg.sendAggregatedResponse(batchIdentifierHash, batchData.BatchMerkleRoot, batchData.SenderAddress, nonSignerStakesAndSignature)
+	if err == nil {
+		agg.logger.Info("Aggregator successfully responded to task",
+			"taskIndex", blsAggServiceResp.TaskIndex,
+			"batchIdentifierHash", "0x"+hex.EncodeToString(batchIdentifierHash[:]))
+		return
 	}
 
 	agg.logger.Error("Aggregator failed to respond to task, this batch will be lost",
```
Oppen, line 284: This log would no longer be valid if we're handling retries in the call-site. That should only be logged (at least with the "batch will be lost" part) after a permanent error or exhausting retries.

core/connection.go
```diff
@@ -0,0 +1,114 @@
...
+			val, err = functionToRetry()
+			i++
+			if perm, ok := err.(PermanentError); err != nil && ok {
+				err = backoff.Permanent(perm.Inner)
+			}
+		}()
+		return val, err
+	}
+
+	randomOption := backoff.WithRandomizationFactor(0)
```
JulianVentura: I think we should add some jitter here. See [this article](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/)

core/connection.go
```diff
@@ -0,0 +1,119 @@
...
+	return ok
+}
+
+const MinDelay = 1000
+const RetryFactor = 2
+const NumRetries = 3
+
+// Same as Retry only that the functionToRetry can return a value upon correct execution
+func RetryWithData[T any](functionToRetry func() (T, error), minDelay uint64, factor float64, maxTries uint64) (T, error) {
+	i := 0
```
JulianVentura: Is this `i` variable being used for something? Same goes for `Retry` function

core/connection.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Retries a given function in an exponential backoff manner.
+// It will retry calling the function while it returns an error, until the max retries.
+// If maxTries == 0 then the retry function will run indefinitely until success
+// from the configuration are reached, or until a `PermanentError` is returned.
+// The function to be retried should return `PermanentError` when the condition for stop retrying
+// is met.
+func Retry(functionToRetry func() error, minDelay uint64, factor float64, maxTries uint64) error {
+	i := 0
+	f := func() error {
```
JulianVentura: It took me a while to understand why this function wrap is necessary. We have two `PermanentError` here, right?
The function we receive may return a `PermanentError` (ours), but the library we are using is expecting to return a `backoff.Permanent`. So we are basically transforming one into the other right here.
I think we should have an auxiliary function which converts the error to a `backoff.Permanent`, so the code is easier to read and we don't repeat that same logic in both `Retry` and `RetryWithData` functions.
I imagine the function to be:

```golang
func convertErrorToLibraryRetryError(err error) error {
	if perm, ok := err.(PermanentError); err != nil && ok {
		return backoff.Permanent(perm.Inner)
	}
        return err
}
```

core/connection.go
```diff
@@ -0,0 +1,119 @@
...
+			}
+		}()
+		return val, err
+	}
+
+	randomOption := backoff.WithRandomizationFactor(0)
+
+	initialRetryOption := backoff.WithInitialInterval(time.Millisecond * time.Duration(minDelay))
+	multiplierOption := backoff.WithMultiplier(factor)
+	expBackoff := backoff.NewExponentialBackOff(randomOption, multiplierOption, initialRetryOption)
```
MarcosNicolau: I would add a `MaxInterval` here, maybe to something like `60s`.

core/connection.go
```diff
@@ -0,0 +1,119 @@
...
+// Same as Retry only that the functionToRetry can return a value upon correct execution
+func RetryWithData[T any](functionToRetry func() (T, error), minDelay uint64, factor float64, maxTries uint64) (T, error) {
+	i := 0
+	f := func() (T, error) {
+		var (
+			val T
+			err error
+		)
+		func() {
+			defer func() {
```
JulianVentura: Don't we need this logic in the `Retry` function too? I think that callback function may panic too. I would write an auxiliary function to handle the recover for both of them.

core/connection.go
```diff
@@ -0,0 +1,119 @@
...
+
+type PermanentError struct {
+	Inner error
+}
+
+func (e PermanentError) Error() string { return e.Inner.Error() }
+func (e PermanentError) Unwrap() error {
+	return e.Inner
+}
+func (e PermanentError) Is(err error) bool {
```
MarcosNicolau: Is this function used anywhere? The name is confusing, maybe I am dumb, but I am having a hard time understanding what it does. What is it supposed to do?

core/connection.go
```diff
@@ -0,0 +1,119 @@
...
+
+type TransientError struct {
+	Inner error
+}
+
+func (e TransientError) Error() string { return e.Inner.Error() }
+func (e TransientError) Unwrap() error {
+	return e.Inner
+}
+func (e TransientError) Is(err error) bool {
```
MarcosNicolau: ditto.

core/connection.go
```diff
@@ -0,0 +1,119 @@
...
+func RetryWithData[T any](functionToRetry func() (T, error), minDelay uint64, factor float64, maxTries uint64) (T, error) {
+	i := 0
+	f := func() (T, error) {
+		var (
+			val T
+			err error
+		)
+		func() {
+			defer func() {
+				if r := recover(); r != nil {
```
MarcosNicolau: Do we need this?

core/chainio/avs_writer.go
```diff
@@ -166,18 +153,152 @@ func (w *AvsWriter) compareAggregatorBalance(amount *big.Int, aggregatorAddress
...
+}
+
+func (w *AvsWriter) BalanceAtRetryable(ctx context.Context, aggregatorAddress common.Address, blockNumber *big.Int) (*big.Int, error) {
+	var (
+		aggregatorBalance *big.Int
+		err               error
+	)
+	balanceAt_func := func() (*big.Int, error) {
+		aggregatorBalance, err = w.Client.BalanceAt(ctx, aggregatorAddress, blockNumber)
+		//aggregatorBalance, err := connection.RetryWithData(balanceAt_func, connection.MinDelay, connection.RetryFactor, connection.NumRetries)
```
JulianVentura: There is some commented code right here

core/connection.go
PatStiles: @Oppen any context regarding these implementations?. The original PR I branched off of included these functions for the Permanent error type. I implemented them following for the Transient error as well.

core/connection.go
```diff
@@ -0,0 +1,119 @@
...
+// It will retry calling the function while it returns an error, until the max retries.
+// If maxTries == 0 then the retry function will run indefinitely until success
+// from the configuration are reached, or until a `PermanentError` is returned.
+// The function to be retried should return `PermanentError` when the condition for stop retrying
+// is met.
+func Retry(functionToRetry func() error, minDelay uint64, factor float64, maxTries uint64) error {
+	i := 0
+	f := func() error {
+		err := functionToRetry()
+		i++
```
MarcosNicolau: If we do need the panic handling, then we need to add it here as well.

core/connection.go
PatStiles: I added this as a means of catching a panic with one of the contract bindings functions that I ended up fixing so it did not panic (not the one that is in the process of being fixed). @Oppen should we still include a panic catch here?

core/utils/eth_client_utils.go
```diff
@@ -2,29 +2,41 @@ package utils
...
+			if err.Error() == "not found" {
+				return nil, connection.TransientError{Inner: err}
+			}
+			if strings.Contains(err.Error(), "connect: connection refused") {
+				return nil, connection.TransientError{Inner: err}
+			}
+			if strings.Contains(err.Error(), "read: connection reset by peer") {
+				return nil, connection.TransientError{Inner: err}
+			}
+			return nil, connection.PermanentError{Inner: fmt.Errorf("Permanent error: Unexpected Error while retrying: %s\n", err)}
```
MarcosNicolau: If we have a maxRetry, I think we should not throw permanent errors. Unless the transaction reverts.

MarcosNicolau: Actually, never mind, this will be handled in #1286.

core/chainio/avs_writer.go
```diff
@@ -166,18 +153,152 @@ func (w *AvsWriter) compareAggregatorBalance(amount *big.Int, aggregatorAddress
...
+				}
+				if strings.Contains(err.Error(), "connect: connection refused") {
+					err = connection.TransientError{Inner: err}
+					return state, err
+				}
+				if strings.Contains(err.Error(), "read: connection reset by peer") {
+					return state, connection.TransientError{Inner: err}
+				}
+				err = connection.PermanentError{Inner: fmt.Errorf("Permanent error: Unexpected Error while retrying: %s\n", err)}
+			}
```
JulianVentura: I'm noticing a common pattern. We are doing the following steps in all these retrying functions:
1. Try with the main contract
2. If error, try with the fallback contract.
3. If error, match over "not found", "connection refused" or "connection reset by peer" errors converting them to `TransientError`
4. If no match, convert the error to `PermanentError`

This logic is repeated in those functions, we should make an auxiliary function for it. I think we could use `WaitForTransactionReceiptRetryable` util function, right?

core/connection.go
PatStiles: Addressed!

JulianVentura (changes requested):
Left some comments, I'll test later.

core/connection.go
PatStiles: addressed

core/connection.go
PatStiles: addressed!

core/connection_test.go
```diff
@@ -0,0 +1,928 @@
...
+	}
+
+	// Run the command
+	err = cmd.Start()
+	if err != nil {
+		fmt.Printf("Error: %s\n", err)
+	}
+
+	// Delay needed for anvil to start
+	time.Sleep(750 * time.Millisecond)
```
JulianVentura: This may introduce some errors and flashy tests. Instead we could query anvil until it's alive, like with the following:
`cast rpc --rpc-url http://localhost:8545 eth_blockNumber`

core/connection.go
PatStiles: I think we should leave the code inlined as is as the section itself is dense. I agree that the conversion is not intuitive and left as clarifying comment (as discussed).

core/connection.go
PatStiles: I think we should leave the code inlined as is as the code section itself is quite dense.

core/chainio/avs_writer.go
PatStiles: Addressed!

core/connection_test.go
```diff
@@ -0,0 +1,929 @@
...
+	"github.com/stretchr/testify/assert"
+	servicemanager "github.com/yetanotherco/aligned_layer/contracts/bindings/AlignedLayerServiceManager"
+	connection "github.com/yetanotherco/aligned_layer/core"
+	"github.com/yetanotherco/aligned_layer/core/chainio"
+	"github.com/yetanotherco/aligned_layer/core/config"
+	"github.com/yetanotherco/aligned_layer/core/utils"
+)
+
+func DummyFunction(x uint64) (uint64, error) {
+	fmt.Println("Starting Anvil on Port ")
```
JulianVentura: I think we should delete this log

core/connection_test.go
```diff
@@ -0,0 +1,929 @@
...
+func TestRetryWithData(t *testing.T) {
+	function := func() (*uint64, error) {
+		x, err := DummyFunction(43)
+		return &x, err
+	}
+	data, err := connection.RetryWithData(function, 1000, 2, 3, connection.MaxInterval)
+	if err != nil {
+		t.Errorf("Retry error!: %s", err)
+	} else {
+		fmt.Printf("DATA: %d\n", data)
```
JulianVentura: I don't think we need to log data here, I would make the tests pass without logs and only log on error cases

aggregator/pkg/aggregator.go
```diff
@@ -387,6 +383,16 @@ func (agg *Aggregator) AddNewTask(batchMerkleRoot [32]byte, senderAddress [20]by
 	agg.logger.Info("New task added", "batchIndex", batchIndex, "batchIdentifierHash", "0x"+hex.EncodeToString(batchIdentifierHash[:]))
 }
 
+// |---RETRYABLE---|
+
+func (agg *Aggregator) InitializeNewTaskRetryable(batchIndex uint32, taskCreatedBlock uint32, quorumNums eigentypes.QuorumNums, quorumThresholdPercentages eigentypes.QuorumThresholdPercentages, timeToExpiry time.Duration) error {
+	initilizeNewTask_func := func() error {
+		// Returns an error if the
```
avilagaston9: Incomplete comment

JulianVentura (commented):
Tests are failing on my machine

aggregator/pkg/aggregator.go
```diff
@@ -387,6 +383,16 @@ func (agg *Aggregator) AddNewTask(batchMerkleRoot [32]byte, senderAddress [20]by
...
+		return agg.blsAggregationService.InitializeNewTask(batchIndex, taskCreatedBlock, quorumNums, quorumThresholdPercentages, 100*time.Second)
```
avilagaston9: Shouldn't we need to convert to retry errors here as well?

aggregator/pkg/server.go
```diff
@@ -121,3 +127,35 @@ func (agg *Aggregator) ServerRunning(_ *struct{}, reply *int64) error {
 	*reply = 1
 	return nil
 }
+
+// |---RETRYABLE---|
+
+// Error throw is ______
+func (agg *Aggregator) ProcessNewSignatureRetryable(ctx context.Context, taskIndex uint32, taskResponse interface{}, blsSignature *bls.Signature, operatorId eigentypes.Bytes32) error {
```
avilagaston9: I would keep the original names of the functions (`ProcessNewSignature` in this case) since the retryable functions are the inner ones (`processNewSignature_func`).

aggregator/pkg/server.go
```diff
@@ -121,3 +127,35 @@ func (agg *Aggregator) ServerRunning(_ *struct{}, reply *int64) error {
 	*reply = 1
 	return nil
 }
+
+// |---RETRYABLE---|
+
+// Error throw is ______
```
uri-99: ?

aggregator/pkg/server.go
```diff
@@ -121,3 +127,35 @@ func (agg *Aggregator) ServerRunning(_ *struct{}, reply *int64) error {
...
+func (agg *Aggregator) ProcessNewSignatureRetryable(ctx context.Context, taskIndex uint32, taskResponse interface{}, blsSignature *bls.Signature, operatorId eigentypes.Bytes32) error {
+	var err error
+	processNewSignature_func := func() error {
+		err = agg.blsAggregationService.ProcessNewSignature(
+			ctx, taskIndex, taskResponse,
+			blsSignature, operatorId,
+		)
+		if err != nil {
+			// Note return type will be nil
```
uri-99: i dont understand this comment

config-files/config-aggregator-test.yaml
avilagaston9, line 1: Do we want to push this one?

aggregator/pkg/server.go
```diff
@@ -121,3 +127,35 @@ func (agg *Aggregator) ServerRunning(_ *struct{}, reply *int64) error {
...
+			if err.Error() == "not found" {
+				err = connection.TransientError{Inner: err}
+				return err
+			}
```
uri-99: Operator not founds sounds like a permanent error

core/chainio/avs_subscriber.go
```diff
@@ -67,17 +71,20 @@ func (s *AvsSubscriber) SubscribeToNewTasksV2(newTaskCreatedChan chan *servicema
...
+		s.logger.Error("Failed to subscribe to new AlignedLayer tasks after %d retries", connection.NumRetries, "err", err)
 		return nil, err
 	}
+	s.logger.Info("Subscribed to new AlignedLayer V2 tasks")
 
-	subFallback, err := subscribeToNewTasksV2(s.AvsContractBindings.ServiceManagerFallback, internalChannel, s.logger)
+	subFallback, err := SubscribeToNewTasksV2Retrayable(s.AvsContractBindings.ServiceManagerFallback, internalChannel, s.logger)
 	if err != nil {
-		s.logger.Error("Failed to subscribe to new AlignedLayer tasks", "err", err)
+		s.logger.Error("Failed to subscribe to new AlignedLayer tasks %d retries", connection.NumRetries, "err", err)
```
avilagaston9: Same here. I would hide all the retry logic inside `subscribeToNewTasksV2` and leave this untouched. But maybe it's a matter of taste.

aggregator/pkg/server.go
```diff
@@ -121,3 +127,35 @@ func (agg *Aggregator) ServerRunning(_ *struct{}, reply *int64) error {
...
+			if strings.Contains(err.Error(), "connect: connection refused") {
+				err = connection.TransientError{Inner: err}
+				return err
+			}
+			if strings.Contains(err.Error(), "read: connection reset by peer") {
+				err = connection.TransientError{Inner: err}
+				return err
+			}
+			err = connection.PermanentError{Inner: fmt.Errorf("Permanent error: Unexpected Error while retrying: %s\n", err)}
```
uri-99: I don't think matching every other error case as Permanent is prudent

core/connection_test.go
PatStiles: addressed

core/connection_test.go
PatStiles: addressed

core/chainio/avs_subscriber.go
```diff
@@ -67,17 +71,20 @@ func (s *AvsSubscriber) SubscribeToNewTasksV2(newTaskCreatedChan chan *servicema
...
 		return nil, err
 	}
+	s.logger.Info("Subscribed to new AlignedLayer V2 tasks")
```
avilagaston9, line 80: Is this duplicated? I think we should log only once at the end of the two subscriptions

core/chainio/avs_subscriber.go
avilagaston9: Same with `subscribeToNewTasksV3`

core/chainio/avs_subscriber.go
avilagaston9: Same with `subscribeToNewTasksV3`

core/chainio/avs_writer.go
avilagaston9: Maybe an auxiliary function just to handle the received errors might work

config-files/config-aggregator-test.yaml
uri-99, line 1: maybe remove commented values as there are a lot of them

core/chainio/avs_writer.go
MauroToscano: Let's not try to abstract this, since it may not be exactly the same always.

But use a case statemente, and group errors of the same type in the same cas:

```if err != nil {
    switch {
    case strings.Contains(err.Error(), "not found"),
         strings.Contains(err.Error(), "connect: connection refused"),
         strings.Contains(err.Error(), "read: connection reset by peer"):
        err = connection.TransientError{Inner: err}
    default:
        err = connection.PermanentError{Inner: fmt.Errorf("Permanent error: Unexpected Error while retrying: %v", err)}
    }
    return state, err
}```

**Review by avilagaston9 — COMMENTED** [review]
Left some comments!

File: core/chainio/avs_subscriber.go [hunk 26]
```diff
@@ -137,17 +144,20 @@ func (s *AvsSubscriber) SubscribeToNewTasksV3(newTaskCreatedChan chan *servicema
...
+	sub, err := SubscribeToNewTasksV3Retryable(s.AvsContractBindings.ServiceManager, internalChannel, s.logger)
 	if err != nil {
-		s.logger.Error("Failed to subscribe to new AlignedLayer tasks", "err", err)
+		s.logger.Error("Failed to subscribe to new AlignedLayer tasks after %d retries", MaxRetries, "err", err)
 		return nil, err
 	}
+	s.logger.Info("Subscribed to new AlignedLayer V3 tasks")
 
-	subFallback, err := subscribeToNewTasksV3(s.AvsContractBindings.ServiceManagerFallback, internalChannel, s.logger)
+	//TODO: collapse this
```
uri-99: what is this TODO ?

core/chainio/avs_subscriber.go
```diff
@@ -447,3 +399,221 @@ func (s *AvsSubscriber) WaitForOneBlock(startBlock uint64) error {
...
+				return sub, err
+			}
+			if strings.Contains(err.Error(), "connect: connection refused") {
+				err = connection.TransientError{Inner: err}
+				return sub, err
+			}
+			if strings.Contains(err.Error(), "read: connection reset by peer") {
+				return sub, connection.TransientError{Inner: err}
+			}
+			err = connection.PermanentError{Inner: fmt.Errorf("Permanent error: Unexpected Error while retrying: %s\n", err)}
```
uri-99: I don't think matching every other error case as Permanent is prudent

core/chainio/avs_subscriber.go
```diff
@@ -447,3 +399,221 @@ func (s *AvsSubscriber) WaitForOneBlock(startBlock uint64) error {
...
```
uri-99: +1

core/chainio/avs_subscriber.go
```diff
@@ -447,3 +399,221 @@ func (s *AvsSubscriber) WaitForOneBlock(startBlock uint64) error {
 // func (s *AvsSubscriber) ParseTaskResponded(rawLog types.Log) (*cstaskmanager.ContractAlignedLayerTaskManagerTaskResponded, error) {
 // 	return s.AvsContractBindings.TaskManager.ContractAlignedLayerTaskManagerFilterer.ParseTaskResponded(rawLog)
 // }
+
+// |---RETRYABLE---|
+
```
uri-99: I think retriable functions in their own file is going to be better.
This is being done in the batcher as well

core/chainio/avs_subscriber.go
```diff
@@ -447,3 +399,221 @@ func (s *AvsSubscriber) WaitForOneBlock(startBlock uint64) error {
...
+					return latestBlock, err
+				}
+				if strings.Contains(err.Error(), "connect: connection refused") {
+					err = connection.TransientError{Inner: err}
+					return latestBlock, err
+				}
+				if strings.Contains(err.Error(), "read: connection reset by peer") {
+					return latestBlock, connection.TransientError{Inner: err}
+				}
+				err = connection.PermanentError{Inner: fmt.Errorf("Permanent error: Unexpected Error while retrying: %s\n", err)}
```
uri-99: +1 casting every uncatched error to permanent

core/chainio/avs_subscriber.go
```diff
@@ -447,3 +399,221 @@ func (s *AvsSubscriber) WaitForOneBlock(startBlock uint64) error {
...
+				return logs, err
+			}
+			if strings.Contains(err.Error(), "connect: connection refused") {
+				err = connection.TransientError{Inner: err}
+				return logs, err
+			}
+			if strings.Contains(err.Error(), "read: connection reset by peer") {
+				return logs, connection.TransientError{Inner: err}
+			}
+			err = connection.PermanentError{Inner: fmt.Errorf("Permanent error: Unexpected Error while retrying: %s\n", err)}
```
uri-99: +1

core/chainio/avs_subscriber.go
```diff
@@ -447,3 +399,221 @@ func (s *AvsSubscriber) WaitForOneBlock(startBlock uint64) error {
...
```
uri-99: +1

core/chainio/avs_subscriber.go
```diff
@@ -447,3 +399,221 @@ func (s *AvsSubscriber) WaitForOneBlock(startBlock uint64) error {
...
+				err = connection.TransientError{Inner: err}
+				return state, err
+			}
+			if strings.Contains(err.Error(), "connect: connection refused") {
+				err = connection.TransientError{Inner: err}
+				return state, err
+			}
+			if strings.Contains(err.Error(), "read: connection reset by peer") {
+				return state, connection.TransientError{Inner: err}
+			}
```
uri-99: These ifs should be a switch

core/chainio/avs_subscriber.go
```diff
@@ -447,3 +399,221 @@ func (s *AvsSubscriber) WaitForOneBlock(startBlock uint64) error {
...
+			err = connection.PermanentError{Inner: fmt.Errorf("Permanent error: Unexpected Error while retrying: %s\n", err)}
```
uri-99: +1 default case permanent error

core/chainio/avs_subscriber.go
```diff
@@ -447,3 +399,221 @@ func (s *AvsSubscriber) WaitForOneBlock(startBlock uint64) error {
...
+				}
+				if strings.Contains(err.Error(), "connect: connection refused") {
+					err = connection.TransientError{Inner: err}
+					return sub, err
+				}
+				if strings.Contains(err.Error(), "read: connection reset by peer") {
+					return sub, connection.TransientError{Inner: err}
+				}
+				err = connection.PermanentError{Inner: fmt.Errorf("Permanent error: Unexpected Error while retrying: %s\n", err)}
+			}
```
uri-99: lol

aggregator/pkg/aggregator.go
```diff
@@ -387,6 +383,16 @@ func (agg *Aggregator) AddNewTask(batchMerkleRoot [32]byte, senderAddress [20]by
 	agg.logger.Info("New task added", "batchIndex", batchIndex, "batchIdentifierHash", "0x"+hex.EncodeToString(batchIdentifierHash[:]))
 }
 
+// |---RETRYABLE---|
+
```
uri-99, line 386: +1, retryables on their own file could be better.

core/utils/eth_client_utils.go
```diff
@@ -2,29 +2,41 @@ package utils
...
-func WaitForTransactionReceipt(client eth.InstrumentedClient, ctx context.Context, txHash gethcommon.Hash) (*types.Receipt, error) {
-	for i := 0; i < maxRetries; i++ {
-		receipt, err := client.TransactionReceipt(ctx, txHash)
+/*
+Errors:
+- "not found": (Transient) Call successfully returns but the tx receipt was not found.
+- "connect: connection refused": (Transient) Could not connect.
+*/
+func WaitForTransactionReceiptRetryable(client eth.InstrumentedClient, ctx context.Context, txHash gethcommon.Hash) (*types.Receipt, error) {
+	// For if no receipt and no error TransactionReceipt return "not found" as an error catch all ref: https://github.com/ethereum/go-ethereum/blob/master/ethclient/ethclient.go#L313
```
uri-99: this comment is not clear

core/connection_test.go
uri-99: +1

core/connection_test.go
```diff
@@ -0,0 +1,926 @@
...
+	if err := cmd.Process.Kill(); err != nil {
+		fmt.Printf("error killing process: %v\n", err)
+		return
+	}
+}
+
+/*
+func TestInitializeNewTaskRetryable(t *testing.T) {
+
+	//Start Anvil
```
uri-99: uncomment or remove tests?

uri-99 (changes requested):

core/chainio/avs_writer.go
uri-99, line 1: +1 to similar comments on other files.
Retryable on another file? cast default errors to permanent?

core/connection.go
```diff
@@ -0,0 +1,134 @@
...
+	return ok
+}
+
+const MinDelay = 1000
+const RetryFactor = 2
+const NumRetries = 3
+const MaxInterval = 60000
+
+// Same as Retry only that the functionToRetry can return a value upon correct execution
+func RetryWithData[T any](functionToRetry func() (T, error), minDelay uint64, factor float64, maxTries uint64, maxInterval uint64) (T, error) {
```
uri-99: i dont think `retryWithData` is a good variation of the name `retry` that returns data.
`retryReturingData` or similar is better IMO

core/connection.go
```diff
@@ -0,0 +1,134 @@
...
+	expBackoff := backoff.NewExponentialBackOff(randomOption, multiplierOption, initialRetryOption, maxIntervalOption)
+	var maxRetriesBackoff backoff.BackOff
+
+	if maxTries > 0 {
+		maxRetriesBackoff = backoff.WithMaxRetries(expBackoff, maxTries)
+	} else {
+		maxRetriesBackoff = expBackoff
+	}
+
+	return backoff.RetryWithData(f, maxRetriesBackoff)
```
MarcosNicolau: The library has another option: `WithMaxElapsedTime` that allows us to stop the retry when the backoff timer has surpassed the given time. This is very useful when we want to set the retries based on time more than the number of retries (when waiting for a receipt for example). If we set the default value to `0` there is no restriction over time (which is how it works now).

PatStiles: Addressed!

core/retry.go
```diff
@@ -0,0 +1,141 @@
...
+		}()
+		return err
+	}
+
+	randomOption := backoff.WithRandomizationFactor(0)
+
+	initialRetryOption := backoff.WithInitialInterval(time.Millisecond * time.Duration(minDelay))
+	multiplierOption := backoff.WithMultiplier(factor)
+	maxIntervalOption := backoff.WithMaxInterval(time.Millisecond * time.Duration(maxInterval))
+	maxElapsedTimeOption := backoff.WithMaxElapsedTime(time.Millisecond * time.Duration(maxElapsedTime))
```
MarcosNicolau: You shouldn't enforce the time to be passed in `time.Milisecond`. `maxElapsedTime` should be of type `time.duration`.

core/retry.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+const MinDelay = 1000
+const RetryFactor = 2
+const NumRetries = 3
+const MaxInterval = 60000
+const MaxElapsedTime = 0
+
+// Same as Retry only that the functionToRetry can return a value upon correct execution
+func RetryWithData[T any](functionToRetry func() (T, error), minDelay uint64, factor float64, maxTries uint64, maxInterval uint64, maxElapsedTime uint64) (T, error) {
```
MarcosNicolau: This params should be `time.Duration` instead of `uint`

core/chainio/avs_subscriber.go
PatStiles: addressed!

core/connection.go
PatStiles: This comes from the underlying backoff library ref: https://github.com/cenkalti/backoff/blob/v4/retry.go#L42

core/connection_test.go
PatStiles: removed

core/chainio/avs_writer.go
PatStiles: removed.

core/utils/eth_client_utils.go
PatStiles: removed

aggregator/pkg/aggregator.go
PatStiles: removed.

core/chainio/avs_subscriber.go
PatStiles: removed

config-files/config-aggregator-test.yaml
PatStiles: addressed

core/chainio/avs_subscriber.go
PatStiles: removed

core/chainio/avs_subscriber.go
PatStiles: removed

core/chainio/avs_subscriber.go
PatStiles: removed

core/chainio/avs_subscriber.go
PatStiles: The issue is in go you cannot declare methods outside of the local package. I was able to consolidate retryable functions within chainio but to consolidate them all I believe we would have to make additional refactors to the codebase.

aggregator/pkg/server.go
PatStiles: removed

aggregator/pkg/server.go
PatStiles: removed.

aggregator/pkg/server.go
PatStiles: removed

aggregator/pkg/server.go
PatStiles: removed

core/chainio/avs_subscriber.go
PatStiles: Good catch! Addressed!

config-files/config-aggregator-test.yaml
PatStiles: yes Its used for retry tests for the aggregator retryable functions that will be added once the sdk is fixed.

aggregator/pkg/aggregator.go
PatStiles: removed!

aggregator/pkg/aggregator.go
PatStiles: Yes, but after consulting with you and pelito I decided to remove the catches as we have yet to identify all permanent error cases.

core/chainio/avs_subscriber.go
PatStiles: I see your point.. I'm more of a fan of the suffix as it separates the logic from the retry logic and makes it explicit that the function (that is wraps) will be retried. I'd tend towards just inlining the retry library calling the retry function but for testing it makes sense to have separate functions to test.

core/chainio/avs_subscriber.go
PatStiles: removed

core/chainio/avs_subscriber.go
PatStiles: removed

core/chainio/avs_subscriber.go
PatStiles: removed

core/chainio/avs_subscriber.go
PatStiles: revmoved

core/retry.go
PatStiles: addressed

core/retry.go
PatStiles: addressed

core/connection.go
PatStiles: transient error has been removed

MarcosNicolau:
@PatStiles We should add the new tests to the ci.

aggregator/pkg/server.go
PatStiles: Sounds good! I think its best we remain consistent with the naming scheme in the rust ones so I'll make the change. My only take is I think it  could be cleaner to distinguish between a the function with the logic and then have the `Retryable` suffix reserved for the function that calls the retry library with said function.

PatStiles: Addresed!

core/connection.go
PatStiles: After consulting with @Oppen I've decided to leave it

core/connection.go
PatStiles: addressed

core/chainio/avs_writer.go
PatStiles: I elected to just remove the catch statements and add more explicit permanent errors in a later pr

core/retry.go
```diff
@@ -0,0 +1,130 @@
...
+const (
+	MinDelay               = 1 * time.Second
+	MaxInterval            = 60 * time.Second
+	MaxElapsedTime         = 0 * time.Second
+	RetryFactor    float64 = 2
+	NumRetries     uint64  = 3
+)
+
+// Same as Retry only that the functionToRetry can return a value upon correct execution
+func RetryWithData[T any](functionToRetry func() (T, error), minDelay time.Duration, factor float64, maxTries uint64, maxInterval time.Duration, maxElapsedTime time.Duration) (T, error) {
```
MarcosNicolau: These functions receive a lot of parameters and it is easy to get lost. I would refactor this to receive a struct instead with all the config. This will also make passing a default (which is what we are doing in most if not all retryable functions) much easier, as you would only pass one param and reuse the struct. I'll give you a hint:

```go
type RetryConfig {
MinDelay time.Duration
MaxInterval time.Duration
MaxDelay time.Duration
RetryFactor float64
NumRetries uint64
}

func RetryWithData[T any](functionToRetry func() (T,error), config *RetryConfig) { }

// The you would you use like so 
func foo() {
RetryWithData(funcToRetry, &RetryConfig{ RetryFactor: 5 , ...})
}
```

core/retry.go
```diff
@@ -0,0 +1,184 @@
...
+
+/*
+Retry and RetryWithData are custom retry functions used in Aligned's aggregator and operator to facilitate consistent retry logic across the system.
+They are interfaces for around Cenk Alti (https://github.com/cenkalti) backoff library (https://github.com/cenkalti/backoff). We would like to thank him for his great work.
+
+The `Retry` and `RetryWithData` retry a supplied function at maximum `NumRetries` number of times. Upon execution, if the called function returns an error the retry library either re-executes the function (Transient Error) or exits and returns the error to the calling context (Permanent error) .
+If the call is successful and no error is returned the library returns the result. `Permanent` errors are explicitly typed while `Transient` errors are implied by go's builtin error type.
+For completeness:
+
+Transient: The error is recoverable and the function is retried after failing. `Transient` errors are do not have a defined error type and are implicitly defined by go's builtin `error` type.
```
JulianVentura:
```suggestion
Transient: The error is recoverable and the function is retried after failing. `Transient` errors do not have a defined error type and are implicitly defined by go's builtin `error` type.
```

core/retry.go
```diff
@@ -0,0 +1,184 @@
...
+
+Permanenet: The error is not recoverable and the function is not retried exit to a larger context to handle the error. Permanent errors are explicitly typed and defined by wrapping the err within with `PermanentError`.
```
JulianVentura: This paragraph is a bit confusing:
"Permanenet: The error is not recoverable and the function is not retried exit to a larger context to handle the error."

aggregator/pkg/server.go
```diff
@@ -121,3 +111,52 @@ func (agg *Aggregator) ServerRunning(_ *struct{}, reply *int64) error {
...
 	return nil
 }
+
+// |---RETRYABLE---|
+
+/*
+  - Errors:
+    Permanent:
+  - TaskNotFoundError: Task corresponding to signature is not found within BLS Aggregation Service. (https://github.com/Layr-Labs/eigensdk-go/blob/dev/services/bls_aggregation/blsagg.go#L33)
+  - SignatureVerificationError: Verification of the sigature failed within the BLS Aggregation Service failed. (https://github.com/Layr-Labs/eigensdk-go/blob/dev/services/bls_aggregation/blsagg.go#L42)
```
JulianVentura:
```suggestion
  - SignatureVerificationError: Verification of the sigature within the BLS Aggregation Service failed. (https://github.com/Layr-Labs/eigensdk-go/blob/dev/services/bls_aggregation/blsagg.go#L42)
```

JulianVentura (commented):
Go tests working on my machine, left some nit comments of documentation

Oppen (approved):

aggregator/pkg/server.go
PatStiles: addressed!

core/retry.go
PatStiles: addressed!

core/retry.go
PatStiles: addressed!

JulianVentura (approved):
LGTM

uri-99 (approved):
Ready when ci passes

PR #1308, opened 2024-10-23
hotfix(batcher): rm MIN_FEE_PER_PROOF check in batcher

### LINKED ISSUE #1268: fix(price_estimate): adjust `estimate_fee` parameters
In the process of bumping the sdk version for zkRust I discovered the `PriceEstimate::Default` setting computed fee was lower than the [MIN_FEE_PER_PROOF](https://github.com/yetanotherco/aligned_layer/blob/testnet/batcher/aligned-batcher/src/lib.rs#L55). Using `PriceEstimate::Instant` succeeded Intermittently. 

We should remove this check in the batcher: https://github.com/yetanotherco/aligned_layer/blob/testnet/batcher/aligned-batcher/src/lib.rs#L472

PatStiles:
# Hotfix Removes MIN_FEE_PER_PROOF check from Batcher

closes #1268 

## Description

Removes a check that the` max_fee` of proof submitted to the batcher is greater than `MIN_FEE_PER_PROOF` https://github.com/yetanotherco/aligned_layer/blob/testnet/batcher/aligned-batcher/src/lib.rs#L472 .

## Type of change

- [x] Bug fix

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/batcher/aligned-batcher/src/lib.rs
+++ b/batcher/aligned-batcher/src/lib.rs
@@ -469,12 +469,6 @@ impl Batcher {
 
         // Nonce and max fee verification
         let max_fee = nonced_verification_data.max_fee;
-        if max_fee < U256::from(MIN_FEE_PER_PROOF) {
-            error!("The max fee signed in the message is less than the accepted minimum fee to be included in the batch.");
-            send_message(ws_conn_sink.clone(), ValidityResponseMessage::InvalidMaxFee).await;
-            return Ok(());
-        }
-
         // Check that we had a user state entry for this user and insert it if not.
 
         // We aquire the lock first only to query if the user is already present and the lock is dropped.
--- a/batcher/aligned-batcher/src/lib.rs
+++ b/batcher/aligned-batcher/src/lib.rs
@@ -467,10 +467,6 @@ impl Batcher {
             return Ok(());
         }
 
-        // Nonce and max fee verification
-        let max_fee = nonced_verification_data.max_fee;
-        // Check that we had a user state entry for this user and insert it if not.
-
         // We aquire the lock first only to query if the user is already present and the lock is dropped.
         // If it was not present, then the user nonce is queried to the Aligned contract.
         // Lastly, we get a lock of the batch state again and insert the user state if it was still missing.


Oppen:
Questions:
1. Are we sure the fix is not increasing the default to surpass the minimum fee instead?
2. We're pointing to testnet. Since we merged to testnet by accident a few times, I want to be extra sure this time it's on purpose.

Observation: CI is failling due to a lint:
```
   --> aligned-batcher/src/lib.rs:471:13
    |
471 |         let max_fee = nonced_verification_data.max_fee;
    |             ^^^^^^^ help: if this is intentional, prefix it with an underscore: `_max_fee`
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -467,16 +467,6 @@ impl Batcher {
             return Ok(());
         }
 
-        // Nonce and max fee verification
-        let max_fee = nonced_verification_data.max_fee;
-        if max_fee < U256::from(MIN_FEE_PER_PROOF) {
```
JuArce, line 472: There is missing remove MIN_FEE_PER_PROOF variable

JuArce (changes requested):

batcher/aligned-batcher/src/lib.rs
PatStiles: @JuArce the MIN_FEE_PER_PROOF is used here as well: https://github.com/yetanotherco/aligned_layer/blob/testnet/batcher/aligned-batcher/src/lib.rs#L617 .

PatStiles:
@Oppen thanks for the find:
1.) The issue stems from the MIN_FEE_PER_PROOF (and respective constants) using being derived with respect to values that are not adjusted to the holesky gas_price (currentlty ~10^7 wei at time of writing) and is out of date. We could modify the parameters. My understanding is this check was added at some point and is unneeded as later the batcher checks that a user has the funds available to pay for proof submission. Pinging @MauroToscano for further comment.
2.) We want to deploy this quickly so users can submit there proofs to mainnet.

batcher/aligned-batcher/src/lib.rs
JuArce: I think if it is used in another place it will have the problem in some point

MauroToscano:
This is going to testnet. There is no point in having a minimium fee. The whole thing can be removed

MarcosNicolau (approved):
If this is a hotfix I think the solution is fine, I tested it by doing:
- Add the check back again
- Add `--max_fee` param in `batcher_send_risc0_burst` to a low value, for example: `--max_fee 100`
- Send proofs and see they get invalidated
- Remove the check, restart the batcher, send proofs again, see they are not rejected now

Note: you need to add `--private_key` so that it does not use the non-paying addres.

In the future, I think we should have a message in the `batcher` to retrieve the current `MinFeePerProof`.

uri-99 (approved):

JuArce (approved):

PR #1317, opened 2024-10-24
fix: add retry and timeout to create_new_task

### LINKED ISSUE #1293: Abort batch in Ethereum in case of timeout

avilagaston9:
# Add retry and timeout to `create_new_task`

> [!WARNING]
> This PR requires the deploy of new metrics/dashboards

## Motivation
Currently, if the gas price rises abruptly, we don't have a mechanism to abort the task sent by the batcher, causing it to get stuck while waiting for the transaction Receipt.

### General Description
- This PR adds a timeout of 3 blocks for waiting on the created task receipt.
- If the timeout is reached, proceed as follows:
    1. Gets a `bumped_gas_price` based on the current gas price of the net, the previous used one and the iteration number.
    2. Sends a transaction with the same nonce and the `bumped_gas_price` to override the previous transaction, aborting the batch inclusion.
    3. Waits for another 3 blocks to get the Receipt of the cancel transaction. 
    4. If the timeout is reached, increments the iteration number and retries from step `1` using exponential backoff with a `usize::MAX` number of retries until the Receipt is received.

Observation: We use an exponential backoff because the timeout is not the only thing that can fail. For example, the `get_gas_price` call or sending the transaction can fail too.

### Bump Gas Price Mechanism
- The formula used to bump each gas price is as follows:
    - `plus_per_iteration = GAS_PRICE_INCREMENT_PERCENTAGE_PER_ITERATION * iteration`
    - `override_gas_multiplier = OVERRIDE_GAS_PRICE_PERCENTAGE_MULTIPLIER + plus_per_iteration`.
    - `bumped_gas_price = gas_price * override_gas_multiplier / PERCENTAGE_DIVIDER`.

### Constants added
- `transaction_wait_timeout`: 36_000 milliseconds (3 blocks)
- `GAS_PRICE_INCREMENT_PERCENTAGE_PER_ITERATION`: 5
- `OVERRIDE_GAS_PRICE_PERCENTAGE_MULTIPLIER`: 120

### Additional Changes
- Replaces the `http_providers` with `batcher_signers` because they are needed to sign the cancel transaction.
- Removes the `GasEscalatorMiddleware` from our `SignerMiddlewareT` type, as we don't want the `Provider` to bump the created tasks for us.
- Moves `get_gas_price` and `get_current_nonce` functions to `batcher/aligned-batcher/src/eth/utils.rs`.
- Adds the `transaction_wait_timeout` constant to the batcher config file. This is the configurable time we wait for a transaction to be included in Ethereum on each attempt.
- Adds metrics and a new Grafana panel to track the count of canceled batches.

image

## How to test
Modify the `create_new_task` function in `batcher/aligned-batcher/src/lib.rs` to call `create_new_task_retryable` with a timeout of only one millisecond:
```Rust
create_new_task_retryable(
                    batch_merkle_root,
                    batch_data_pointer.clone(),
                    proof_submitters.clone(),
                    fee_params.clone(),
                    1,
                    &self.payment_service,
                    &self.payment_service_fallback,
                )
```
Then, run the following:
```
make batcher_start_local
```

After some time, you should see the following logs in the batcher:
```
2024-10-25T00:56:17Z INFO  aligned_batcher::retry::batcher_retryables] Creating task for: 0x32e14361580d28b131fc38ac0238e80428062f839629b650cb03e53234a9ff7b
[2024-10-25T00:56:17Z WARN  aligned_batcher::retry::batcher_retryables] Error while waiting for batch inclusion: deadline has elapsed
[2024-10-25T00:56:17Z INFO  aligned_batcher] Cancelling createNewTask transaction...
[2024-10-25T00:56:17Z INFO  aligned_batcher] createNewTask transaction successfully canceled
[2024-10-25T00:56:17Z ERROR aligned_batcher] Failed to send batch to contract, batch will be lost: Receipt not found
```

You can also modify the call to `cancel_create_new_task_retryable`in the same way to see the retries while canceling the `created_task`:
```Rust
cancel_create_new_task_retryable(
                    &self.batcher_signer,
                    &self.batcher_signer_fallback,
                    bumped_gas_price,
                    1,
                )
```

## Type of change

- [x] New feature
- [x] Bug fix
- [ ] Optimization
- [x] Refactor

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added
  - [ ] Benchmarks added/run
- [x] Has a known issue
  - #1293


batcher/aligned-batcher/src/lib.rs
```diff
@@ -1189,39 +1199,67 @@ impl Batcher {
...
+            Ok(receipt) => Ok(receipt),
+            Err(RetryError::Permanent(BatcherError::ReceiptNotFoundError)) => {
+                self.override_created_task(fee_params.gas_price).await;
+                Err(BatcherError::ReceiptNotFoundError)
             }
+            Err(_) => Err(BatcherError::TransactionSendError),
         }
     }
 
+    pub async fn override_created_task(&self, task_gas_price: U256) {
```
entropidelic: small nit, I think it makes the intent a little clearer
```suggestion
    pub async fn cancel_create_new_task_tx(&self, task_gas_price: U256) {
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1189,39 +1199,67 @@ impl Batcher {
...
+        let from_address = self.batcher_signer.address();
```
```suggestion
        let batcher_addr = self.batcher_signer.address();
```

batcher/aligned-batcher/src/lib.rs
entropidelic: is the `task_gas_price` argument the bumped gas price to cancel the old one? or is it the old gas price? Let's make the argument name a little clearer about this.

edit: seeing the rest of the code I understanded which one is but making it a little more explicit doesn't hurt (:

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1189,39 +1199,67 @@ impl Batcher {
...
+        let current_nonce = self
+            .batcher_signer
+            .get_transaction_count(from_address, None)
+            .await
+            .unwrap();
```
entropidelic: let's not unwrap 👀

entropidelic: maybe we should think about what to do if the cancelling of the transaction fails

entropidelic: if there is an error here, we may want to try with fallback too

batcher/aligned-batcher/src/lib.rs
entropidelic: it might make sense to ask for the current gas price again inside the function and bump it a little to make sure that the cancelling tx is included rather than using the old gas price

batcher/aligned-batcher/src/lib.rs
avilagaston9: Totally missed it, thank you! Trying it with the fallback sounds good too.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1189,39 +1199,67 @@ impl Batcher {
...
-                Ok(receipt)
+            },
+            DEFAULT_MIN_RETRY_DELAY,
+            DEFAULT_BACKOFF_FACTOR,
+            DEFAULT_MAX_RETRIES,
+        )
+        .await;
+        match result {
+            Ok(receipt) => Ok(receipt),
+            Err(RetryError::Permanent(BatcherError::ReceiptNotFoundError)) => {
```
entropidelic, line 1253: lets add a counter metric when this happens so we have good visibility

batcher/aligned-batcher/src/eth/utils.rs
```diff
@@ -10,3 +16,24 @@ pub fn get_provider(eth_rpc_url: String) -> Result<Provider<Http>, anyhow::Error
...
+
+pub async fn get_batcher_signer(
+    provider: Provider<Http>,
+    ecdsa_config: ECDSAConfig,
+) -> Result<Arc<SignerMiddlewareT>, anyhow::Error> {
+    let chain_id = provider.get_chainid().await?;
+
+    let escalator = GeometricGasPrice::new(GAS_MULTIPLIER, GAS_ESCALATOR_INTERVAL, None::<u64>);
+
+    let provider = GasEscalatorMiddleware::new(provider, escalator, Frequency::PerBlock);
```
entropidelic: is this needed? I think it may be some legacy code

batcher/aligned-batcher/src/lib.rs
entropidelic: also, we may want to make this function a retryable and do some retries.

batcher/aligned-batcher/src/eth/utils.rs
avilagaston9: In theory, this increases the transaction's gas price every block. I can remove it if you want

MarcosNicolau: I would prefer to write this code ourselves, it shouldn't be so hard. We already have some of it written in go.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1226,7 +1226,7 @@ impl Batcher {
     }
 
     pub async fn override_created_task(&self, task_gas_price: U256) {
-        let from_address = self.batcher_signer.address();
+        let batcher_addr = self.batcher_signer.address();
```
MarcosNicolau: Careful here, you have not changed the references and the project is not compiling.

batcher/aligned-batcher/src/lib.rs
avilagaston9: Yes, I will create a retryable function for this!

batcher/aligned-batcher/src/eth/utils.rs
MarcosNicolau: @entropidelic what do you think?

batcher/aligned-batcher/src/lib.rs
MarcosNicolau: We could also try indefinitely bumping the gas price until it gets accepted, at some point, it should be accepted, this way we would avoid flushing the whole queue. What do you think @entropidelic?

batcher/aligned-batcher/src/retry/batcher_retryables.rs
```diff
@@ -77,12 +85,78 @@ pub async fn get_gas_price_retryable(
...
+                    warn!("Transaction reverted {:?}", err);
+                    return Err(RetryError::Permanent(BatcherError::TransactionSendError));
+                }
+                _ => return Err(RetryError::Transient(BatcherError::TransactionSendError)),
+            }
+        }
+    };
+
+    // timeout to prevent a deadlock while waiting for the transaction to be included in a block.
+    timeout(Duration::from_millis(BATCH_INCLUSION_DELAY), pending_tx)
```
MarcosNicolau: Shouldn't this return `TransientErrors` as they should be recoverable by sending the tx again?

batcher/aligned-batcher/src/eth/utils.rs
entropidelic: @MarcosNicolau yeah I think it would be good to control all this retry logic ourselves

batcher/aligned-batcher/src/retry/batcher_retryables.rs
entropidelic: @MarcosNicolau if the timeout is reached, we should cancel the tx. This is because as the gas price might have changed, if we want to re-send the same transaction we should re-calculate what users want to have their proof included in the batch depending on the max fee they have set. 
As this is a very complicated logic, it is much easier to just cancel the transaction for now, so the timeout error should be a permanent one

batcher/aligned-batcher/src/lib.rs
entropidelic: @MarcosNicolau yes I agree, the cancel transaction should be included, if not the processing of tasks will get locked. A similar logic to what you did in the aggregator bump PR could be implemented 👍

batcher/aligned-batcher/src/retry/batcher_retryables.rs
MarcosNicolau: Oh yes!

batcher/aligned-batcher/src/lib.rs
avilagaston9: Made `cancel_create_new_tx` infinitely retryable!

batcher/aligned-batcher/src/eth/utils.rs
avilagaston9: It was legacy code. Removed!

batcher/aligned-batcher/src/lib.rs
avilagaston9: Fixed!

batcher/aligned-batcher/src/lib.rs
avilagaston9: Done!

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1189,39 +1192,62 @@ impl Batcher {
...
+                Err(BatcherError::ReceiptNotFoundError)
             }
+            Err(_) => Err(BatcherError::TransactionSendError),
         }
     }
 
+    // Sends a transaction to Ethereum with the same nonce as the previous one to override it.
+    // We use the maximum number of retries here because this transaction must be included.
+    pub async fn cancel_create_new_task_tx(&self, old_task_gas_price: U256) {
+        info!("Canceling created task");
```
entropidelic: logging style nit :)
```suggestion
        info!("Cancelling createNewTask transaction...");
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1189,39 +1192,62 @@ impl Batcher {
...
+                self.cancel_create_new_task_tx(fee_params.gas_price).await;
+                Err(BatcherError::ReceiptNotFoundError)
             }
+            Err(_) => Err(BatcherError::TransactionSendError),
         }
     }
 
+    // Sends a transaction to Ethereum with the same nonce as the previous one to override it.
+    // We use the maximum number of retries here because this transaction must be included.
+    pub async fn cancel_create_new_task_tx(&self, old_task_gas_price: U256) {
```
```suggestion
    pub async fn cancel_create_new_task_tx(&self, old_tx_gas_price: U256) {
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1189,39 +1192,62 @@ impl Batcher {
...
+                    previous_gas_price.clone(),
+                )
+            },
+            DEFAULT_MIN_RETRY_DELAY,
+            DEFAULT_BACKOFF_FACTOR,
+            usize::MAX,
+        )
+        .await
+        {
+            error!("Could not cancel created task: {e}");
```
```suggestion
            error!("Could not cancel createNewTask transaction: {e}");
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1189,39 +1192,62 @@ impl Batcher {
...
+            return;
+        };
+
+        info!("Task successfully canceled");
```
```suggestion
        info!("createNewTask transaction successfully canceled");
```

batcher/aligned-batcher/src/retry/batcher_retryables.rs
```diff
@@ -60,20 +88,151 @@ pub async fn user_balance_is_unlocked_retryable(
...
+            RetryError::Transient(BatcherError::TransactionSendError)
+        })?
+        .ok_or(RetryError::Permanent(BatcherError::ReceiptNotFoundError))
+}
+
+pub async fn cancel_create_new_task_retryable(
+    batcher_signer: &SignerMiddlewareT,
+    batcher_signer_fallback: &SignerMiddlewareT,
+    iteration: Arc<Mutex<usize>>,
+    previous_gas_price: Arc<Mutex<U256>>,
```
entropidelic: maybe I'm not seeing something but do we really need the `Arc<Mutex>`s?

batcher/aligned-batcher/src/eth/utils.rs
```diff
@@ -1,12 +1,137 @@
...
+}
+
+/// Calculates an increased gas price for retrying a transaction override.
+/// The gas price rises with each retry by applying a multiplier based on the iteration count.
+pub fn get_bumped_gas_price(
+    previous_gas_price: U256,
+    current_gas_price: U256,
+    iteration: usize,
+) -> U256 {
+    let override_gas_multiplier = U256::from(OVERRIDE_GAS_PRICE_MULTIPLIER) + (5 * iteration);
```
JuArce: What is the 5 meaning?

batcher/aligned-batcher/src/eth/utils.rs
```diff
@@ -1,12 +1,137 @@
...
+    let bumped_previous_gas_price =
+        previous_gas_price * override_gas_multiplier / U256::from(PERCENTAGE_DIVIDER);
+
+    let bumped_current_gas_price =
+        current_gas_price * override_gas_multiplier / U256::from(PERCENTAGE_DIVIDER);
+    bumped_current_gas_price.max(bumped_previous_gas_price)
+}
```
JuArce: Why are we calculating two gas prices and choosing the max?

batcher/aligned-sdk/src/core/constants.rs
```diff
@@ -1,4 +1,6 @@
 /// Batcher ///
+pub const BATCH_INCLUSION_DELAY: u64 = 36_000; // 3 blocks
+pub const TRANSACTIONS_INCLUSION_DELAY: u64 = 14_000; // 1 block + 2 secs
```
JuArce: These values are not delays, these are timeouts
Also, the name is not clear. 
I'd suggest
`TRANSACTION_WAIT_TIMEOUT`
Also, I don't think it is necessary to use two variables

batcher/aligned-sdk/src/core/constants.rs
```diff
@@ -13,6 +15,7 @@ pub const MIN_FEE_PER_PROOF: u128 = ADDITIONAL_SUBMISSION_GAS_COST_PER_PROOF * 1
 pub const RESPOND_TO_TASK_FEE_LIMIT_PERCENTAGE_MULTIPLIER: u128 = 250; // fee_for_aggregator -> respondToTaskFeeLimit modifier
 pub const DEFAULT_AGGREGATOR_FEE_PERCENTAGE_MULTIPLIER: u128 = 150; // feeForAggregator modifier
 pub const GAS_PRICE_PERCENTAGE_MULTIPLIER: u128 = 110; // gasPrice modifier
+pub const OVERRIDE_GAS_PRICE_MULTIPLIER: u128 = 120; // gasPrice modifier to override previous transactions
```
JuArce:
```suggestion
pub const OVERRIDE_GAS_PRICE_PERCENTAGE_MULTIPLIER: u128 = 120; // gasPrice modifier to override previous transactions
```

JuArce (changes requested):
It should add a note in the description that it requires a deploy of new metrics/dashboards ✅

batcher/aligned-batcher/src/eth/utils.rs
```diff
@@ -1,12 +1,137 @@
...
 pub fn get_provider(eth_rpc_url: String) -> Result<Provider<Http>, anyhow::Error> {
     let provider = Http::from_str(eth_rpc_url.as_str())
         .map_err(|e| anyhow::Error::msg(format!("Failed to create provider: {}", e)))?;
     Ok(Provider::new(provider))
 }
+
+pub async fn get_batcher_signer(
+    provider: Provider<Http>,
+    ecdsa_config: ECDSAConfig,
+) -> Result<Arc<SignerMiddlewareT>, anyhow::Error> {
```
IAvecilla: This is the same as `anyhow::Result<Arc<SignerMiddlewareT>` the anyhow Result already wraps the anyhow error.

batcher/aligned-batcher/src/eth/utils.rs
```diff
@@ -1,12 +1,137 @@
 use std::str::FromStr;
+use std::sync::Arc;
 
+use crate::config::ECDSAConfig;
+use crate::retry::batcher_retryables::{get_current_nonce_retryable, get_gas_price_retryable};
+use crate::retry::retry_function;
+use crate::types::errors::BatcherError;
```
IAvecilla: nit: you can group this into one `use` statement doing `use crate::{config::{...}, retry::{...}, ...}`

batcher/aligned-batcher/src/eth/utils.rs
```diff
@@ -1,12 +1,137 @@
...
+    let bumped_current_gas_price =
+        current_gas_price * override_gas_multiplier / U256::from(PERCENTAGE_DIVIDER);
+    bumped_current_gas_price.max(bumped_previous_gas_price)
+}
+
+pub async fn get_current_nonce(
+    eth_http_provider: &Provider<Http>,
+    eth_http_provider_fallback: &Provider<Http>,
+    addr: H160,
+) -> Result<U256, String> {
```
IAvecilla: I'm not a fan of returning a String as an error here, I suggest two options here that we can discuss:
- Just use the `ProviderError` from ethers instead of parsing it everytime to a String we already can use those errors and in case we need a custom one there's a `ProviderError::CustomError(string)` ready to use
- Just use a new error type similar to the `BatcherError` that wraps the provider errors and use that for the `RetryError<>`

Usually works with errors as a strings is not a good idea but it's not importante so maybe we can address this in the future.

batcher/aligned-batcher/src/retry/batcher_retryables.rs
```diff
@@ -40,6 +51,23 @@ pub async fn get_user_nonce_from_ethereum_retryable(
...
 }
 
+pub async fn get_current_nonce_retryable(
+    eth_http_provider: &Provider<Http>,
+    eth_http_provider_fallback: &Provider<Http>,
+    addr: Address,
+) -> Result<U256, RetryError<String>> {
+    if let Ok(current_nonce) = eth_http_provider.get_transaction_count(addr, None).await {
+        return Ok(current_nonce);
+    }
```
IAvecilla: Maybe you can use a `match` here instead of `if let`. Something like:
```
match eth_http_provider.get_transaction_count(addr, None).await {
  Ok(nonce) => Ok(nonce),
  Err(_) => eth_http_provider_fallback ...
}

File: batcher/aligned-batcher/src/retry/batcher_retryables.rs [hunk 21]
```diff
@@ -60,20 +88,151 @@ pub async fn user_balance_is_unlocked_retryable(
 }
 
 pub async fn get_gas_price_retryable(
-    eth_ws_provider: &Provider<Http>,
-    eth_ws_provider_fallback: &Provider<Http>,
+    eth_http_provider: &Provider<Http>,
+    eth_http_provider_fallback: &Provider<Http>,
 ) -> Result<U256, RetryError<String>> {
-    if let Ok(gas_price) = eth_ws_provider
+    if let Ok(gas_price) = eth_http_provider
```
IAvecilla: Same as above

grafana/provisioning/dashboards/aligned/aggregator_batcher.json
```diff
@@ -1323,6 +1513,6 @@
   "timezone": "browser",
   "title": "Aggregator Data",
   "uid": "aggregator",
-  "version": 6,
+  "version": 2,
   "weekStart": ""
-}
+}
```
IAvecilla: I know that this is a generated Grafana file but I would add the EOL here

batcher/aligned-batcher/src/retry/batcher_retryables.rs
avilagaston9: We need these to avoid Rust errors when attempting to pass a `&mut` into an async closure. We discussed this with @Oppen and decided to handle `Arc<Mutex>` only within the `retry_function`, removing them from this function.

batcher/aligned-batcher/src/eth/utils.rs
avilagaston9: I want to cover the case where the `gas_price` decreases and the `previous_gas_price` sent is higher than the current one, to avoid sending a transaction with a lower gas_price than the previous one.

batcher/aligned-batcher/src/eth/utils.rs
avilagaston9: Added the constant `GAS_PRICE_INCREMENT_PERCENTAGE_PER_ITERATION` for this!

batcher/aligned-sdk/src/core/constants.rs
avilagaston9: Changed! Left only `TRANSACTION_WAIT_TIMEOUT` with a time of 3 blocks.

batcher/aligned-sdk/src/core/constants.rs
avilagaston9: Updated!

batcher/aligned-batcher/src/eth/utils.rs
avilagaston9: Updated!

batcher/aligned-batcher/src/eth/utils.rs
avilagaston9: Done!

batcher/aligned-batcher/src/retry/batcher_retryables.rs
avilagaston9: Done!

batcher/aligned-batcher/src/retry/batcher_retryables.rs
avilagaston9: Done!

grafana/provisioning/dashboards/aligned/aggregator_batcher.json
avilagaston9: Actually, my formatter was removing it. Thanks!

batcher/aligned-batcher/src/eth/utils.rs
avilagaston9: Changed to`ProvederError`s!

batcher/aligned-batcher/src/retry/batcher_retryables.rs
```diff
@@ -60,29 +87,140 @@ pub async fn user_balance_is_unlocked_retryable(
...
+            .await
+            .map_err(|e| RetryError::Transient(ProviderError::CustomError(e.to_string())))?,
+    };
+
+    // timeout to prevent a deadlock while waiting for the transaction to be included in a block.
+    timeout(Duration::from_millis(transaction_wait_timeout), pending_tx)
+        .await
+        .map_err(|e| {
+            warn!("Timeout while waiting for transaction inclusion: {e}");
+            RetryError::Transient(ProviderError::CustomError("Receipt not found".to_string()))
```
Oppen: Wouldn't the error be something timeout related here?

batcher/aligned-batcher/src/retry/batcher_retryables.rs
```diff
@@ -60,29 +87,140 @@ pub async fn user_balance_is_unlocked_retryable(
...
+    let pending_tx = match batcher_signer.send_transaction(tx.clone(), None).await {
+        Ok(pending_tx) => pending_tx,
+        Err(_) => batcher_signer_fallback
+            .send_transaction(tx.clone(), None)
+            .await
+            .map_err(|e| RetryError::Transient(ProviderError::CustomError(e.to_string())))?,
+    };
+
+    // timeout to prevent a deadlock while waiting for the transaction to be included in a block.
+    timeout(Duration::from_millis(transaction_wait_timeout), pending_tx)
```
Oppen, line 201: I'm not sure what this timeout's role is. Didn't we already await on the tx?

JuArce (approved):

batcher/aligned-batcher/src/retry/batcher_retryables.rs
avilagaston9: The first `.await` returns an [ethers::PendingTransaction](https://docs.rs/ethers/latest/ethers/providers/struct.PendingTransaction.html) that has to be awaited to get the transaction Receipt. In other words, the first `await` is to send the transaction and the second one is to get the `Receipt`.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1167,39 +1178,83 @@ impl Batcher {
...
-                // Since transaction was reverted, we don't want to retry with fallback.
-                warn!("Transaction reverted {:?}", err);
-
-                Err(BatcherError::TransactionSendError)
-            }
-            Err(_) => {
-                let receipt = try_create_new_task(
+        let result = retry_function(
+            || {
+                create_new_task_retryable(
```
MarcosNicolau, line 1227: Shouldn't we bump the gas price here as well, not just when canceling?

batcher/aligned-batcher/src/retry/batcher_retryables.rs
avilagaston9: You want me to improve the description of the error here?

batcher/aligned-batcher/src/lib.rs
avilagaston9: We can't bump the task since we are handling the user's fees here.

MarcosNicolau: Right

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1167,39 +1178,83 @@ impl Batcher {
...
+                    &self.batcher_signer,
+                    &self.batcher_signer_fallback,
+                    bumped_gas_price,
+                    self.transaction_wait_timeout,
+                )
+                .await
+            },
+            DEFAULT_MIN_RETRY_DELAY,
+            DEFAULT_BACKOFF_FACTOR,
+            usize::MAX,
```
MarcosNicolau: I would set a limited number of retries, as the batcher might get stuck here and I think that is worse than losing a transaction.

MarcosNicolau: Also, we should send an error message to the clients to finalize the connection.

MarcosNicolau (changes requested):
I tested on macos and linux and everything went well. I left some comments for minor improvements.

batcher/aligned-batcher/src/retry/batcher_retryables.rs
Oppen: Yes.

batcher/aligned-batcher/src/lib.rs
avilagaston9: We use a `MAX` number of retries here because we won’t be able to create a new task until this transaction is included in Ethereum. We also won’t get stuck here as we are bumping the transaction with each attempt.

batcher/aligned-batcher/src/retry/batcher_retryables.rs
avilagaston9: Done!

MarcosNicolau (approved):

batcher/aligned-batcher/src/lib.rs
Oppen: Wouldn't that lead to unbounded gas costs?

avilagaston9: As discussed offline, we will review that scenario in the future.

MauroToscano (changes requested):
Don't bump fee at 0.5 s, nothing in Ethereum gets done in half a second. Use more reasonable constants

MauroToscano:
Add the constants used in the PR, I want to check the mechanism

avilagaston9:
> Add the constants used in the PR, I want to check the mechanism
>
> @MauroToscano 

My bad. I’ve updated the description to include the changes made. Let me know if you’d like me to review anything

batcher/aligned-batcher/src/lib.rs
Oppen: Is there an issue? Can you add a `TODO` comment?

avilagaston9: Done! Created #1380 and added a reference to the issue in a `TODO` comment.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1232,25 +1250,63 @@ impl Batcher {
...
-                    proof_submitters,
-                    fee_params,
-                    &self.payment_service_fallback,
+            Err(_) => Err(BatcherError::TransactionSendError),
+        }
+    }
+
+    // Sends a transaction to Ethereum with the same nonce as the previous one to override it.
+    // Uses exponential backoff with the maximum number of retries because this transaction must be included.
+    // Bumps the gas price used and waits for 3 blocks on each attepmt to get the Receipt of the cancel transaction.
```
```suggestion
    /// Sends a transaction to Ethereum with the same nonce as the previous one to override it.
    /// In case of a recoverable error, it will retry with an exponential backoff forever
    /// A tx not included in 3 blocks will be considered an error, and will trigger a bump of the fee, with the rules on ```get_bumped_gas_price```
    /// This will do 5 bumps every 3 blocks, and then the exponential backoff will dominate, doing bumps at 8,13,24,45,89 and so on
   /// Errors on get_gas_price are considered transient, so this function will keep going
```

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1232,25 +1250,63 @@ impl Batcher {
...
+                {
+                    Ok(gas_price) => gas_price,
+                    Err(e) => return Err(RetryError::Transient(e)),
+                };
 
-                Ok(receipt)
-            }
-        }
+                let bumped_gas_price =
+                    get_bumped_gas_price(*previous_gas_price, current_gas_price, *iteration);
```
```suggestion
                    get_bumped_gas_price(*previous_gas_price, current_gas_price, *iteration);
```

Change name to calculate_bumped_gas_price()

avilagaston9: Done!

batcher/aligned-batcher/src/lib.rs
MauroToscano: Solved

Oppen (approved):

MauroToscano (approved):

Issue #919, opened 2024-09-03
feat: Batcher Client should inform the user if his batch was reverted


uri-99:
Currently if a batch is submitted, but reverts in the batcher tx, the information the client receives seems like his batch was submitted accordingly:

Client:
<img width="1280" alt="image" src="https://github.com/user-attachments/assets/70c19e63-d956-4631-91e5-713d09b5f661">

Batcher:
<img width="1209" alt="image" src="https://github.com/user-attachments/assets/ec0821d9-6a88-4c52-bcb9-58eab7cb90c7">

Issue #1314, opened 2024-10-23
fix: sender receive BatchInclusionData or CreateNewTaskError in ValidityResponseMessage


uri-99:
This error was found under heavy stress testing load;

If a sender is sending many proofs, the batcher will eventually close a batch with his proofs while still receiving new ones. This means the batcher will then send the response (BatchInclusionData or CreateNewTaskError) in the same websocket the sender is still using to send proofs. This makes the sender revert with serializationError.

To fix this, the sender should be able to handle this 2 response messages (BatchInclusionData and CreateNewTaskError) when expecting a ValidityResponseMessage. 
To do this, ValidityResponseMessage should implement these 2 new implementations.

JuArce:
Addressed in #1325

PR #1325, opened 2024-10-25
fix: consolidate 2-step listen after sending proofs

uri-99:
# Consolidate 2-step listen after sending proofs

## Description

This PR fixes a bug found when stress testing the batcher. The batcher would respond with `BatchInclusionData` to a websocket that was still sending proofs, expecting more `ValidityResponseMessage`s . This case errored with an undescriptive `Serialization error`

Update: The fix to this issue will be to remove the 2-step listening from the sender side, to only listen 1 message type, either proof rejected by x reason or proof is valid and included in x merkle_root.

These 2 steps are sequencial (first send the burst of proofs, then read their responses).

This PR also refactors some of the nearby code, hopefully making it more understandable.

## Type of change

- [x] Bug fix
- [x] Refactor

## Checklist
- [x] Linked to Github Issue #1314 
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

## To Test

- Make sure everything works normally
- it is important to send different proofs in different ways, because this refactors how the senders sends the proofs
- for example, sending different bursts
  - `batcher_send_sp1_burst`
  - `batcher_send_risc0_burst`
  - `batcher_send_plonk_bn254_burst`
  - `batcher_send_plonk_bls12_381_burst`
  - `batcher_send_burst_groth16`
- sending by individual tasks
  - `batcher_send_sp1_task`
  - `batcher_send_risc0_task` or `batcher_send_risc0_task_no_pub_input`
  - `batcher_send_plonk_bn254_task`
  - `batcher_send_plonk_bls12_381_task`
  - `batcher_send_groth16_bn254_task`
- sending infinite
  - `batcher_send_infinite_sp1`
  - `batcher_send_infinite_groth16`

## To Deploy
- SDK
  - Batcher
  - CLI (minor error format change from SDK)


batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,198 @@ pub async fn send_messages(
...
-            verification_data_commitments_rev,
-            num_responses.clone(),
-        )
-        .await?;
 
-        if *num_responses.lock().await == total_messages {
-            debug!("All messages responded. Closing connection...");
-            ws_write.lock().await.close().await?;
-            return Ok(aligned_verification_data);
+        let batch_inclusion_data_message = handle_batcher_response(msg).await?;
```
JulianVentura: Here if one response is an error we are returning from this function. Couldn't other responses be valid?

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,198 @@ pub async fn send_messages(
...
+            &mut sent_verification_data,
+        )?;
+
+        let aligned_verification_data =
+            process_batcher_response(&batch_inclusion_data_message, &related_verification_data)?;
+
+        aligned_submitted_data.push(aligned_verification_data);
+        info!("Message response handled succesfully");
+
+        if batch_inclusion_data_message.user_nonce == last_proof_nonce {
```
JulianVentura, line 160: How are we sure we should quit this loop once we receive the last proof nonce? Are we sure we receive the responses from batcher in nonce-order? Why don't we quit after receiving the N pending responses, I think that's what we want to do. With that we could save the `last_proof_nonce` calculation.

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,198 @@ pub async fn send_messages(
...
+// Used to match the message received from the batcher,
+// with the NoncedVerificationData you sent
+// This is used to verify the proof you sent was indeed included in the batch
+fn match_batcher_response_with_stored_verification_data(
+    batch_inclusion_data: &BatchInclusionData,
+    sent_verification_data: &mut Vec<NoncedVerificationData>,
+) -> Result<VerificationDataCommitment, SubmitError> {
+    debug!("Matching verification data with batcher response ...");
+    let mut index = None;
+    for (i, sent_nonced_verification_data) in sent_verification_data.iter_mut().enumerate() {
```
JulianVentura: This is ok, but we should add an issue and leave a TODO here to improve this in the future. We should be able to sort the `sent_verification_data` by nonce and do a binary search over it in order to match the inclusion data. If I'm not mistaken, we don't even need to sort the `sent_verification_data` array, since it is constructed after sending each request sequentially

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,198 @@ pub async fn send_messages(
...
     // Responses are filtered to only admit binary or close messages.
     let mut response_stream = response_stream.lock().await;
+    let mut aligned_submitted_data: Vec<AlignedVerificationData> = Vec::new();
+    let last_proof_nonce = get_biggest_nonce(&sent_verification_data);
 
-    let mut aligned_verification_data: Vec<AlignedVerificationData> = Vec::new();
-
+    // read from WS
     while let Some(Ok(msg)) = response_stream.next().await {
+        // unexpected WS close:
```
Oppen, line 92: Let's log instead.

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,198 @@ pub async fn send_messages(
...
+        return Ok(verification_data.verification_data.clone().into());
+    }
+
+    Err(SubmitError::InvalidProofInclusionData)
+}
+
+// Returns the biggest nonce from the sent verification data
+// Used to know which is the last proof sent to the Batcher,
+// to know when to stop reading the WS for responses
+fn get_biggest_nonce(sent_verification_data: &[NoncedVerificationData]) -> U256 {
```
JulianVentura: We should be able to get the biggest nonce from the latest element, right?

Oppen (approved):

Oppen:
LGTM. Will test later.

JulianVentura (commented):
I'm having some issues while testing.
Started testing bursts and then a combination of tasks. Got an "Error while deleting nonce file" on `batcher_send_groth16_bn254_task` and from that point all tasks are failing. Seems that some nonce state got corrupted.
I added a dbg to the `handle_submit_err()` function and I'm getting an `InvalidProofInclusionData` error. I tried nuking all docker containers and restarting all over and got the same error.

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: We should quit when we receive the las proof nonce because if that one was included we can safely assume the previous ones were included as well.

Also, we don't quit after receiving N messages because we can't know how many responses we actually expect. 
For example, if i send N messages with the same nonce, in which case I will (correctly) receive only 1 response because internally the batcher has interpreted them as replacement messages. This also means I can be sending 3 messages with different nonces, but the nonces were already in the batch queue, so they also end up being replacement messages.

Currently it seems there is no way of knowing exactly how many messages you will end up receiving.

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: Good comment. Other valid proofs sent in the same burst will be deducted from the user's balance. So, the user should have access to the verified ones

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: At the same time, I agree it would be nice to catch any responses if they arrive after the last nonce (which in theory it shouldn't happen)

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: sent_verification_data is already sorted by nonce, yes. Also incoming batch_inclusion_data are theoretically also in nonce-order (batcher sends in order of its inclusion in a batch, which is always processed in nonce-order, and all are sent in the same TCP WS so they should always be arriving in the order they were sent). So iterating in this order ends up being more efficient.

uri-99: Anyhow, the swap_remove could be changed to a `if element is the first one, do pop instead of swap_remove` for more efficiency

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,198 @@ pub async fn send_messages(
...
+        let related_verification_data = match_batcher_response_with_stored_verification_data(
+            &batch_inclusion_data_message,
+            &mut sent_verification_data,
+        )?;
+
+        let aligned_verification_data =
+            process_batcher_response(&batch_inclusion_data_message, &related_verification_data)?;
+
+        aligned_submitted_data.push(aligned_verification_data);
+        info!("Message response handled succesfully");
```
MarcosNicolau: I would avoid this print and only leave the response data location, because when sending a bunch of proofs the terminal looks very bloated:

Screenshot 2024-11-06 at 3 34 55 PM

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,198 @@ pub async fn send_messages(
...
-            }
-        };
+        debug!("{:?} Message sent", idx);
 
-        sent_verification_data.push(verification_data.clone());
+        // Save the verification data commitment to read its response later
+        sent_verification_data.push(verification_data);
     }
 
+    info!("All messages sent");
```
MarcosNicolau: Same with this one, I would change it to `proof sent`.

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,198 @@ pub async fn send_messages(
...
+
+        if batch_inclusion_data_message.user_nonce == last_proof_nonce {
+            break;
         }
     }
 
-    Err(SubmitError::GenericError(
-        "Connection was closed without close message before receiving all messages".to_string(),
-    ))
+    debug!("All message responses handled succesfully");
```
MarcosNicolau:
```suggestion
    debug!("All message responses handled successfully");
```

MarcosNicolau (approved):
Small nit comments. I tested many cases and all seem to work fine. Also, I see the `CreateNewTask` and `BatchUpload` errors are now being displayed in the console, This closes #919.

batcher/aligned-sdk/src/communication/messaging.rs
JulianVentura: Ohh ok, I thought we were always receiving a response from the batcher, even indicating replacements or errors.
Yes, if this is not a 1-1 relationship, then we can't be sure of how many messages we are expecting.
If we can be 100% sure that once a proof gets included, all proofs with lower nonce must have been included, then I'm ok with the implementation. Anyways, I have a question related with this: [Here](https://github.com/yetanotherco/aligned_layer/blob/1314-fix-sender-receive-batchinclusiondata-or-createnewtaskerror-in-validityresponsemessage/batcher/aligned/src/main.rs#L345) we are submitting multiple proofs and then saving the results. If we stop executing when we find the last proof nonce we will miss the intermediate proof results!

batcher/aligned-sdk/src/communication/messaging.rs
JulianVentura: But, if we only care for the last nonce we should iterate the vec in reverse order, filtering any errors until we find one successful submit. That will ensure us all previous proofs were also submitted (if I understood your other comment correctly)

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: Talking with Mauro the result will be a Vec<Result()> , so it has the successful responses in Ok(response) , and errors in Error(e)

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,198 @@ pub async fn send_messages(
...
+            error!("Error while deserializing batch inclusion data: {}", e);
+            Err(SubmitError::SerializationError(e))
+        }
+    }
+}
+
+// Used to match the message received from the batcher,
+// with the NoncedVerificationData you sent
+// This is used to verify the proof you sent was indeed included in the batch
+fn match_batcher_response_with_stored_verification_data(
```
MarcosNicolau, line 261: This function has an issue with the non-paying wallet. As the bather will include the non-paying wallet nonce, and so the verification_data won't be found returning `Err(SubmitError::InvalidProofInclusionData)`

uri-99:
converted to draft until some PR comment issues are addressed

batcher/aligned-sdk/src/communication/messaging.rs
avilagaston9: I agree with Juli. The last element of `sent_verification_data` is the one with the lowest nonce, so we should iterate over it in reverse order. In that case, we wouldn't need `swap_remove` since removing the last element has the same complexity as `pop()`

uri-99: Actually `.push()` adds at the end of the array , same as `.pop()`. I will reverse the array to avoid `swap_remove` while taking advantage of the low cost of removing from the end of the array.
If i find an easy way to iterate in reverse i will implement it as well.
Edit: done

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: yes but not necesarrily. replacement messages can come after the biggest one, and they will be responsded in nonce order

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: The last message the user receives as response is the one of biggest nonce, because they are processed and responded in nonce order, and sent via a TCP WS

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: changed to "all proofs sent", bc its plural

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: done

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,235 @@ pub async fn send_messages(
...
-                return Err(SubmitError::GenericError(
-                    "Connection was closed without close message before receiving all messages"
-                        .to_string(),
-                ));
+        let msg_bin = match cbor_serialize(&msg).map_err(SubmitError::SerializationError) {
+            Ok(bin) => bin,
+            Err(e) => {
+                error!("Error while serializing message: {:?}", e);
+                sent_verification_data.push(Err(e));
+                return sent_verification_data;
```
avilagaston9: Maybe we can avoid the `map_err` here

batcher/aligned-sdk/src/sdk.rs
```diff
@@ -86,22 +85,33 @@ pub async fn submit_multiple_and_wait_verification(
...
+        if let Ok(aligned_verification_data_item) = aligned_verification_data_item {
+            if let Err(e) = await_batch_verification(aligned_verification_data_item, eth_rpc_url, network).await {
+                error_awaiting_batch_verification = Some(e);
+                break;
+            }
+        }
+    }
+    if let Some(error_awaiting_batch_verification) = error_awaiting_batch_verification {
+        aligned_verification_data.push(Err(error_awaiting_batch_verification));
     }
```
avilagaston9, line 112: IMO, we should return a `Vec<Result<AlignedVerificationData>>` containing all the proofs sent by the user. However, `Ok()` should indicate that the verification data was actually verified, not simply sent. In the case of an `Err()`, we should include additional information about the `VerificationData`. For example, if the error is a timeout, the user could decide whether to verify the batch inclusion manually.

JulianVentura:
> I'm having some issues while testing. Started testing bursts and then a combination of tasks. Got an "Error while deleting nonce file" on `batcher_send_groth16_bn254_task` and from that point all tasks are failing. Seems that some nonce state got corrupted. I added a dbg to the `handle_submit_err()` function and I'm getting an `InvalidProofInclusionData` error. I tried nuking all docker containers and restarting all over and got the same error.

Tested again, still having same issues. This time it broke doing a `make batcher_send_burst_groth16`

batcher/aligned-sdk/src/communication/messaging.rs
uri-99: handled in commit `e513997` , it worked for me, please make sure to test it again

uri-99:
> > I'm having some issues while testing. Started testing bursts and then a combination of tasks. Got an "Error while deleting nonce file" on `batcher_send_groth16_bn254_task` and from that point all tasks are failing. Seems that some nonce state got corrupted. I added a dbg to the `handle_submit_err()` function and I'm getting an `InvalidProofInclusionData` error. I tried nuking all docker containers and restarting all over and got the same error.
> 
> Tested again, still having same issues. This time it broke doing a `make batcher_send_burst_groth16`

I think it was related to the case @MarcosNicolau pointed out, which i commited a fix for. please try again

batcher/aligned-sdk/src/sdk.rs
uri-99: This could be addressed in another PR, since it falls out of scope of this one.
Please make an issue of it before resolving this conversation

avilagaston9: Created #1393!

avilagaston9 (approved):
Works in my machine!
Just found a bug when running `make batcher_send_risc0_task_no_pub_input`:
```
error: unexpected argument '--payment_service_addr' found
```
But it works when removing that flag

JulianVentura (approved):
Tested again and it works with the last commit, thank you!

batcher/aligned-sdk/src/communication/messaging.rs
```diff
@@ -55,178 +55,248 @@ pub async fn send_messages(
...
-                return Err(SubmitError::GenericError(
-                    "Connection was closed without close message before receiving all messages"
-                        .to_string(),
-                ));
+        let msg_bin = match cbor_serialize(&msg) {
+            Ok(bin) => bin,
+            Err(e) => {
+                error!("Error while serializing message: {:?}", e);
+                sent_verification_data.push(Err(SubmitError::SerializationError(e)));
+                return sent_verification_data;
```
Oppen, line 64: For a later improvement (to not hold this one back): we might serialize first in a loop, fail the whole thing in that case, and then after all serialization succeeds we can try to send everything, possibly "waiting in parallel" as the async model promotes.

PR #1327, opened 2024-10-25
refactor(operator): use url.Parse instead of regex for sending rpc information to telemetry service

### LINKED ISSUE #1326: refator(operator): use built in url parser instead of regex for sending rpc information to telemetry service

JuArce:
# Use url.Parse instead of regex for sending rpc information to telemetry service

## Description

The previous version was using a regex to get the host of the RPC provider. This makes difficult to be sure it covers al cases and if the process fails, regexp.MustCompile panics

The new version uses the url.Parse from the Go net library (https://gobyexample.com/url-parsing). This way the process of parse urls is more robust and safe.

## Type of change

- [x] Refactor

## How to Test

1. `cd operator/pkg`
2. `go test`

## Checklist

- Unit tests added

### DIFF
--- a/operator/pkg/utils.go
+++ b/operator/pkg/utils.go
@@ -1,10 +1,9 @@
 package operator
 
 import (
-	"fmt"
 	"github.com/yetanotherco/aligned_layer/common"
 	"math/big"
-	"regexp"
+	"net/url"
 )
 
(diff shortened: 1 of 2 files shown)


MarcosNicolau (approved):
Tested on linux and macos.

entropidelic (approved):

Oppen (approved):

uri-99 (approved):

PR #1335, opened 2024-10-25
feat: Telemetry for the batcher and improve current tracing

### LINKED ISSUE #1322: feat: add telemetry for the batcher and improve current tracing

IAvecilla:
# Telemetry integration for the Batcher

## Description

This PR includes the telemetry integration on the batcher side to log the start of a task and some extra steps between adding more visibility on the different parts that process a given batch.
Now we include what are called `subspans` that are intermediate logs that will pop up instantly on the Jaeger dashboard, now we are able to track on which part of the process a task failed or get lost.

## How to test

- Setup all the aligned components (anvil, aggregator, operator and batcher)
- Run the telemetry module
   - `make telemetry_create_env`
   - `make telemetry_run_db`
   - `make open_telemetry_start`
   - `make telemetry_start`
- Send proofs using `make batcher_send_burst_groth16`
- Open the Jaeger dashboard in http://localhost:16686
- You should see new logs for the batcher like:
image
- Also you should see new logs for the aggregator
image

You can take down the operator to check that also the expired tasks should be logged in the aggregator span after some minutes.

## Type of change

- [x] New feature

## Checklist

- [x] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


MarcosNicolau (approved):
Code looks very good. Tested all traces on my machine and everything worked as expected.

telemetry_api/lib/telemetry_api/traces.ex
```diff
@@ -23,31 +23,29 @@ defmodule TelemetryApi.Traces do
...
+            %{
+              attributes: [
+                {:merkle_root, merkle_root},
+                {:total_stake, total_stake}
+              ]
+            }
+          )
+
+        Tracer.set_current_span(aggregator_subspan_ctx)
+        Tracer.add_event("New task taken", [])
```
MauroToscano:
```suggestion
        Tracer.add_event("New task received", [])
```

JuArce (commented):
Would be nice a How to test section and a section with which entities requires a new deploy so is easier to prepare the release

batcher/aligned-batcher/src/lib.rs
```diff
@@ -220,6 +223,11 @@ impl Batcher {
         }
         .expect("Failed to get disabled verifiers");
 
+        let telemetry = TelemetrySender::new(format!(
+            "http://{}",
```
JuArce, line 219: Protocol is not necessary

JuArce (changes requested):
Since this PR includes tracing improvements, I would add the following improvements
- Add an event in the aggregator span when the task finally failed or is expired
- Add a event of the Ethereum transaction in the aggregator txs
-

batcher/aligned-batcher/src/lib.rs
IAvecilla: I think it is needed it will use the port specified in the config but the protocol is needed to communicate with the telemetry api

Oppen (commented):
Nit: most of the `if let Err(e) ...` case can be replaced by calls to `.inspect_err()`.

JuArce (approved):

MauroToscano (approved):

PR #1351, opened 2024-11-04
fix(aggregator): (WIP) fetch task on task response if not cached

### LINKED ISSUE #1350: bug: aggregator misses batches while down
Batches created while the aggregator is down will be lost, since it doesn't have a way to recover missed batches.

JulianVentura:
# Aggregator fetch missed batches

## Description

This PR adds a fix so the aggregator is able to recover and process missed batches when an unknown response from an operator is received.

## Type of change

- [x] Bug fix

## Checklist

- [x] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [x] Has a known issue
  - #1350 

## Basic flow

On recover, if the aggregator has missed some new batches, the operators will process them and start sending signed responses to the aggregator. The main idea of this PR is that once this happens, the aggregator should first check if the received task is known to him (internal map) and if it isn't, try to fetch if from Ethereum logs. This should be done in a retry fashion, since network may be congested and events may take longer to arrive to certain RPC nodes.

## Retry logic

After a response is received, the aggregator will check for the corresponding task in its internal map with 3 retries, waiting 1 sec, 2 sec, and 4 sec, respectively.
If the task is not found in internal maps, it will try to fetch logs. While doing so, some calls will be made to Ethereum, each of them having its own retry logic:

* Get the current block number: Retry times (3 retries): 1 sec, 2 sec, 4 sec.
* Filter batch: Retry times (3 retries): 1 sec, 2 sec, 4 sec.
* Get batch state: Retry times (3 retries): 12 sec (1 Blocks), 24 sec (2 Blocks), 48 sec (4 Blocks)

## How to test

1. Start all services, including a local explorer and 3 operators.
2. Shut down the aggregator.
3. Start sending proofs with `make batcher_send_infinite_sp1`.
4. Wait a few minutes, see how the batches start to accumulate on the explorer as `Pending`, the more you wait, the better.
5. Start the aggregator.
6. You should see how some of the missed batches get verified, as well as the new ones.

You should bump the operator `MaxRetries` under `rpc_client.go` so the operators keep retrying sending responses to the aggregator.

You may also modify the `pending_batch_fetch_block_range` config value under `config-files/config-aggregator.yaml` to test for different scenarios. On the same file, you may also bump `garbage_collector_period` or `garbage_collector_tasks_age` so batches are not removed before they get verified.

We should stress test this PR so we are sure that no concurrency bug is possible, and for that you should try with the following:

* Use as many operators as possible
* When aggregator is down, wait as much time as possible, so many batches accumulate ~50 or more.
* Modify the operator's retries variables so they retry indefinitely and very rapidly (try with wait times below 1 second).
* Set the `pending_batch_fetch_block_range` variable to a large number, so all batches are fetched.

Closes #1350


aggregator/internal/pkg/server.go
```diff
@@ -35,6 +35,48 @@ func (agg *Aggregator) ServeOperators() error {
...
 }
 
+// Waits for the arrival of task associated with signedTaskResponse and returns true on success or false on failure
+// If the task is not present in the internal map, it will try to fetch it from logs and retry.
+// The number of retries is specified by `waitForEventRetries`, and the waiting time between each by `waitForEventSleepSeconds`
+func (agg *Aggregator) waitForTask(signedTaskResponse *types.SignedTaskResponse) bool {
+	for i := 0; i < waitForEventRetries; i++ {
+		// Lock
+		agg.taskMutex.Lock()
+		agg.AggregatorConfig.BaseConfig.Logger.Info("- Locked Resources: Starting processing of Response")
```
MarcosNicolau: This log is repeated on `ProcessOperatorSignedTaskResponseV2`.

aggregator/internal/pkg/server.go
```diff
@@ -49,27 +91,25 @@ func (agg *Aggregator) ProcessOperatorSignedTaskResponseV2(signedTaskResponse *t
...
-		agg.AggregatorConfig.BaseConfig.Logger.Info("- Locked Resources: Starting processing of Response")
-		taskIndex, ok = agg.batchesIdxByIdentifierHash[signedTaskResponse.BatchIdentifierHash]
-		if !ok {
-			agg.taskMutex.Unlock()
-			agg.logger.Info("- Unlocked Resources: Task not found in the internal map")
-			time.Sleep(waitForEventSleepSeconds)
-		} else {
-			break
-		}
+	if !agg.waitForTask(signedTaskResponse) {
```
MarcosNicolau: I find `waitForTask` a bit weird primarily because it mutates the map and it isn't very clear. Personally I would prefer to consolidate everything into a single function (`getTask` for example). Or at least, renaming it to something more descriptive, such as `verifyTaskInMap` or `fetchTask`.

aggregator/internal/pkg/server.go
```diff
@@ -35,6 +35,48 @@ func (agg *Aggregator) ServeOperators() error {
...
+// If the task is not present in the internal map, it will try to fetch it from logs and retry.
+// The number of retries is specified by `waitForEventRetries`, and the waiting time between each by `waitForEventSleepSeconds`
+func (agg *Aggregator) waitForTask(signedTaskResponse *types.SignedTaskResponse) bool {
+	for i := 0; i < waitForEventRetries; i++ {
+		// Lock
+		agg.taskMutex.Lock()
+		agg.AggregatorConfig.BaseConfig.Logger.Info("- Locked Resources: Starting processing of Response")
+		_, ok := agg.batchesIdxByIdentifierHash[signedTaskResponse.BatchIdentifierHash]
+		// Unlock
+		agg.logger.Info("- Unlocked Resources: Task not found in the internal map")
```
MarcosNicolau: Ditto.

core/chainio/avs_reader.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/Layr-Labs/eigensdk-go/logging"
 )
 
+const (
+	BatchFetchBlocksRange uint64 = 1000
```
MarcosNicolau: I am curious, is this just a random number or it has some logic behind?

core/chainio/avs_reader.go
```diff
@@ -150,3 +154,47 @@ func (r *AvsReader) GetOldTaskHash(nBlocksOld uint64, interval uint64) (*[32]byt
 	batchIdentifierHash := *(*[32]byte)(crypto.Keccak256(batchIdentifier))
 	return &batchIdentifierHash, nil
 }
+
+// Returns a pending batch from its merkle root or nil if it doesn't exist
+// Searches the last `BatchFetchBlocksRange` blocks at most
+func (r *AvsReader) GetPendingBatchFromMerkleRoot(merkleRoot [32]byte) (*servicemanager.ContractAlignedLayerServiceManagerNewBatchV3, error) {
```
MarcosNicolau: I can imagine this function being useful in the future. I would refactor it and remove the `responded` check and receive the `fromBlock` as a parameter instead of it being hardcoded. Then the aggregator in `waitForTask` should check for `responded` and calculate the `fromBlock`.

MarcosNicolau (changes requested):
I tested on macos and linux and everything worked fined. I left a few minor comments.

core/chainio/avs_reader.go
JulianVentura: As random as any other number actually 😄. I think it's a good idea to limit it, but may not actually be necessary.

aggregator/internal/pkg/server.go
JulianVentura: Yes, I can change it if that's more clear

core/chainio/avs_reader.go
MarcosNicolau: Yeah, `1000` is quite the number but since we are using our own rpc we won't have problems.

aggregator/internal/pkg/server.go
JulianVentura: What do you mean by consolidating everything in one function? You mean deleting `waitForTask` and including its contents inside `ProcessOperatorSignedTaskResponseV2`? I think I prefer to have an auxiliary function.
I also find `waitForTask` not very descriptive either, but the same happens with other names. I'll keep thinking an alternative, `tryToFetchTask` is my first candidate.

core/chainio/avs_reader.go
JulianVentura: I think this would be an easy refactor to make when that moment arrives. For now, I think we should keep it as simple as possible.
Besides that, I'm curious. How would you calculate the `fromBlock`? I think we are left with checking all blocks (`fromBlock = 0`) or with a random constant as I used (`fromBlock = currentBlock - 1000`).

MarcosNicolau: Yes, you would still have a hardcoded constant, but the function `GetPendingBatchFromMerkleRoot` would take it as a parameter.

aggregator/internal/pkg/server.go
JulianVentura: Done

aggregator/internal/pkg/server.go
JulianVentura: Done

aggregator/internal/pkg/server.go
JulianVentura: Done

MarcosNicolau (approved):

aggregator/internal/pkg/server.go
```diff
@@ -35,6 +35,48 @@ func (agg *Aggregator) ServeOperators() error {
...
+			agg.logger.Warn("Error fetching task from logs: %v", err)
+		}
+
+		if batch == nil {
+			agg.logger.Info("Task not found in logs")
+		}
+
+		// Task was not found, wait and retry
+		time.Sleep(waitForEventSleepSeconds)
+	}
```
Oppen: Why a new separate retry logic here?

JulianVentura: It's not a new separate retry logic, it's the same one but moved to an auxiliary function. You can check [here](https://github.com/yetanotherco/aligned_layer/blob/testnet/aggregator/internal/pkg/server.go#L55) how it was before.
This retry logic is needed because operators may receive the new task event before aggregators. Very unusual, but possible.
We may refactor this in the future once [the retry refactor PR](https://github.com/yetanotherco/aligned_layer/pull/1304/files#diff-566a66135148f53094f82c777b2a3f1a3cab38cdf10feabe0790e2548df58a93R133-R142) is merged.

aggregator/internal/pkg/server.go
```diff
@@ -109,9 +149,6 @@ func (agg *Aggregator) ProcessOperatorSignedTaskResponseV2(signedTaskResponse *t
 		*reply = 0
 	}
 
-	agg.AggregatorConfig.BaseConfig.Logger.Info("- Unlocked Resources: Task response processing finished")
-	agg.taskMutex.Unlock()
```
Oppen, line 113: I can't find the matching lock for this. Was this a bug?

Oppen:
Looks good, need to test later.

aggregator/internal/pkg/server.go
```diff
@@ -35,6 +35,48 @@ func (agg *Aggregator) ServeOperators() error {
 	return err
 }
 
+// Waits for the arrival of task associated with signedTaskResponse and returns true on success or false on failure
+// If the task is not present in the internal map, it will try to fetch it from logs and retry.
+// The number of retries is specified by `waitForEventRetries`, and the waiting time between each by `waitForEventSleepSeconds`
+func (agg *Aggregator) waitForTaskAndFetchIfLost(signedTaskResponse *types.SignedTaskResponse) bool {
```
MauroToscano: This should be splitted, I also changed the name of the PR to reflect more the real issue, since we have another one that does something similar. Keep the old ```waitForTaskResponses```, and simply make a function that fetches a task response if you don't have it when an operator answers.

MauroToscano (changes requested):

aggregator/internal/pkg/server.go
MauroToscano: It also returns a boolean but has no indication of what it means

aggregator/pkg/server.go
```diff
@@ -51,15 +52,28 @@ func (agg *Aggregator) ProcessOperatorSignedTaskResponseV2(signedTaskResponse *t
...
 		"operatorId", hex.EncodeToString(signedTaskResponse.OperatorId[:]))
-	taskIndex := uint32(0)
 
 	taskIndex, err := agg.GetTaskIndex(signedTaskResponse.BatchIdentifierHash)
 
 	if err != nil {
-		agg.logger.Warn("Task not found in the internal map, operator signature will be lost. Batch may not reach quorum")
-		*reply = 1
-		return nil
+		agg.logger.Warn("Task not found in the internal map, might have been missed. Trying to fetch it from logs")
```
MauroToscano:
```suggestion
		agg.logger.Warn("Task not found in the internal map, might have been missed. Trying to fetch task data from Ethereum")
```

aggregator/pkg/server.go
```diff
@@ -51,15 +52,28 @@ func (agg *Aggregator) ProcessOperatorSignedTaskResponseV2(signedTaskResponse *t
...
-		*reply = 1
-		return nil
+		agg.logger.Warn("Task not found in the internal map, might have been missed. Trying to fetch task data from Ethereum")
+		batch, err := agg.avsReader.GetPendingBatchFromMerkleRoot(signedTaskResponse.BatchMerkleRoot, agg.AggregatorConfig.Aggregator.PendingBatchFetchBlockRange)
+		if err != nil || batch == nil {
+			agg.logger.Warnf("Pending task with merkle root 0x%x not found in logs", signedTaskResponse.BatchMerkleRoot)
+			*reply = 1
+			return nil
+		}
+		agg.logger.Info("Task was found in the logs, adding it to the internal map")
```
MauroToscano:
```suggestion
		agg.logger.Info("Task was found in Ethereum, adding it to the internal map")
```

MauroToscano (approved):

MauroToscano:
Need one more test and a review

avilagaston9 (changes requested):
Batches are not being verified after the merge:
image

core/chainio/avs_reader.go
```diff
@@ -150,3 +150,46 @@ func (r *AvsReader) GetOldTaskHash(nBlocksOld uint64, interval uint64) (*[32]byt
 	batchIdentifierHash := *(*[32]byte)(crypto.Keccak256(batchIdentifier))
 	return &batchIdentifierHash, nil
 }
+
+// Returns a pending batch from its merkle root or nil if it doesn't exist
+// Searches the last `blockRange` blocks at most
+func (r *AvsReader) GetPendingBatchFromMerkleRoot(merkleRoot [32]byte, blockRange uint64) (*servicemanager.ContractAlignedLayerServiceManagerNewBatchV3, error) {
```
uri-99, line 156: Sound like it a `lastReadBlock` would be better than a `blockRange`

core/chainio/avs_reader.go
```diff
@@ -150,3 +150,46 @@ func (r *AvsReader) GetOldTaskHash(nBlocksOld uint64, interval uint64) (*[32]byt
...
+	}
+	if err := logs.Error(); err != nil {
+		return nil, err
+	}
+
+	if !logs.Next() {
+		return nil, nil //Not an error, but no tasks found
+	}
+
+	batch := logs.Event
```
uri-99, line 180: `logs` is an iterator, it can have many batches, here i think you are only getting the last (or first) batch in the list.

aggregator/pkg/server.go
```diff
@@ -51,15 +52,28 @@ func (agg *Aggregator) ProcessOperatorSignedTaskResponseV2(signedTaskResponse *t
...
 	if err != nil {
-		agg.logger.Warn("Task not found in the internal map, operator signature will be lost. Batch may not reach quorum")
-		*reply = 1
-		return nil
+		agg.logger.Warn("Task not found in the internal map, might have been missed. Trying to fetch task data from Ethereum")
+		batch, err := agg.avsReader.GetPendingBatchFromMerkleRoot(signedTaskResponse.BatchMerkleRoot, agg.AggregatorConfig.Aggregator.PendingBatchFetchBlockRange)
+		if err != nil || batch == nil {
+			agg.logger.Warnf("Pending task with merkle root 0x%x not found in logs", signedTaskResponse.BatchMerkleRoot)
+			*reply = 1
+			return nil
```
uri-99: In the future we might want to add an error saying this.

aggregator/pkg/server.go
```diff
@@ -51,15 +52,28 @@ func (agg *Aggregator) ProcessOperatorSignedTaskResponseV2(signedTaskResponse *t
...
+		}
+		agg.logger.Info("Task was found in Ethereum, adding it to the internal map")
+		agg.AddNewTask(batch.BatchMerkleRoot, batch.SenderAddress, batch.TaskCreatedBlock)
```
uri-99, line 68: This will probably change when reading all the events read by getLogs

uri-99 (changes requested):

core/chainio/avs_reader.go
uri-99: This could be acceptable in another PR

core/chainio/avs_reader.go
JulianVentura: Yes, that's right, but we are also filtering by `merkleRoot`. Can we get more than one response?

aggregator/pkg/server.go
JulianVentura: Yes, wanted to simplify the code since the logic would be the same, except for the log message.
Do you want me to add a different log message when there is an error?

core/chainio/avs_reader.go
uri-99: My bad! filtering by markleRoot is enough

core/chainio/avs_reader.go
uri-99: Filtering by MerkleRoot makes the blockRange quite more acceptable

aggregator/pkg/server.go
uri-99: I think it is OK for now, i will add a short TODO comment about how we could return the error so maybe the operator finds out his signature is being ignored

uri-99 (approved):

avilagaston9 (approved):
Tested with 3 operators!

MarcosNicolau (changes requested):
It isn't working for me. I've tested it with 4 operators waited `15min` and restarted the aggregator. Only a few tasks were verified, not all of them. Then, I tried sending new tasks with the network working normally and the aggregator wouldn't send the aggregated responses.

The part of the task retrieval does work thou.

uri-99 (changes requested):
still wip. 
also missing: avoid getting logs for same merkle_root in concurrency. otherwise, you will fetch in the log in blockchain once for every operator, even tho they are all of the same merkle_root

PR #1382, opened 2024-11-05
ci: deploy operator with ansible

JuArce:
# Ansible setup for operator

> [!NOTE]  
> Merge after #1384

## Description
 
This PR adds the setup for the operator

The current inventory is set to deploy the operator to `aligned-holesky-stage-1` server. 

1. First create the env files:

```shell
make ansible_operator_create_env
```

2. Then complete the env files.

3. Deploy the operator

```shell
make ansible_operator_deploy INVENTORY=./infra/ansible/stage_inventory.yaml ECDSA_KEYSTORE=<> BLS_KEYSTORE=<>
```

## How to Test

Follow the README instructions

## Type of change

- [x] CI


infra/ansible/playbooks/operator.yaml
```diff
@@ -1,47 +1,181 @@
-- import_playbook: setup.yaml
-- import_playbook: go.yaml
-- import_playbook: rust.yaml
-- import_playbook: eigenlayer-cli.yaml
+- name: Run setup playbook
+  ansible.builtin.import_playbook: setup.yaml
+  vars:
+    host: operator
```
klaus993, line 4: For a future improvement: parametrize this variable so it's not repeated so much

Makefile
```diff
@@ -1024,3 +1028,19 @@ ansible_batcher_deploy: ## Deploy the Batcher. Parameters: INVENTORY, KEYSTORE
...
+
+ansible_operator_deploy: ## Deploy the Operator. Parameters: INVENTORY
+	@if [ -z "$(INVENTORY)" ]  || [ -z "$(ECDSA_KEYSTORE)" ]  || [ -z "$(BLS_KEYSTORE)" ]; then \
+		echo "Error: INVENTORY, ECDSA_KEYSTORE, BLS_KEYSTORE must be set."; \
+		exit 1; \
+	fi
+	@ansible-playbook infra/ansible/playbooks/operator.yaml \
+		-i $(INVENTORY) \
+		-e "ecdsa_keystore_path=$(ECDSA_KEYSTORE)" \
+		-e "bls_keystore_path=$(BLS_KEYSTORE)"
```
klaus993, line 1110: Why are we sending these 2 variables via CLI/env var and not via ini file as the rest of the configs?

JuArce: Those two values are your local paths. In the ini file we set the paths for the remote host. Maybe I should clarify that

infra/ansible/README.md
```diff
@@ -22,4 +26,95 @@ Deploy the Batcher:
...
+## Create ECDSA Keystore
+
+Make sure you have installed:
+
+- [Foundry](https://book.getfoundry.sh/getting-started/installation)
+
+Now you can create the ECDSA keystore using the following command:
+
+```shell
+cast wallet new .
```
MarcosNicolau, line 146: I would put the eigen command here to create the wallet as the `eigenlayercli` expects for the eigenlayer type of keystore. That or add the command to import `eigenlayer` from an existing privatekey: 

```shell
 eigenlayer operator keys import --key-type <TYPE> <KEY_NAME> <PRIV_KEY>
```

infra/ansible/playbooks/ini/config-register-operator.ini.example
MarcosNicolau, line 1: ditto

infra/ansible/playbooks/ini/config-operator.ini.example
MarcosNicolau, line 1: I would an explanation for each field.

MarcosNicolau (approved):
Small nit comments. Everything worked alright. I believe a small comment for each of the fields in the config would be great. I had a hard time setting them up because I didn't understand where they should point to.

infra/ansible/playbooks/operator.yaml
```diff
@@ -1,47 +1,181 @@
...
         rule: allow
         port: 9092
         proto: tcp
-        src: 100.0.0.0/24
\ No newline at end of file
+        src: 100.0.0.0/24
+      vars:
+        ansible_ssh_user: "{{ admin_user }}"
+
+    - name: Build the Operator
```
MarcosNicolau: Build operator should be first above `Deposit into wETH strategy`

JuArce: Done

infra/ansible/README.md
JuArce: Done

MarcosNicolau (approved):

klaus993 (approved):
It's really important that in the short term we also parametrize the playbook to send it the stage which we want to deploy (stage, testnet or mainnet) so default values can be set automatically.
Right now the ini files need to be populated manually and many of the values can be easily filled in automatically according to the stage (the values shown in the table of the README).
We need to have as minimal setup as possible so it's easier if we're in a hurry, and these values I'm talking about are stuff that doesn't change, so it should be really easy to set them automatically.

PR #1392, opened 2024-11-08
feat(aggregator): Add more unit tests for retryable logic.

### LINKED ISSUE #1391: feat(aggregator): Add additional unit tests for aggregator retries
Currently the unit tests for the aggregator retries only test the returned value and effects of all retries being provided. We want to also validate the intermediate states that can occur over multiple retries. To do so we should first separate the external logic of each `Retryable` from the retry library by declaring them as separate functions for testing purposes and create tests for these functions following a similar pattern to the current unit tests.

PatStiles:
# TITLE

## Description

closes #1391 

Adds additional unit tests for the retryable logic to the aggregator. Should be merged after #1304 

## Type of change

- [x] New feature

## Checklist

- [x] Linked to Github Issue
- [x] Unit tests added
- [ ] Has a known issue
  - [Link](https://github.com/yetanotherco/aligned_layer/issues/1391)


MarcosNicolau:
This should point to #1304, right?

PatStiles:
Yes! Good catch!

PR #1399, opened 2024-11-11
refactor(aggregator): move `timeToExpiry` task to config file

### LINKED ISSUE #1207: chore: change `timeToExpiry` parameter passed to `blsAggregationService.InitializeNewTask`
The parameter is set to 100 seconds right now, which is too little
https://github.com/yetanotherco/aligned_layer/blob/477de1d5801b93507427eb7cb8e3504ac55dc617/aggregator/internal/pkg/aggregator.go#L355

we should change to something like 1 week

MarcosNicolau:
## Description

This pr moves the `timeToExpiry` time set when Initializing a new task in the aggregator to the config file and sets it to a week by default.

## Testing

1. Setup a localtestnet with 1 operator.
2. Kill the operator
3. Sends proofs
5. Restart the operator after some time and you should expect the right behavior based on the value you provided in the config file.

## Type of change

- [x] Refactor

## Checklist

- [x] “Hotfix” to testnet, everything else to staging
- [x] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [x] Has a known issue
  - Closes #1207 

### DIFF
--- a/aggregator/internal/pkg/aggregator.go
+++ b/aggregator/internal/pkg/aggregator.go
@@ -215,8 +215,6 @@ func (agg *Aggregator) Start(ctx context.Context) error {
 
 const MaxSentTxRetries = 5
 
-const BLS_AGG_SERVICE_TIMEOUT = 100 * time.Second
-
 func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsAggregationServiceResponse) {
 	agg.taskMutex.Lock()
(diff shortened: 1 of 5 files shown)


MauroToscano (approved):

JuArce:
Just a note. If we set a time too high for the task expiration (bls_service_task_timeout: 168h), we won't see the trace in the telemetry service until the task is expired because the trace is not closed

JuArce (approved):

Issue #1383, opened 2024-11-05
Checks of the amount of money spent by users in the queue are not correct


MauroToscano:
- [x] ```self.check_min_balance``` should use the limit set by the users instead of a constant

- [ ] We should evaluate filtering the batch queue if users unlocked funds, or if they don't have enough money to pay

PR #1404, opened 2024-11-12
Fix: Batcher User Balance Checks

PatStiles:
# Fix: Batcher User Balance Checker

## Description

Replaces the hardcoded constant in https://github.com/yetanotherco/aligned_layer/blob/testnet/batcher/aligned-batcher/src/lib.rs#L607 with the user's `min_fee`. Addresses the first point of #1383 . 

This fix is done by adding `total_fees_in_queue` variable in the UserState struct, and comparing if user has enough balance to pay for the needed amount.

This PR also refactors the infamous `min_fee` variable, to be called `max_fee_limit` / `last_max_fee_limit` 

This PR also adds some metrics of this. dont forget to visualize them in the following tests running
```
make run_metrics
```

#### test 1
- Run a local network and ensure proofs are being validated
- Submit a single proof with no balance in the BatcherPaymentService Contract :
1.) Start a local testnet with an no balance:
2.) Create a wallet with no funds
```sh
cast wallet new <KEYSTORE_PATH>
```
3.)
```bash
echo "Sending SP1 fibonacci task to Batcher..."
cd batcher/aligned/ && cargo run --release -- submit \
		--proving_system SP1 \
		--proof ../../scripts/test_files/sp1/sp1_fibonacci.proof \
		--vm_program ../../scripts/test_files/sp1/sp1_fibonacci.elf \
		--repetitions 1 \
		--proof_generator_addr <WALLET_ADDRESS> \
		--keystore_path <KEYSTORE_PATH> \
		--network devnet
```
ensure the proof submission errors within the batcher with `InsufficientBalance`.

#### test 2
- Submit multiple proofs and not have enough to pay.
1.) Fund the wallet by sending eth from a prefunded anvil account:

```bash
cast send <WALLET_ADDRESS --rpc-url http://localhost:8545 --private-key 2a871d0798f97d79848a013d4936a73bf4cc922c825d33c1cf7073dff6d409c6 --value 1ether
```
2.) Fund the batcher with an insufficient balance for multiple proofs.
```bash
aligned deposit-to-batcher \
--network devnet \ 
--keystore_path <KEYSTORE_PATH> \
--amount 0.0026ether
```
3.) Send one proof, see it succeeds.
```bash
cargo run --release -- submit \
                --proving_system SP1 \
                --proof ../../scripts/test_files/sp1/sp1_fibonacci.proof \
                --vm_program ../../scripts/test_files/sp1/sp1_fibonacci.elf \
                --repetitions 1 \
                --proof_generator_addr <WALLET_ADDRESS \
                --keystore_path <KEYSTORE_PATH> \
                --network devnet
```
4.) Send a burst of 10 or more proofs, observe it submission fails due to insufficient balance.
```bash
cargo run --release -- submit \
                --proving_system SP1 \
                --proof ../../scripts/test_files/sp1/sp1_fibonacci.proof \
                --vm_program ../../scripts/test_files/sp1/sp1_fibonacci.elf \
                --repetitions 10 \
                --proof_generator_addr <WALLET_ADDRESS \
                --keystore_path <KEYSTORE_PATH> \
                --network devnet
```

## Type of change

- [x] Refactor
- [x] Fix

## Checklist

- [x] Linked to Github Issue
  - [ ] Benchmarks added/run


batcher/aligned-batcher/src/lib.rs
```diff
@@ -672,7 +672,20 @@ impl Batcher {
...
-        if !self.check_min_balance(proofs_in_batch + 1, user_balance) {
+        let msg_max_fee = nonced_verification_data.max_fee;
+        let Some(user_min_fee) = batch_state_lock.get_user_min_fee(&addr).await else {
+            std::mem::drop(batch_state_lock);
+            send_message(
+                ws_conn_sink.clone(),
+                SubmitProofResponseMessage::InvalidNonce,
+            )
+            .await;
+            self.metrics.user_error(&["invalid_nonce", ""]);
```
MarcosNicolau: Is the err alright? shouldn't it be something related to the balance or max fee?

PatStiles: This was in staging: https://github.com/yetanotherco/aligned_layer/blob/staging/batcher/aligned-batcher/src/lib.rs#L732. Lets address it in a separate issue/pr.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -672,7 +672,20 @@ impl Batcher {
...
+            return Ok(());
+        };
+
+        // We estimate the min balance to be the minimum balance needed to pay for the
```
```suggestion
        // We estimate the min balance to be the minimum balance needed to pay for the
```

This comment is incommplete

batcher/aligned-batcher/src/lib.rs
```diff
@@ -778,8 +779,20 @@ impl Batcher {
...
-        let min_balance = U256::from(user_proofs_in_batch) * U256::from(MIN_FEE_PER_PROOF);
+    fn check_min_balance(
+        &self,
+        user_min_fee: U256,
+        user_proofs_in_batch: usize,
+        user_balance: U256,
+        user_max_fee: U256,
+    ) -> bool {
+        // If user proof has not been submitted yet its default is U256::max_value().
+        // In this case we check the user `max_fee` can pay for its proof.
```
```suggestion
        // Remember min_fee is the minimum max fee the user has to put in a tx. So, it's also the maximium price that the user will pay for each proof. We use this to upper bound the costs of the user and check if it has enough funds to pay for the proof
```

PatStiles: addressed!

batcher/aligned-batcher/src/lib.rs
PatStiles: addressed!

batcher/aligned-batcher/src/lib.rs
PatStiles: Issue created: https://github.com/yetanotherco/aligned_layer/issues/1412

MauroToscano (approved):

MauroToscano:
Code looks good, may need some extra testing

batcher/aligned-batcher/src/lib.rs
```diff
@@ -781,9 +783,20 @@ impl Batcher {
...
+        user_proofs_in_batch: usize,
+        user_balance: U256,
+        user_max_fee: U256,
+    ) -> bool {
+        // `user_min_fee` is the minimum `max_fee` the user submitted to the batcher and represents the maximium price that the user will pay for each submitted proof. We define 'user_min_fee' as an upper bound for the proof submission cost of the user, and use it to validate the user's balance has enough fund available to pay for all submitted proofs.
+        let mut min_fee = user_min_fee;
+        if user_min_fee == U256::max_value() {
+            min_fee = user_max_fee
+        }
+        let min_balance: U256 = U256::from(user_proofs_in_batch) * min_fee;
```
MarcosNicolau: Maybe I am wrong here but why don't we move the `min_balance` check after the `msg_max_fee` check? This way, in the `min_balance` check we'll verify against the `msg_max_fee` directly (as it will become the new min_fee if everything goes ok), we'll save ourselves this if:

```rust
        if user_min_fee == U256::max_value() {
            min_fee = user_max_fee
        }
```
It is hard to understand that `U256::max_value()` is the default value.

MarcosNicolau (commented):
I tested on macos, everything went fine.

batcher/aligned-batcher/src/lib.rs
PatStiles: My thoughts are to leave this check in as it handles an explicit case that a user's proof has been submitted for the first time not the replacement `max_fee` invariant. I we move this check behind the `msg_max_fee` check we still need to have logic to check and specify that if a proof is submitted for the first time we use it's submitted `max_fee` instead of the user_max_fee. I think it's worth considering verifying that these checks are not redundant, but lets address that in another issue.

uri-99 (changes requested):
the removed var, `MIN_FEE_PER_PROOF`, is no longer used in the codebase.
We should remove it from the constants file as well

uri-99 (changes requested):
the check is not accurate, it is multiplying user's amt_of_proofs x min_fee_per_proof. This is a big wrong estimation, see pic attached:
image

uri-99:
Its working now but I can't approve it because i made this pr

batcher/aligned-batcher/src/types/batch_state.rs
```diff
@@ -117,42 +97,115 @@ impl BatchState {
...
     pub(crate) fn update_user_state(
         &mut self,
         addr: &Address,
         new_nonce: U256,
-        new_min_fee: U256,
+        new_max_fee_limit: U256,
         new_proof_count: usize,
-    ) -> Option<(U256, U256, usize)> {
+        new_total_fees_in_queue: U256,
+    ) -> Option<(U256, U256, usize, U256)> {
```
MarcosNicolau, line 150: Returning an Option with the same values it receives seems redundant and odd. It would be better to return a Result here. If it is too complicated to refactor let's leave it for a future pr.

MauroToscano (approved):

batcher/aligned-batcher/src/lib.rs
```diff
@@ -886,9 +894,24 @@ impl Batcher {
...
+        let updated_max_fee_limit_in_batch = batch_state_lock.get_user_min_fee_in_batch(&addr);
         if batch_state_lock
-            .update_user_min_fee(&addr, updated_min_fee_in_batch)
+            .update_user_max_fee_limit(&addr, updated_max_fee_limit_in_batch)
+            .is_none()
+        {
+            std::mem::drop(batch_state_lock);
+            warn!("User state for address {addr:?} was not present in batcher user states, but it should be");
+            return;
+        };
```
MarcosNicolau: We should send a message to the client if this fails, as his connection won't stop otherwise.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -984,13 +1007,25 @@ impl Batcher {
...
         };
 
+        let Some(current_total_fees_in_queue) = batch_state_lock
+            .get_user_total_fees_in_queue(&proof_submitter_addr)
+            .await
+        else {
+            error!("User state of address {proof_submitter_addr} was not found when trying to update user state. This user state should have been present");
+            std::mem::drop(batch_state_lock);
+            return Err(BatcherError::AddressNotFoundInUserStates(
+                proof_submitter_addr,
```
MarcosNicolau, line 1027: Ditto

MarcosNicolau (approved):
Tested on macos, everything worked fine. Code looks good, I leave some comments that you can address if you want.

batcher/aligned-batcher/src/types/batch_state.rs
uri-99: I agree. I impemented it this way to mimick the other `update_x()` funtions.
I would leave this refactor to another PR, and refactor all of them.
I have added TODO comments of this

MarcosNicolau (approved):

batcher/aligned-batcher/src/lib.rs
uri-99: not here , it returns error, and the caller sends the msg

batcher/aligned-batcher/src/lib.rs
uri-99: done

uri-99 (approved):

uri-99:
1 sec testing something

PR #1407, opened 2024-11-12
fix(contracts): remove `ExceededMaxRespondFee` error

### LINKED ISSUE #1402: Aggregator gas fee should be able to be bumped indefinitely
Aggregator gas fee should be able to be bumped indefinitely

Currenlty it cannot be bumped up to the limit the batcher is willing to pay. 

The contract should be changed to reflect to allow paying more gas price, but withdrawing as much as the batcher is willing to pay.

The aggregator needs to bump indefinitely

MarcosNicolau:
## Description

Removes `ExceededMaxRespondFee` error in `AlignedLayerServiceManager`:
- `ExceededMaxRespondFee` is removed from the Aligned contract and now the aggregator wallet will pay for the difference and the batcher will pay for the `feeLimit` he set. This case happens when the tx cost is higher than the `respondToTaskFeeLimit`.
- Metrics were added for: 
   - A count for the number of times the aggregator has paid for the difference.
   - An accumulated gas paid for the difference in eth.
  
Screenshot 2024-11-13 at 3 30 29 PM

- Now sending `respondToTask` in the aggregator will return a permanent error if and only if either the aggregator or the batcher doesn't have enough funds in their wallets.
 
 ## Testing

1. Modify the bumpedGasPrice percentage to 10000 so that the txCost is bigger than the `batch.RespondToTaskFeeLimit`:
```go
102: bumpedGasPrice := utils.CalculateGasPriceBumpBasedOnRetry(gasPrice, 10000, gasBumpIncrementalPercentage, I)
```
2. Setup a local testnet as usual.
3. Send a batch of proofs.
4. Check with the metrics and logs that the aggregator is paying for the batcher. 
5. Check the Batcher and Aggregator balance before and after the aggregator sends the response, the contract should've charged him with `batch.RespondToTaskFeeLimit`:
Get Agg balance before tx:
```shell
cast balance 0x15d34aaf54267db7d7c367839aaf71a00a2c6a65
```
Get batcher balance before transaction:
(Remember Batcher sends some funds when submitting task! so you must check the funds between the task submittion and the response)
```shell
cast call 0x851356ae760d987E095750cCeb3bC6014560891C "balanceOf(address)(uint256)" 0x7bc06c482dead17c0e297afbc32f6e63d3846650
```
  -  Get balances after tx.
  -  Aggregator should have decreased because he had to pay some of the tx:
  -  The difference in Batcher should be equal to `batch.RespondToTaskFeeLimit`. You can identify the `RespondToTaskFeeLimit` value from the aggregator logs.

Another test:
- you can undo the changes in the contract, re-deploy contracts in anvil, and check that the `RespondToTask` reverts.

## Type of change

- [x] Bug fix

## Checklist

- [x] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [x] Has a known issue
  - [Link to the open issue addressing it](#1402 ) 


contracts/src/core/AlignedLayerServiceManager.sol
```diff
@@ -215,14 +215,12 @@ contract AlignedLayerServiceManager is
...
 
         if (txCost > currentBatch.respondToTaskFeeLimit) {
-            revert ExceededMaxRespondFee(
-                currentBatch.respondToTaskFeeLimit,
-                txCost
-            );
+            // Subtract the txCost from the batcher's balance
+            batchersBalances[senderAddress] -= currentBatch.respondToTaskFeeLimit;
+        } else {
+            batchersBalances[senderAddress] -= txCost;
```
uri-99: Transfer should also be here

uri-99 (changes requested):

core/chainio/avs_writer.go
```diff
@@ -145,29 +144,44 @@ func (w *AvsWriter) SendAggregatedResponse(batchIdentifierHash [32]byte, batchMe
...
-	// At this point, batchState was successfully retrieved
-	// Proceed to check values against RespondToTaskFeeLimit
 	respondToTaskFeeLimit := batchState.RespondToTaskFeeLimit
-	w.logger.Info("Batch RespondToTaskFeeLimit", "RespondToTaskFeeLimit", respondToTaskFeeLimit)
 
-	if respondToTaskFeeLimit.Cmp(simulatedCost) < 0 {
-		return fmt.Errorf("cost of transaction is higher than Batch.RespondToTaskFeeLimit")
+	txCost := new(big.Int).Mul(big.NewInt(int64(tx.Gas())), tx.GasPrice())
+	// todo: 70_000 should come from the config or at least be a constant
+	txCost = txCost.Add(txCost, big.NewInt(70_0000))
```
uri-99: This 70k is hardcoded in the contract, but should not be taken into account here. previous txCost is the actual cost the transaction had.
the 70k is hardcoded in the contract to aproximate a value near the final txCost of the tx.

uri-99 (changes requested):

core/chainio/avs_writer.go
uri-99: fixed

contracts/src/core/AlignedLayerServiceManager.sol
uri-99: fixed

uri-99 (approved):

core/utils/eth_client_utils.go
```diff
@@ -51,6 +51,16 @@ func BytesToQuorumThresholdPercentages(quorumThresholdPercentagesBytes []byte) e
 	return quorumThresholdPercentages
 }
 
+func WeiToEth(wei *big.Int) float64 {
```
Oppen, line 54: I see this is only used for metrics, can't the conversion be handled on the visualization side of those metrics?
This should also not be used for any other purpose other than metrics and logs, because this is a lossy conversion (floats are not good for handling money because of that).

Oppen: Alternatively, why don't we keep the big.Float? That *might* be lossless (but I'd need to check the docs).

MarcosNicolau: Yes, it would be better to handle the conversion from grafana. I'll see if it is possible.

uri-99: @Oppen I agree no one likes floats, but Prometheus' Gauge accepts only float64 type

MauroToscano (approved):

PR #1410, opened 2024-11-12
feat: batcher should not be able to build batches too big

### LINKED ISSUE #1381: feat: batcher should not be able to build batches too big
If a batch is too big (bigger than 3k proofs aprox) , the BatcherPaymentService will end up consuming more gas than what is available on an ETH block.

I would put a max batch size of 1500 proofs. But this number has not been debated yet.
This value should be configurable.

uri-99:
# Batcher max_batch_len limit

## Description

This PR adds a feature on the Batcher that prevents it from making batches too big, that would revert in ETH due to gas cost limitations.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Refactor

## Checklist

- [x] Linked to Github Issue
- [x] Unit tests added
- [x] This change requires new documentation.
  - [ ] Benchmarks added/run
- [ ] Has a known issue

## To Test

You can play around with this new `max_batch_len` variable, set in `config-batcher.yaml` .
For example, set it at 16, then send a burst of 20.
You should see the batcher correctly sending one batch of 16 and another of 4.


MarcosNicolau (approved):
Worked on macos and linux.

batcher/aligned-batcher/src/types/batch_queue.rs
```diff
@@ -147,6 +147,7 @@ pub(crate) fn try_build_batch(
     batch_queue: BatchQueue,
     gas_price: U256,
     max_batch_size: usize,
+    max_batch_len: usize,
```
Oppen: What would be the difference between `max_batch_size` and `max_batch_len` here?

batcher/aligned-batcher/src/types/batch_queue.rs
```diff
@@ -529,4 +533,113 @@ mod test {
...
+            payment_service_addr,
+        );
+        let vd_commitment_1: VerificationDataCommitment = nonced_verification_data_1.clone().into();
+        let entry_1 = BatchQueueEntry::new_for_testing(
+            nonced_verification_data_1,
+            vd_commitment_1,
+            dummy_signature,
+            sender_addr,
+        );
+        let batch_priority_1 = BatchQueueEntryPriority::new(max_fee_1, nonce_1);
```
Oppen, line 581: Maybe just write a little function that does all of this. You could parameterize for nonce only, and add the nonce to a base fee to adapt the fee. Then you can write very short tests for both cases: acceptable and unacceptable lengths.

batcher/aligned-batcher/src/types/batch_queue.rs
uri-99: size is of bytes, len if of amount of proofs,
Note, batch_len is the length of the array of proofs that make up the batch

Oppen: I would either add a comment or rename to something like `max_batch_proofs`. That would remove the ambiguity. The other parameter could be `max_batch_bytes`.

batcher/aligned-batcher/src/types/batch_queue.rs
Oppen: Let's ignore this one since the fix is critical. Might be a later refactor.

Oppen (approved):

batcher/aligned-batcher/src/types/batch_queue.rs
uri-99: renamed

MauroToscano (approved):

MauroToscano (approved):

PR #1484, opened 2024-11-25
doc: warn about address reuse

Oppen:
Add a warning to the operator guide for operators' to make sure they don't
use the same address for earnings and for registering as operators.
Doing so would risk the private key becoming compromised and their funds
being stolen.


PatStiles (approved):

JuArce (approved):

MauroToscano:
Description of the issue is not correct, closing this

PR #1485, opened 2024-11-25
fix: batcher timeout to prevent simple DoS

avilagaston9:
# Batcher timeout to prevent simple DOS

## Description

Handles simple DOS attack to the Batcher.

1. Modifies `listener.accept()` to handle error cases.
2. Adds a timeout to the WebSocket handshake.
3. Adds a timeout to the read of the first user message.

### Test with SDK

1. Edit the SDK code to sleep before sending the `getNonce` message in `batcher/aligned-sdk/src/sdk.rs:590`
```Rust
let msg_bin = cbor_serialize(&msg)
        .map_err(|_| GetNonceError::SerializationError("Failed to serialize msg".to_string()))?;

info!("Sleeping to keep the wb blocked.");
sleep(Duration::from_secs(70)).await;
info!("Waking up");
```
`tokio::time::sleep` and `std::time::Duration` should be imported.

4. Start anvil:
```bash
```
5. Start batcher:
```bash
make batcher_start_local
```
6. Start grafana metrics:
```bash
make run_metrics
```
7. Run the SDK:
```bash
make batcher_send_risc0_task
```

You should be able to see that the Batcher is cancelling the connection in the logs:
```bash
2024-11-25T18:24:50Z WARN aligned_batcher] [127.0.0.1:51905] deadline has elapsed
```

Also, check the **User Error Count** panel in the Batcher Grafana Dashboard in `localhost:3000`:

image

### Using python script

1. Change the file descriptor limit for the PoC (in `Makefile` let’s add `ulimit -n 1000 &&`).

  ```bash
   ...

   batcher_start_local: user_fund_payment_service
   	@echo "Starting Batcher..."
   	@$(MAKE) run_storage &
   	@ulimit -n 200 && cargo run --manifest-path ./batcher/aligned-batcher/Cargo.toml --release -- --config ./config-files/config-batcher.yaml --env-file ./batcher/aligned-batcher/.env.dev

   ...
   ```

2. Create the following script `script_poc.py`:
```python
import asyncio
import websockets

async def attack(target_url, delay, num_connections):
   async def open_connection():
       try:
           async with websockets.connect(target_url) as websocket:
               await asyncio.sleep(delay)
       except Exception as e:
           print(f"Connection failed: {e}")

   tasks = []
   for _ in range(num_connections):
       tasks.append(open_connection())
   await asyncio.gather(*tasks)

target_url = "ws://localhost:8080"
delay = 300000
num_connections = 20000

asyncio.run(attack(target_url, delay, num_connections))
```

3. Run in 3 different terminals:
```Bash
make batcher_start_local
(Wait until the batcher has the port 8080 open) python script_poc.py
```

You should see that the Batcher doesn't finish execution and logs the following:
```bash
2024-11-25T21:27:08Z ERROR aligned_batcher] Couldn't accept new connection: Too many open files (os error 24)
[2024-11-25T21:27:11Z WARN  aligned_batcher] [127.0.0.1:60786] deadline has elapsed
```
Also check that the Batcher can accept new connections after the attack.

_You should also test normal usage._

## Type of change

- [x] Bug fix

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/batcher/aligned-batcher/src/lib.rs
+++ b/batcher/aligned-batcher/src/lib.rs
@@ -11,6 +11,7 @@ use retry::batcher_retryables::{
     get_user_nonce_from_ethereum_retryable, user_balance_is_unlocked_retryable,
 };
 use retry::{retry_function, RetryError};
+use tokio::time::timeout;
 use types::batch_state::BatchState;
 use types::user_state::UserState;
 
@@ -18,6 +19,7 @@ use std::collections::HashMap;
 use std::env;
 use std::net::SocketAddr;
 use std::sync::Arc;
+use std::time::Duration;
 
 use aligned_sdk::core::constants::{
     ADDITIONAL_SUBMISSION_GAS_COST_PER_PROOF, AGGREGATOR_GAS_COST, BUMP_BACKOFF_FACTOR,
@@ -379,8 +381,32 @@ impl Batcher {
             .send(Message::binary(serialized_protocol_version_msg))
             .await?;
 
-        match incoming
-            .try_filter(|msg| future::ready(msg.is_binary()))
+        let mut incoming_filter = incoming.try_filter(|msg| future::ready(msg.is_binary()));
+        let future_msg = incoming_filter.try_next();
+        // timeout to prevent a DOS attack
+        match timeout(Duration::from_secs(5), future_msg).await {
+            Ok(Ok(Some(msg))) => {
+                self.clone().handle_message(msg, outgoing.clone()).await?;
+            }
+            Err(elapsed) => {
+                error!("[{}] {}", &addr, elapsed);
+                self.metrics.open_connections.dec();
+                self.metrics.user_error(&["user_timeout", ""]);
+                return Ok(());
+            }
+            Ok(Ok(None)) => {
+                info!("[{}] Connection closed by the other side", &addr);
+                self.metrics.open_connections.dec();
+                return Ok(());
+            }
+            Ok(Err(e)) => {
+                error!("Unexpected error: {}", e);
+                self.metrics.open_connections.dec();
+                return Ok(());
+            }
+        };
+
+        match incoming_filter
             .try_for_each(|msg| self.clone().handle_message(msg, outgoing.clone()))
             .await
         {
--- a/batcher/aligned-batcher/src/lib.rs
+++ b/batcher/aligned-batcher/src/lib.rs
@@ -23,7 +23,7 @@ use std::time::Duration;
 
 use aligned_sdk::core::constants::{
     ADDITIONAL_SUBMISSION_GAS_COST_PER_PROOF, AGGREGATOR_GAS_COST, BUMP_BACKOFF_FACTOR,
-    BUMP_MAX_RETRIES, BUMP_MAX_RETRY_DELAY, BUMP_MIN_RETRY_DELAY, CONNECTION_READ_TIMEOUT,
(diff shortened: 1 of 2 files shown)


batcher/aligned-sdk/src/core/constants.rs
```diff
@@ -8,6 +8,7 @@ pub const CONSTANT_GAS_COST: u128 =
         + BATCHER_SUBMISSION_BASE_GAS_COST;
 pub const DEFAULT_MAX_FEE_PER_PROOF: u128 =
     ADDITIONAL_SUBMISSION_GAS_COST_PER_PROOF * 100_000_000_000; // gas_price = 100 Gwei = 0.0000001 ether (high gas price)
+pub const CONNECTION_READ_TIMEOUT: u64 = 60; // 1 minute
```
Oppen: I don't think this protection is very strong. During the 60 seconds wait they could open at least a few hundred connections which in turn will also be granted 60 seconds of grace. Even with little distribution that could mean thousands of concurrent connections doing nothing.

avilagaston9: Totally right. Changed to two seconds!

batcher/aligned-batcher/src/lib.rs
```diff
@@ -267,13 +267,18 @@ impl Batcher {
...
-            let batcher = self.clone();
-            tokio::spawn(batcher.handle_connection(stream, addr));
+        loop {
+            match listener.accept().await {
+                Ok((stream, addr)) => {
+                    self.metrics.open_connections.inc();
+                    let batcher = self.clone();
+                    // Let's spawn the handling of each connection in a separate task.
+                    tokio::spawn(batcher.handle_connection(stream, addr));
+                    self.metrics.open_connections.dec();
```
Oppen: Note this will lead to lying metrics. The `tokio::spawn` call is not being waited for, so the next line runs concurrently with the actual connection handling, indicating it finished before that happens.

avilagaston9: Oh, right! Thank you, I will leave it as before.

avilagaston9: Addressed in [664fa44](https://github.com/yetanotherco/aligned_layer/pull/1485/commits/664fa44f2013b688567b5ab5c53dbe37ba0a8f54)

batcher/aligned-sdk/src/core/constants.rs
```diff
@@ -8,6 +8,7 @@ pub const CONSTANT_GAS_COST: u128 =
         + BATCHER_SUBMISSION_BASE_GAS_COST;
 pub const DEFAULT_MAX_FEE_PER_PROOF: u128 =
     ADDITIONAL_SUBMISSION_GAS_COST_PER_PROOF * 100_000_000_000; // gas_price = 100 Gwei = 0.0000001 ether (high gas price)
+pub const CONNECTION_TIMEOUT: u64 = 2; // 2 secs
```
MauroToscano:
```suggestion
pub const CONNECTION_TIMEOUT: u64 = 5; // 5 secs
```

MarcosNicolau (approved):
Works as expected on macos. You could've use:
```shell
make task_sender_test_connections_devnet NUM_SENDERS=<X>
```
to test the dos scenario.

JuArce (approved):

Oppen (approved):

PR #1488, opened 2024-11-25
fix: remove unneeded uses of ECDSA private key

Oppen:
Both the operator and aggregator were using the `BuildAll` helper that required a private key due to creation of signers, only to discard the signers and keep the reader and subscriber.
Changing it to `BuildReadClients` allows removing some of it, in the case of the operator meaning only registration requires the ECDSA key to be present, reducing risks of losing funds.

A later change should update the docs to note the ECDSA key is unnecessary after registration and should be deleted from the operator. Left for later as a bigger restructure of the operator guide is needed.

## Type of change

- [x] Refactor

## Checklist

- [x] This change requires new documentation.
- [x] Known issue: this breaks operator metrics until they start signing their announcement with BLS. @MarcosNicolau is working on that on another PR.


MarcosNicolau (commented):
Everything works fine! We need to update the documentation regarding the management of the `ecdsa` key.

JulianVentura (commented):
Code looks fine and it seems to be working. I'll test again when the telemetry fix is merged

MarcosNicolau (approved):

JuArce (approved):
Approved, but once #1488 it will have merge conflicts

core/config/operator.go
uri-99, line 1: line 48 should also be removed

uri-99 (changes requested):

uri-99 (approved):

PR #1489, opened 2024-11-25
fix: use bls signature in telemetry api for the operator

MarcosNicolau:
## Description

When sending the operator version to the telemetry service we verify the operator signature using the `ECDSA` key. This pr changes it to verify the signature with the operator `BLS` corresponding to the one registered on the aligned avs.

Something to note is that in the `telemetry` message from the operator we provide the `signature` and the `publickey` in the `G2` generator. The last part is needed as the `BLSAPKRegistry` stores the operator `publickey` on `G1` and the signature is also in `G1`. So the verification has two steps:
1. That the operator provided `G2` `publickey` corresponds to its `G1` equivalence registered on ethereum (we perform the pairing check)
2. The actual signature verification using the `G2` pubkey.

Finally, the code might not be so elixir idiomatic (I am open to any criticism to improve it if necessary).

> Note: when building telemetry now we need to run the following:
> ```shell
> make telemetry_compile_bls_verifier
> ```
> This is already done on `make telemetry_full_start`

**Testing**

You can test it by registering an operator, verifying the signature is well verified, and then changing its config file to point to another `bls` keystore, this would fail as the `G1` -> `G2` correspondence of the keys won't match:
1. Start telemetry: `make telemetry_full_start`
2. Start Aggregator: `make aggregator_start`
3. Register operator: `make operator_full_registration CONFIG_FILE=config-files/config-operator-1.yaml`
4. Build operator: `make build_operator_macos`
5. Start operator: `operator/build/aligned-operator start --config config-files/config-operator-1.yaml`
6. Change line `19` in `config-files/config-operator-1.yaml` for the operator 2 bls key: 
```yaml
  private_key_store_path: "config-files/devnet/keys/operator-2.bls.key.json"
```
7. Re start operator (you should get a `401` err): `operator/build/aligned-operator start --config config-files/config-operator-1.yaml`
8. Go to `http://localhost:4001/api/operators` and check telemetry still works.

## Type of change

- [x] Bug fix
- [x] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


MauroToscano (approved):

avilagaston9 (approved):
Code LGTM and It works in my machine!

Oppen:
The Ansible recipe would need to include the signature verifier program IIUC.
@klaus993 @samoht9277 @JuArce  is this correct?

telemetry_api/bls_verifier/bls_verify.go
Oppen, line 1: Beautiful.

operator/pkg/operator.go
```diff
@@ -646,13 +646,12 @@ func (o *Operator) SendTelemetryData(ctx *cli.Context) error {
 	hash.Write([]byte(ctx.App.Version))
 
 	// get hash
-	version := hash.Sum(nil)
+	var version [32]byte // All zeroed initially
+	copy(version[:], hash.Sum(nil))
```
Oppen, line 650: I believe signing only the version makes us susceptible to replay attacks. Can anyone with a firmer grasp of cryptography take a look? @MauroToscano @JuArce 
Note if that's true the issue was already there, it's not introduced by this PR, just preserved.

Oppen: If it's just for counting operators it should not matter, but if it's used to identify for real it might be a problem, right?

MauroToscano: It goes encrypted to the net, and it's for our services. We can fake being an operator and break our telemetry but we won't do that

JulianVentura (approved):
Code looks good and works on my machine!

MauroToscano (changes requested):
Delete old module to verify ECDSA signatures

telemetry_api/lib/telemetry_api/contract_managers/bls_apk_registry.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  {status, config_json_string} = File.read(config_file_path)
+
+  case status do
+    :ok ->
+      Logger.debug("Aligned deployment file read successfully")
+
+    :error ->
+      raise(
+        "Config file not read successfully, did you run make explorer_create_env? If you did,\n make sure Eigenlayer config file is correctly stored"
```
uri-99:
```suggestion
        "Config file not read successfully, make sure your .env is correctly created, and make sure Eigenlayer config file is correctly stored"
```

MauroToscano (approved):

JuArce (approved):

uri-99 (approved):

PR #1558, opened 2024-12-03
fix(sdk): Remove max_fee function from sdk

PatStiles:
# Removes max_fee function from sdk

## Description

Removes a `max_fee` function that computed the total fee for a user submitting a batch of `n` proof into a batch of size `num_proofs_in_batch`. `fee_per_proof` is renamed to `max_fee_per_proof`. Open to renaming suggestions.

## Type of change

- [x] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


batcher/aligned-sdk/src/sdk.rs
```diff
@@ -825,50 +804,3 @@ fn save_response_json(
 
     Ok(())
 }
-
-#[cfg(test)]
```
MarcosNicolau, line 829: Shouldn't we rewrite the tests for `max_fee_per_proof` instead of removing them?

PatStiles: I am in favor of just removing them as they rely on an rpc and could be flaky when used on the ci.

MarcosNicolau (approved):

batcher/aligned-sdk/src/sdk.rs
Oppen: If that's the case, I would suggest a different approach. Just extract the computation to a helper and test the helper. Then the RPC call will be untested by unit tests but it will necessarily be tested on integration ones (as the function is used). The latter don't guarantee we use the right numbers for the right case.

Oppen: It's a bit like mocking but without lying to ourselves about having tested the interaction with the RPC.

Oppen:
How shall we test it?

batcher/aligned-sdk/src/sdk.rs
```diff
@@ -180,7 +159,7 @@ pub async fn compute_max_fee(
 /// # Errors
 /// * `EthereumProviderError` if there is an error in the connection with the RPC provider.
 /// * `EthereumGasPriceError` if there is an error retrieving the Ethereum gas price.
-pub async fn fee_per_proof(
+pub async fn max_fee_per_proof(
     eth_rpc_url: &str,
     num_proofs_per_batch: usize,
```
uri-99, line 164: change to `num_proofs_in_batch`

batcher/aligned-sdk/src/sdk.rs
```diff
@@ -180,7 +159,7 @@ pub async fn compute_max_fee(
 /// # Errors
 /// * `EthereumProviderError` if there is an error in the connection with the RPC provider.
 /// * `EthereumGasPriceError` if there is an error retrieving the Ethereum gas price.
-pub async fn fee_per_proof(
+pub async fn max_fee_per_proof(
```
uri-99, line 162: change to suggest_fee_per_proof

PR #1568, opened 2024-12-04
feat(sdk): some SDK improvements

PatStiles:
# SDK Changes

## Description

Makes some changes to sdk:
- Derives `batcher_url` from Network paramater.
- Adds `--private_key` flag to DepositToBatcher
- Adds msg to batcher errors telling a user there funds have not been spent.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


batcher/aligned/src/main.rs
```diff
@@ -66,9 +66,10 @@ pub struct SubmitArgs {
     #[arg(
         name = "Batcher connection address",
         long = "batcher_url",
+        required = false,
         default_value = "ws://localhost:8080"
     )]
-    batcher_url: String,
+    batcher_url: Option<String>,
```
uri-99: this should no longer be an option. with --network is enough.

batcher/aligned/src/main.rs
```diff
@@ -146,6 +147,8 @@ pub struct DepositToBatcherArgs {
     network: NetworkArg,
     #[arg(name = "Amount to deposit", long = "amount", required = true)]
     amount: String,
+    #[arg(name = "Private key", long = "private_key")]
+    private_key: Option<String>,
```
uri-99, line 140: if this is an option, keystore should not be `required=true`.
also we should only accept 1 of these 2 options

batcher/aligned/src/main.rs
```diff
@@ -207,15 +210,23 @@ pub struct GetUserNonceArgs {
     #[arg(
         name = "Batcher connection address",
         long = "batcher_url",
+        required = false,
         default_value = "ws://localhost:8080"
     )]
```
uri-99: remove

batcher/aligned/src/main.rs
```diff
@@ -279,9 +290,13 @@ async fn main() -> Result<(), AlignedError> {
 
             let max_fee =
                 U256::from_dec_str(&submit_args.max_fee).map_err(|_| SubmitError::InvalidMaxFee)?;
+            let network: Network = submit_args.network.into();
 
             let repetitions = submit_args.repetitions;
-            let connect_addr = submit_args.batcher_url.clone();
+            let batcher_url = submit_args
+                .batcher_url
```
uri-99: no, get url from network argument

batcher/aligned/src/main.rs
MarcosNicolau: But what if somebody has its own batcher?

batcher/aligned/src/main.rs
```diff
@@ -464,12 +479,17 @@ async fn main() -> Result<(), AlignedError> {
...
                 let password = rpassword::prompt_password("Please enter your keystore password:")
                     .map_err(|e| SubmitError::GenericError(e.to_string()))?;
                 Wallet::decrypt_keystore(keystore_path, password)
                     .map_err(|e| SubmitError::GenericError(e.to_string()))?
+            } else if let Some(private_key) = private_key {
+                private_key
+                    .parse::<LocalWallet>()
+                    .map_err(|e| SubmitError::GenericError(e.to_string()))?
             } else {
                 warn!("Missing keystore used for payment.");
```
uri-99: the missing info is not necesarrily a keystore, it can be private key. so text should be like "missing keystore or private key used for payment"

batcher/aligned/src/main.rs
```diff
@@ -464,12 +479,17 @@ async fn main() -> Result<(), AlignedError> {
                 })?;
 
             let keystore_path = &deposit_to_batcher_args.keystore_path;
+            let private_key = &deposit_to_batcher_args.private_key;
```
uri-99, line 469: add here only 1 of these can be set, to avoid user setting conflicting private key AND keystore.

batcher/aligned/src/main.rs
```diff
@@ -518,7 +538,9 @@ async fn main() -> Result<(), AlignedError> {
         }
         GetUserNonce(args) => {
             let address = H160::from_str(&args.address).unwrap();
-            match get_nonce_from_batcher(&args.batcher_url, address).await {
+            let network: Network = args.network.into();
+            let batcher_url = args.batcher_url.unwrap_or(network.get_batcher_url().into());
```
uri-99: no, get batcher url from network arg

uri-99 (changes requested):
left some comments

MarcosNicolau:
@uri-99, your comments have been addressed at: [2e8465c](https://github.com/yetanotherco/aligned_layer/pull/1568/commits/2e8465ca4791c7022d4356fe8f61f97bd7ee43d3).

uri-99 (changes requested):
missing remove batcher-url from rust task sender.

uri-99 (approved):

Oppen (approved):

JulianVentura (approved):
Works as expected!
There are some docs conflicts

batcher/aligned/src/main.rs
```diff
@@ -474,14 +466,23 @@ async fn main() -> Result<(), AlignedError> {
                 })?;
 
             let keystore_path = &deposit_to_batcher_args.keystore_path;
+            let private_key = &deposit_to_batcher_args.private_key;
 
             let mut wallet = if let Some(keystore_path) = keystore_path {
+                if private_key.is_some() {
+                    error!("Conflicting inputs detected: Both a keystore and a private key were provided. Provide only one option to proceed.");
+                    return Ok(());
+                }
```
avilagaston9, line 475: We could have used a group for this: https://stackoverflow.com/questions/76315540/how-do-i-require-one-of-the-two-clap-options

MauroToscano (changes requested):
This PR should be splitted in different PRs. Also, default network params is fine, but we should have a way to change the batcher url manually, let's re review this over the following days

PR #1576, opened 2024-12-04
refactor: improve user fee bumps

### LINKED ISSUE #1571: refactor: improve user fee bumps
The idea is to refactor the bumping mechanism, when a user wants to bump the fee for a proof.

For this in the SDK there should be a function that takes either:
- % to bump
- amount to bump
- new max fee (current implementation
- bump to match a batch of size N

This functions should also be callable via the aligned CLI.

After this bump mechanism is working, the previous bumping mechanism should be deprecated, the mechanism of replacement messages; Replacing a same nonce with a highest fee brings all sorts of problems, from the WS connection, to the known limits to max_fee, and the need for the user to know the nonces he sent (which he never does).

uri-99:
# Improve user fee bumps

## Description

This PR aims to improve the fee bumping process for a user and his proofs.
More details on this PR's [issue](https://github.com/yetanotherco/aligned_layer/issues/1571)

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [x] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/batcher/aligned-batcher/src/lib.rs
+++ b/batcher/aligned-batcher/src/lib.rs
@@ -31,9 +31,7 @@ use aligned_sdk::core::constants::{
     RESPOND_TO_TASK_FEE_LIMIT_PERCENTAGE_MULTIPLIER,
 };
 use aligned_sdk::core::types::{
-    ClientMessage, GetNonceResponseMessage, NoncedVerificationData, ProofInvalidReason,
-    ProvingSystemId, SubmitProofMessage, SubmitProofResponseMessage, VerificationCommitmentBatch,
-    VerificationData, VerificationDataCommitment,
+    BumpFeeResponseMessage, BumpUnit, ClientMessage, GetNonceResponseMessage, NoncedVerificationData, ProofInvalidReason, ProvingSystemId, SubmitProofMessage, SubmitProofResponseMessage, VerificationCommitmentBatch, VerificationData, VerificationDataCommitment
 };
 
 use aws_sdk_s3::client::Client as S3Client;
@@ -473,6 +471,11 @@ impl Batcher {
                     .handle_submit_proof_msg(msg, ws_conn_sink)
                     .await
             }
+            ClientMessage::BumpFee(bump_unit, amout, proof_qty) => {
+                self.clone()
+                    .handle_bump_fee_msg(bump_unit, amout, proof_qty, ws_conn_sink)
+                    .await
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


batcher/aligned-batcher/src/lib.rs
```diff
@@ -473,6 +471,11 @@ impl Batcher {
                     .handle_submit_proof_msg(msg, ws_conn_sink)
                     .await
             }
+            ClientMessage::BumpFee(bump_unit, amout, proof_qty) => {
+                self.clone()
+                    .handle_bump_fee_msg(bump_unit, amout, proof_qty, ws_conn_sink)
+                    .await
+            }
```
Oppen, line 478: I think it would make sense for the unit conversion to happen in the SDK so the batcher can handle just wei. This makes for faster feedback on invalid units (the SDK doesn't need to wait for the batcher to respond to know it was the wrong unit), reduces attack surface for the batcher and makes for fewer things to check.
I'm not 100% sure this covers the batch size and percentual cases OTOH.

Oppen: It would also be more flexible for the user programs that could use different strategies for increasing proposed fees.

batcher/aligned-sdk/src/core/types.rs
```diff
@@ -396,6 +400,86 @@ pub enum GetNonceResponseMessage {
     InvalidRequest(String),
 }
 
+pub struct BumpFeeMessage {
+    pub conent: BumpFeeMessageContent,
```
Oppen, line 404: Typo:
```suggestion
    pub content: BumpFeeMessageContent,
```

batcher/aligned-sdk/src/core/types.rs
```diff
@@ -396,6 +400,86 @@ pub enum GetNonceResponseMessage {
...
+    fn type_hash() -> Result<[u8; 32], Self::Error> {
+        let mut hasher = Keccak256::new();
+        hasher.update(NONCED_VERIFICATION_DATA_TYPE);
+        Ok(hasher.finalize().into())
+    }
+
+    pub fn new(bump_unit: BumpUnit, value: U256, proof_qty: usize) -> Self {
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards");
```
Oppen, line 435: This is actually possible (for how computers process time, not in the physical reality), so I don't recommend crashing. Returning an error if it would be incorrect wouldn't be bad tho.

Oppen: That said, if it went back to before `UNIX_EPOCH` the user has a misconfigured machine, or hit 2038 without updating.

MauroToscano (changes requested):
We need to re review the cryptography, maybe a first version with the old cryptography is better until we simplify it

PR #1578, opened 2024-12-04
feat: add latency metrics to batcher

avilagaston9:
# Add Latency Metrics to Batcher

## Description

Adds the following metrics to the Batcher:
- `S3 Latency`: Measures the time it takes to upload the batch to S3.
- `CreateNewTask Latency`: Measures the time the `CreateNewTask` function takes. It is broken down into two:
    - `create_new_task`: Time it takes to send the new task to Ethereum.
    - `cancel_new_task`: Time it takes to cancel the new task, in case the receipt is not found.

> [!NOTE]
> The time shown is always the latest one. `cancel_new_task` is set to zero since sometimes the function is not executed.

image

> [!NOTE]
> It always takes 7 secs to get a receipt from anvil.

## How to Test

1. Start Anvil:
```bash
```

2. Start the Batcher:
```bash
make batcher_start_local
```

3. Start the metrics:
```bash
make run_metrics
```

4. Start sending proofs:
```bash
```

5. Go to `localhost:3000` and watch the `System Data` dashboard.

You can also edit the Batcher code to emit some `cancel_new_task` events:

1. Edit the `batcher/aligned-batcher/src/lib.rs` file in line `1569` as follows:
```Rust
Ok(receipt) => {
    if let Err(e) = self
        .telemetry
        .task_sent(&hex::encode(batch_merkle_root), receipt.transaction_hash)
        .await
    {
        warn!("Failed to send task status to telemetry: {:?}", e);
    }
    if start.elapsed().as_millis() % 2 == 0 {
        self.cancel_create_new_task_tx(fee_params.gas_price).await;
    }
    Ok(receipt)
}
```

## Type of change

- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


PatStiles (approved):

MarcosNicolau (approved):

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1559,6 +1560,11 @@ impl Batcher {
             ETHEREUM_CALL_MAX_RETRY_DELAY,
         )
         .await;
+        self.metrics
+            .create_new_task_latency
+            .set(start.elapsed().as_millis() as i64);
```
Oppen: The metric should be per attempt I think. We should also count the retries.

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1786,7 +1795,12 @@ impl Batcher {
             ETHEREUM_CALL_MAX_RETRY_DELAY,
         )
         .await
-        .map_err(|e| BatcherError::BatchUploadError(e.to_string()))
+        .map_err(|e| BatcherError::BatchUploadError(e.to_string()))?;
+
+        self.metrics
+            .s3_latency
+            .set(start.elapsed().as_micros() as i64);
```
Oppen: Did we check if our S3 client already provides a metrics exporter?

uri-99: send value even if it fails should visualize a peak in latency

uri-99 (changes requested):
also i would separate the latencies in grafana in a different section

batcher/aligned-batcher/src/lib.rs
JulianVentura: Measurement of the hole function, including retries, was the request. Maybe we should discuss this again.

batcher/aligned-batcher/src/lib.rs
JulianVentura: Why would it be different than a success case after the max number of retries have been performed?
All errors thrown by this function are transient, so it can only return an error at this point if all retries failed. I think the peak would be the same if that last retry were successful instead of a failure.

batcher/aligned-batcher/src/lib.rs
avilagaston9: Maybe we should rename the metric name to "Duration" instead of latency

batcher/aligned-batcher/src/lib.rs
avilagaston9: Created #1583 to review this!

JulianVentura, line 1816: Here you are using `start.elapsed().as_micros()` instead of `start.elapsed().as_millis()` like in the other two gauges. Anyways, it seems to work fine. Is there a reason for that? Is it because the S3 response latency is some orders of magnitude lower than that of the other two methods?

avilagaston9: > Why would it be different than a success case after the max number of retries have been performed?
All errors thrown by this function are transient, so it can only return an error at this point if all retries failed. I think the peak would be the same if that last retry were successful instead of a failure. @JulianVentura 

The problem here is the "?" at the end of the function call. It will never register any time in case of an error. I will fix this.

JulianVentura (approved):
Code looks fine and seems to work as expected. I'm just leaving a comment because of curiosity

batcher/aligned-batcher/src/lib.rs
avilagaston9: Exactly, It was only for better precision.

avilagaston9: Addressed in [#22ab101](https://github.com/yetanotherco/aligned_layer/pull/1578/commits/22ab10100f7cdf7ef89fabbebd3a23759153aa64)!

uri-99 (approved):

batcher/aligned-batcher/src/lib.rs
JuArce: We can add it in another issue

PR #1586, opened 2024-12-06
Add SSL termination to the explorer

klaus993:
# Add SSL termination to the explorer

## Description

Manage SSL termination directly from the explorer, to remove the reverse-proxy dependency (caddy)

## Changes

* `explorer/lib/explorer_web/router.ex`: make CSP less strict for the website to be able to:
  * `'unsafe-inline'`: execute JS scripts inline
  * `img-src https://w3.org blob: data:;`: fetch some image/data from w3
* `explorer/config/runtime.exs`: handle SSL certificates from elixir endpoint
  * `PORT_SSL` env var sets the port for HTTPS endpoint to listen on, not needed as it's defaulted to 443
    * for the 443 binding to work, it's needed to either:
      * run the release with `sudo` (not recommended for security reasons)
      * add capabilities to the BEAM binary to bind to low ports, for example in staging: `sudo setcap CAP_NET_BIND_SERVICE=+eip /home/app/repos/explorer/aligned_layer/explorer/_build/prod/rel/explorer/erts-14.2.1/bin/beam.smp`
  * `KEYFILE_PATH` and `CERTFILE_PATH` are the local paths to the SSL private key and the certificate, if not set, HTTP endpoint will be brought up, but not HTTPS one
  * NOTE: as this config is all inside the [`if config_env() == :prod`](https://github.com/yetanotherco/aligned_layer/pull/1586/files#diff-6952f3e44db7841171ba8354502fd4b534531fccee4d41a3215800f8bc008c46R23), local development setup changes are not needed


JuArce (changes requested):
There is missing the grafana dashboard changes

PatStiles (approved):

JuArce (approved):

samoht9277 (approved):
Works as intended

PR #1598, opened 2024-12-10
refactor(sdk): Remove `NetworkArg` + Make functions that accept `Network` methods of `Network`

PatStiles:
> [!IMPORTANT]
> Closed in favour of #1730 

# Refactor Network Arg

## Description

Refactors the Network struct in the sdk to:
- Derive `ValueEnum` and thus remove the need for `NetworkArg`
- Refactors functions that accept `Network` and return addresses to be methods of `Network`.

## Type of change

- [x] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


MarcosNicolau (approved):
Nice!

batcher/aligned-sdk/Cargo.toml
```diff
@@ -25,3 +25,4 @@ hex = "0.4.3"
 ciborium = "=0.2.2"
 serde_repr = "0.1.19"
 dialoguer = "0.11.0"
+clap = { version = "4.5.4", features = ["derive"] }
```
MarcosNicolau: You should push the new `.lock` file.

PatStiles: Added!

uri-99 (approved):

batcher/aligned-sdk/src/core/types.rs
```diff
@@ -396,14 +398,38 @@ pub enum GetNonceResponseMessage {
...
+#[derive(Debug, Clone, Copy, ValueEnum)]
 pub enum Network {
     Devnet,
     Holesky,
     HoleskyStage,
     Mainnet,
 }
 
+impl Network {
+    pub fn get_batcher_payment_service_address(&self) -> ethers::types::H160 {
```
Oppen, line 416: Nit: pass by copy:
```suggestion
    pub fn get_batcher_payment_service_address(self) -> ethers::types::H160 {
```

batcher/aligned-sdk/src/core/types.rs
```diff
@@ -396,14 +398,38 @@ pub enum GetNonceResponseMessage {
...
+            Self::Devnet => H160::from_str("0x7bc06c482DEAd17c0e297aFbC32f6e63d3846650").unwrap(),
+            Self::Holesky => H160::from_str("0x815aeCA64a974297942D2Bbf034ABEe22a38A003").unwrap(),
+            Self::HoleskyStage => {
+                H160::from_str("0x7577Ec4ccC1E6C529162ec8019A49C13F6DAd98b").unwrap()
+            }
+            Self::Mainnet => H160::from_str("0xb0567184A52cB40956df6333510d6eF35B89C8de").unwrap(),
+        }
+    }
+
+    pub fn get_aligned_service_manager_address(&self) -> ethers::types::H160 {
```
Oppen, line 428: Nit: pass by copy:
```suggestion
    pub fn get_aligned_service_manager_address(self) -> ethers::types::H160 {
```

Oppen (approved):

batcher/aligned-sdk/src/core/types.rs
PatStiles: Great catch! @Oppen

MauroToscano (changes requested):
- It would be better for NetworkArg, requiring clap in the SDK is a bit overkill
- There needs to be a way to use values that are not the default ones, so in case we lunch another network we can use it without changing the code (Either through a config where you load network names and addresses, or some different input, may need to think a bit this)

batcher/aligned-sdk/src/core/types.rs
PatStiles: Great catch! @Oppen

Oppen:
> * It would be better for NetworkArg, requiring clap in the SDK is a bit overkill
> * There needs to be a way to use values that are not the default ones, so in case we lunch another network we can use it without changing the code (Either through a config where you load network names and addresses, or some different input, may need to think a bit this)

Maybe a `Custom` variant that includes the actual parameters? That way, by default you'd just use the name for the known ones, but unexpected upgrades can be done by way of, e.g., passing `--network custom --network-rpc ...`.

PatStiles:
Sounds good guys

PatStiles:
@Oppen @MauroToscano I addressed the comments based on your suggestions.

1.) I implemented removed the need for using `ValueEnum` by using 
2.) I created a `custom BATCHER_PAYMENT_SERVICE ALIGNED_SERVICE_MANAGER` variant of the enum. The most straightforward way to do this within clap while still defining a single `Network` argument was to have the custom variant require `custom <BATCHER_PAYMENT_SERVICE_ADDRESS, ALIGNED_SERVICE_MANAGER_ADDRESS>` to be surrounded by "" or `` to be accepted as a single arg.

batcher/aligned-sdk/src/core/types.rs
Oppen: Now that `Custom(String, String)` it's no longer a good idea to pass by copy. That's also true for `is_proof_verified`.

batcher/aligned-sdk/src/core/constants.rs
```diff
@@ -37,3 +37,21 @@ pub const BUMP_MIN_RETRY_DELAY: u64 = 500; // milliseconds
...
+pub const BATCHER_PAYMENT_SERVICE_ADDRESS_DEVNET: &str =
+    "0x7bc06c482DEAd17c0e297aFbC32f6e63d3846650";
+pub const BATCHER_PAYMENT_SERVICE_ADDRESS_HOLESKY: &str =
+    "0x815aeCA64a974297942D2Bbf034ABEe22a38A003";
+pub const BATCHER_PAYMENT_SERVICE_ADDRESS_HOLESKY_STAGE: &str =
+    "0x7577Ec4ccC1E6C529162ec8019A49C13F6DAd98b";
+pub const BATCHER_PAYMENT_SERVICE_ADDRESS_MAINNET: &str =
+    "0xb0567184A52cB40956df6333510d6eF35B89C8de";
+/// AlignedServiceManager
+pub const ALIGNED_SERVICE_MANAGER_DEVNET: &str = "0x1613beB3B2C4f22Ee086B2b38C1476A3cE7f78E8";
```
avilagaston9: Shouldn't it be `0x851356ae760d987E095750cCeb3bC6014560891C`?

batcher/aligned-sdk/src/core/constants.rs
```diff
@@ -37,3 +37,21 @@ pub const BUMP_MIN_RETRY_DELAY: u64 = 500; // milliseconds
...
+pub const ALIGNED_SERVICE_MANAGER_HOLESKY: &str = "0x58F280BeBE9B34c9939C3C39e0890C81f163B623";
+pub const ALIGNED_SERVICE_MANAGER_HOLESKY_STAGE: &str =
+    "0x9C5231FC88059C086Ea95712d105A2026048c39B";
+pub const ALIGNED_SERVICE_MANAGER_HOLESKY_MAINNET: &str =
```
avilagaston9:
```suggestion
pub const ALIGNED_SERVICE_MANAGER_MAINNET: &str =
```

batcher/aligned-sdk/src/core/types.rs
```diff
@@ -413,10 +445,24 @@ impl FromStr for Network {
...
+            s => {
+                if !s.contains("custom") {
+                    return Err(
+                        "Invalid network, possible values are: \"holesky\", \"holesky-stage\", \"devnet\", \"mainnet\", \"custom <BATCHER_PAYMENT_SERVICE_ADDRESS> <ALIGNED_SERVICE_MANAGER_ADDRESS>\""
+                            .to_string(),
+                    );
+                }
+                let parts: Vec<&str> = s.split_whitespace().collect();
+
+                if parts.len() == 3 {
```
avilagaston9: We should also enforce `custom` to be `parts[0]`.

JulianVentura (approved):

avilagaston9 (changes requested):

JulianVentura (changes requested):
Should update documentation to add new `custom` network parameter on `docs/3_guides/0_submitting_proofs.md`

avilagaston9:
We should also update `9_aligned_cli.md`

PatStiles:
For now the deployment of a custom network will be complete via stating "custom BATCHER_PAYMENT_SERVICE ALIGNED_SERVICE_MANAGER" in quotes. This solution is not ideal and we will discuss and implement a better solution in a follow up pr.

PatStiles:
Due to merge conflicts with the claim logic which this pr predates I recommend we merge and update the docs in a follow up pr.

batcher/aligned-sdk/src/core/constants.rs
PatStiles: addressed

batcher/aligned-sdk/src/core/constants.rs
PatStiles: addressed

batcher/aligned-sdk/src/core/types.rs
PatStiles: addressed

uri-99:
Closing in favor of #1730

PR #1603, opened 2024-12-10
feat: Improve jaeger traces on the aggregator

JulianVentura:
#  Improve jaeger traces on the aggregator

## Description

This PR makes two modifications on the event traces generated while responding to a task on the aggregator:

1. Removed the `Bump gas price` event, in favor of `Gas price set` so we not only get information of the used gas prices when a bump occurs, but also on the first attempt.
2. Added the `effective_gas_price` attribute to the `Task Sent to Ethereum` event, so we know exactly what was the gas price fee after sending the transaction (which may be lower than the one specified, see [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md))
3. Removed a `LogTaskError` on `sendAggregatedResponse` as it was being duplicated

# How to test

## First test, normal behavior

1. Start all services, including telemetry and metrics (grafana dashboard)
2. Send a task, with `make batcher_send_sp1_task`
3. Wait for the task to be verified
4. You should see that the dashboard "# Bumps Sending Responses to Ethereum" is at zero.
5. You should see on jaeger these events:

image

## Second test, retry behavior

1. Simulate a retry in the aggregator by adding the following to the `avs_writer.go`
```go
178	if i < 3 {
	    i++
	    return nil, fmt.Errorf("My error")
        }
```
2. Start all services, including telemetry and metrics (grafana dashboard)
3. Send a task, with `make batcher_send_sp1_task`
4. You should see that the dashboard "# Bumps Sending Responses to Ethereum" is now indicating 3 bumps
5. You should see on jaeger these events:

image

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [X] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/aggregator/pkg/aggregator.go
+++ b/aggregator/pkg/aggregator.go
@@ -285,10 +285,12 @@ func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsA
 	if err == nil {
 		// In some cases, we may fail to retrieve the receipt for the transaction.
 		txHash := "Unknown"
+		effectiveGasPrice := "Unknown"
 		if receipt != nil {
 			txHash = receipt.TxHash.String()
+			effectiveGasPrice = receipt.EffectiveGasPrice.String()
 		}
-		agg.telemetry.TaskSentToEthereum(batchData.BatchMerkleRoot, txHash)
+		agg.telemetry.TaskSentToEthereum(batchData.BatchMerkleRoot, txHash, effectiveGasPrice)
 		agg.logger.Info("Aggregator successfully responded to task",
 			"taskIndex", blsAggServiceResp.TaskIndex,
 			"batchIdentifierHash", "0x"+hex.EncodeToString(batchIdentifierHash[:]))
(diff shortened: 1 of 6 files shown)


aggregator/pkg/aggregator.go
```diff
@@ -285,10 +285,12 @@ func (agg *Aggregator) handleBlsAggServiceResponse(blsAggServiceResp blsagg.BlsA
 	if err == nil {
 		// In some cases, we may fail to retrieve the receipt for the transaction.
 		txHash := "Unknown"
+		effectiveGasPrice := "Unknown"
```
PatStiles, line 288: Very very much a nit but in a separate pr we should move these to a constants file.

PatStiles (approved):

PatStiles:
Ran locally and works well!

uri-99 (approved):

JuArce (approved):

PR #1622, opened 2024-12-16
feat(explorer): selector to switch network

MarcosNicolau:
## Description

Adds a selector in the explorer nav to switch the network.

**Testing**
- Run the explorer locally and you should see the new selector:

**large screens**:
image

**smalls screens**:
image

## Type of change
- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


explorer/lib/explorer_web/live/utils.ex
```diff
@@ -108,6 +108,24 @@ defmodule ExplorerWeb.Helpers do
...
+    String.capitalize(prefix)
+  end
+
+  @doc """
+    Returns a list of available AlignedLayer networks with their names and explorer URLs.
+  """
+  def get_aligned_networks() do
+    [
+      {"Mainnet", "https://explorer.alignedlayer.com"},
+      {"Holesky", "https://holesky.explorer.alignedlayer.com"}
```
uri-99, line 125: missing staging

uri-99 (approved):

uri-99 (changes requested):
we should go to staging

uri-99:
The many file changes should go away when #1625  is merged

uri-99 (approved):

PatStiles (approved):

Oppen (approved):
It works on my machine. It's a little odd that because it works by redirecting to the other domains the UI itself changes. I would have expected as user that only the data it showed would change.

MarcosNicolau:
> It works on my machine. It's a little odd that because it works by redirecting to the other domains the UI itself changes. I would have expected as user that only the data it showed would change.

It also seemed odd to me at first, but if you check sites like [etherscan](https://etherscan.io/) they do the same.

PR #1627, opened 2024-12-16
feat: add node_exporter metrics

avilagaston9:
# Add node_exporter metrics

## Description
Adds [node_exporter](https://github.com/prometheus/node_exporter) to be deployed in the Aligned servers.
- Creates a script to install node_exporter.
- Creates an Ansible playbook to install node_exporter in the servers.
- Uses the [node_exporter_full](https://grafana.com/grafana/dashboards/1860-node-exporter-full) Grafana dashboard.

## How to Test
1. Install node_exporter:
```bash
make install_node_exporter
```
2. Run node_exporter:
```bash
node_exporter
```
3. Go to `localhost:9100` and you should see the exposed metrics.
4. Run metrics:
```bash
make run_metrics
```
5. Go to `localhost:3000` and you should see the new dashboard `Node Exporter Full` with your metrics displayed.
> [!NOTE]
> If you are on macos, some of the metrics will appear with `no data`.

6. Ask for a server and test the playbook.

## Type of change

- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


Oppen (approved):

PatStiles (approved):

PatStiles:
Ran locally and it worked fine. Did not have access to a server to test the playbook yett.

MarcosNicolau (commented):
Tested it locally. I only see the `node_exporter` dashboard for the batcher, is that alright? Shouldn't I also see the aggregator dashboard?

uri-99 (changes requested):
Looks fine but this PR needs approves from DevOps

JulianVentura (commented):
Run it locally and everything seems to work fine, but mac os doesn't show many graphs. I left it running for a few minutes and if I chose the 24h time frame on Grafana some new graphs appears.
Anyways, I think this PR should really be tested on a server so all graphs are shown and we test the ansible script.

avilagaston9:
> Tested it locally. I only see the node_exporter dashboard for the batcher, is that alright? Shouldn't I also see the aggregator dashboard?
@MarcosNicolau 

If you tested it locally, then the metrics displayed were from your computer, not specifically from the batcher. Did it say 'Batcher' anywhere?

infra/ansible/playbooks/node_exporter.yaml
```diff
@@ -0,0 +1,26 @@
+- name: Node Exporter Setup
+  hosts: "{{ host }}"
+
+  tasks:
```
samoht9277, line 4: This Ansible task assumes the repo is already cloned, which might not be the case.

Also, there are some `sender` servers, like `aligned-holesky-stage-1-sender`, which @JuArce told me to test this PR on. Why are we running this ansible inside the Aggergator and Batcher? Is this whole **Node Exporter** thing an addon to this components, or should it be a separate component with it's own server?

infra/ansible/playbooks/node_exporter.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - name: Install node exporter
+      make:
+        chdir: /home/{{ ansible_user }}/repos/{{ host }}/aligned_layer
+        target: install_node_exporter
```
samoht9277, line 8:
```suggestion
        target: install_node_exporter
        version: add-node-exporter # Temporary branch pin to test changes on the server. Remove this line before merging.
```

prometheus/prometheus.yaml
samoht9277, line 1: Unless this file is for deploying `prometheus` locally, we should also add this scrape target to the actual telemetry servers, right?

In my PR adding Ansible to the whole Telemetry, I have a [config file](https://github.com/yetanotherco/aligned_layer/blob/telemetry_ansible/infra/ansible/playbooks/templates/prometheus/prometheus.yaml.j2) I copy to prometheus.

infra/ansible/playbooks/node_exporter.yaml
samoht9277: Add this line to allow the Node Exporter's Ansible to clone your repo. Otherwise, it will not find the proper make targets, and will fail. Remember to remove the line before merging.

scripts/install_node_exporter.sh
samoht9277, line 1: If you found this script online, we should credit the original author.

Please add a comment below the shebang (`#!/bin/bash`) with a link to the original version.

infra/ansible/playbooks/node_exporter.yaml
avilagaston9: Yes, I made the playbook to be used after the repo was cloned intentionally, so we could include it within the playbooks of the Aggregator and Batcher. But I made a mistake when adding the playbook to the others, as the repository hadn't been cloned at that step yet 🤣. I'll fix that in a moment.

prometheus/prometheus.yaml
avilagaston9: Yes, we need to add the two new jobs to that file as well

scripts/install_node_exporter.sh
avilagaston9: I actually use [install_aligned.sh](https://github.com/yetanotherco/aligned_layer/blob/testnet/batcher/aligned/install_aligned.sh) as a reference.

infra/ansible/playbooks/node_exporter.yaml
avilagaston9: Addressed in [#7923c1a](https://github.com/yetanotherco/aligned_layer/pull/1627/commits/7923c1aeb679737200d784d8d06f0bc2f4e94427)!

uri-99 (approved):

JuArce (approved):

JuArce (approved):

PR #1633, opened 2024-12-17
feat: task sender with slack updates and pager duty alerts

JulianVentura:
# Modify how spent amount is calculated on task sender

## Description

We have the task sender configured to send proofs periodically. After a new batch of proofs is verified, a slack message is sent notifying the cost of this submission and verification. The reported cost is associated with the `Fee per proof` variable, which is alright for the users, but not in this case. We just want to know how much did we spend only by creating the task and responding to it, without any fee to the batcher. This PR accomplishes that, by fetching the transaction receipts and calculating the exact fees.

At the same time, some environment variables were added so it's easy to test this feature on different environments

## How to test

1. Modify the `send_slack_message` function on `sender_with_alert.sh` with the following:

```bash
# Function so send Slack message
# @param message
function send_slack_message() {
    echo "$1"
}
```
2. Fund a wallet or use a rich wallet as the task sender (don't use a non-paying one) 
3. Create an `.env` file under `alerts` directory, using `.env.example` as a template
4. Start all services, including the explorer.
5. Execute the task sender by running `./alerts/sender_with_alert.sh ./alerts/.env`
6. You should see the following logs

image

7. You should do some checks on the calculated spent amount:

  - It should be lower than the `fee_per_proof * number_of_proofs`, which you can see on the explorer
  - You should be able to get the same value calculating it by hand

8. You should also test with multiple batches submissions. 
  - For that, you can modify the `max_batch_proof_qty` variable under the `try_build_batch` function on the batcher, so that the amount of proofs sent is split between two batches.
  - You should send a lower number of proofs than the double of your chosen vale for the `max_batch_proof_qty`, so the second batch doesn't have the same merkle root as the first one. For example, `max_batch_proof_qty=7`, `repetitions=12`.

### How to calculate by hand

1. Go to the explorer and open the batch on which you want to make the calculation off
2. Copy the "Submission Transaction Hash"
3. Do the following RPC call and copy the `gasUsed` and `effectiveGasPrice`
```bash
cast receipt --rpc-url <RPC_URL> <TX_HASH>
```
4. Calculate the fee by doing
```
(gasUsed * effectiveGasPrice) / 1e18
```
5. Repeat the process from step (2) for the "Response Transaction Hash"
6. Sum the two values and compare them to the script, they should be almost the same (floating point error)

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [X] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


alerts/sender_with_alert.sh
```diff
@@ -20,6 +37,27 @@ source "$1"
...
 
+# Function to get the tx cost from the explorer:
+# @param merkle_root
+# @param type_of_hash
+#   - Submission Transaction Hash
+#   - Response Transaction Hash
+function fetch_tx_cost() {
+  # Get the tx hash from the explorer
+  explorer_fetch_url="$EXPLORER_URL/batches/$1"
+  #echo "This is the fetch URL: $explorer_fetch_url"
```
uri-99: uncomment or remove

alerts/sender_with_alert.sh
```diff
@@ -68,16 +103,20 @@ do
     2>&1)
 
   echo "$submit"
-
+  
   explorer_link=$(echo "$submit" | grep alignedlayer.com | awk '{print $4}')
```
uri-99: could we use the new env var here?

alerts/sender_with_alert.sh
```diff
@@ -2,16 +2,33 @@
...
+# - EXPLORER_URL
 # - NETWORK
 # - PRIVATE_KEY
+# - VERIFICATION_WAIT_TIME
+# - SLEEP_TIME
 # - PAGER_DUTY_KEY
 # - PAGER_DUTY_EMAIL
 # - PAGER_DUTY_SERVICE_ID
 
+# IMPROVEMENTS:
```
uri-99: There are TODO, right? for this use the TODO keyword

alerts/sender_with_alert.sh
```diff
@@ -20,6 +37,27 @@ source "$1"
...
+# Function to get the tx cost from the explorer:
+# @param merkle_root
+# @param type_of_hash
+#   - Submission Transaction Hash
+#   - Response Transaction Hash
+function fetch_tx_cost() {
+  # Get the tx hash from the explorer
+  explorer_fetch_url="$EXPLORER_URL/batches/$1"
+  #echo "This is the fetch URL: $explorer_fetch_url"
+  tx_hash=$(curl -s $explorer_fetch_url | grep -C 5 "$2" | grep -oE '0x[[:alnum:]]{64}' | head -n 1)
```
uri-99: this sounds too fragile.
We should do a cast call to get the events related to the merkle root. 
more info on how to do this [here](https://book.getfoundry.sh/reference/cast/cast-logs?highlight=cast%20logs#cast-logs)

uri-99 (changes requested):

alerts/sender_with_alert.sh
JulianVentura: Sure, I'll check that

alerts/sender_with_alert.sh
```diff
@@ -2,16 +2,32 @@
 
 # ENV VARIABLES:
```
JuArce, line 3: There is missing SLACK_WEBHOOK_URL

alerts/sender_with_alert.sh
```diff
@@ -68,16 +95,36 @@ do
...
+  # Fetch the logs of both submission and response
+  from_block_number=$((current_block_number - LOGS_BLOCK_RANGE))
+  if [ $from_block_number -lt 0 ]; then
+    from_block_number=0
+  fi
 
-  ## Get final balance
-  final_balance=$(aligned get-user-balance --rpc_url $RPC_URL --user_addr  $SENDER_ADDRESS --network $NETWORK 2>&1 | awk '{print $7}')
-  echo "Final balance: $final_balance"
+  log=$(cast logs --rpc-url $RPC_URL --from-block $from_block_number --to-block latest 'NewBatchV3 (bytes32 indexed batchMerkleRoot, address senderAddress, uint32 taskCreatedBlock, string batchDataPointer, uint256 respondToTaskFeeLimit)' $merkle_root)
```
JuArce:
```suggestion
  log=$(cast logs --rpc-url $RPC_URL --from-block $from_block_number --to-block latest 'NewBatchV3 (bytes32 indexed batchMerkleRoot, address senderAddress, uint32 taskCreatedBlock, string batchDataPointer, uint256 respondToTaskFeeLimit)' $batch_merkle_root)
```

alerts/sender_with_alert.sh
```diff
@@ -68,16 +95,36 @@ do
...
+  submission_tx_hash=$(echo $log | grep -oE "transactionHash: 0x[[:alnum:]]{64}" | awk '{ print $2 }')
 
-  ## Calculate the spent amount
-  spent_amount=$(echo "$initial_balance - $final_balance" | bc | awk '{printf "%.9f", $1}')
+  log=$(cast logs --rpc-url $RPC_URL --from-block $from_block_number --to-block latest 'BatchVerified (bytes32 indexed batchMerkleRoot, address senderAddress)' $merkle_root)
```
JuArce:
```suggestion
  log=$(cast logs --rpc-url $RPC_URL --from-block $from_block_number --to-block latest 'BatchVerified (bytes32 indexed batchMerkleRoot, address senderAddress)' $batch_merkle_root)
```

alerts/sender_with_alert.sh
```diff
@@ -101,15 +148,16 @@ do
   done
 
   ## Send Update to Slack
-  eth_usd=$(curl https://cryptoprices.cc/ETH/)
+  eth_usd=$(curl -s https://cryptoprices.cc/ETH/)
   spent_ammount_usd=$(echo "$spent_amount * $eth_usd" | bc | awk '{printf "%.2f", $1}')
-  slack_meesage="$REPETITIONS Proofs submitted and verified. Spent amount: $spent_amount ETH ($ $spent_ammount_usd) [ $explorer_link ]"
+  slack_meesage="$REPETITIONS Proofs submitted and verified. Spent amount: $spent_amount ETH ($ $spent_ammount_usd) [ $batch_explorer_url ]"
```
JuArce: Typo in message

JuArce (changes requested):

JulianVentura:
@uri-99 @JuArce please review again

uri-99 (approved):

PatStiles (approved):

alerts/sender_with_alert.sh
```diff
@@ -101,15 +158,18 @@ do
   done
 
   ## Send Update to Slack
-  eth_usd=$(curl https://cryptoprices.cc/ETH/)
+  eth_usd=$(curl -s https://cryptoprices.cc/ETH/)
   spent_ammount_usd=$(echo "$spent_amount * $eth_usd" | bc | awk '{printf "%.2f", $1}')
-  slack_meesage="$REPETITIONS Proofs submitted and verified. Spent amount: $spent_amount ETH ($ $spent_ammount_usd) [ $explorer_link ]"
+  slack_meesage="$REPETITIONS Proofs submitted and verified. Spent amount: $spent_amount ETH ($ $spent_ammount_usd) [ ${batch_explorer_urls[@]} ]"
```
JuArce: In 
```
if echo "$verification" | grep -q not; then
      message="Proof verification failed for $proof [ $explorer_link ]"
```

The `$explorer_link` was renamed to `${batch_explorer_urls[@]}`

alerts/sender_with_alert.sh
```diff
@@ -68,16 +94,47 @@ do
...
+
+    log=$(cast logs --rpc-url $RPC_URL --from-block $from_block_number --to-block latest 'NewBatchV3 (bytes32 indexed batchMerkleRoot, address senderAddress, uint32 taskCreatedBlock, string batchDataPointer, uint256 respondToTaskFeeLimit)' $batch_merkle_root)
+    submission_tx_hash=$(echo "$log" | grep -oE "transactionHash: 0x[[:alnum:]]{64}" | awk '{ print $2 }')
+
+    log=$(cast logs --rpc-url $RPC_URL --from-block $from_block_number --to-block latest 'BatchVerified (bytes32 indexed batchMerkleRoot, address senderAddress)' $batch_merkle_root)
+    response_tx_hash=$(echo "$log" | grep -oE "transactionHash: 0x[[:alnum:]]{64}" | awk '{ print $2 }')
+
+    # Calculate fees for transactions
+    submission_fee_in_wei=$(fetch_tx_cost $submission_tx_hash)
+    response_fee_in_wei=$(fetch_tx_cost $response_tx_hash)
```
JuArce, line 131: If the verification is not succeded, this call fails with
```
error: the following required arguments were not provided:
  <TX_HASH>

Usage: cast receipt --rpc-url <URL> <TX_HASH> [FIELD]

For more information, try '--help'.
./alerts/sender_with_alert.sh: line 47: * : syntax error: operand expected (error token is "* ")
```

To test it, you can change VERIFICATION_WAIT_TIME=1

alerts/sender_with_alert.sh
```diff
@@ -101,15 +158,18 @@ do
   done
 
   ## Send Update to Slack
-  eth_usd=$(curl https://cryptoprices.cc/ETH/)
+  eth_usd=$(curl -s https://cryptoprices.cc/ETH/)
   spent_ammount_usd=$(echo "$spent_amount * $eth_usd" | bc | awk '{printf "%.2f", $1}')
```
JuArce:
```suggestion
  spent_amount_usd=$(echo "$spent_amount * $eth_usd" | bc | awk '{printf "%.2f", $1}')
```

alerts/sender_with_alert.sh
```diff
@@ -101,15 +158,18 @@ do
...
-  slack_meesage="$REPETITIONS Proofs submitted and verified. Spent amount: $spent_amount ETH ($ $spent_ammount_usd) [ $explorer_link ]"
+  slack_meesage="$REPETITIONS Proofs submitted and verified. Spent amount: $spent_amount ETH ($ $spent_ammount_usd) [ ${batch_explorer_urls[@]} ]"
+  
+  echo "$slack_meesage"
```
JuArce:
```suggestion
  echo "$slack_message"
```

alerts/sender_with_alert.sh
```diff
@@ -101,15 +158,18 @@ do
...
+  
   send_slack_message "$slack_meesage"
```
JuArce:
```suggestion
  send_slack_message "$slack_message"
```

alerts/sender_with_alert.sh
JuArce: If the proof verification failed, do not log "... and verified..."

JuArce:
```suggestion
  slack_message="$REPETITIONS Proofs submitted and verified. Spent amount: $spent_amount ETH ($ $spent_ammount_usd) [ ${batch_explorer_urls[@]} ]"
```

JuArce (changes requested):

alerts/sender_with_alert.sh
```diff
@@ -68,16 +94,47 @@ do
...
+  # Get all the batches merkle roots
+  batch_merkle_roots=$(echo "$submit" | grep "Batch merkle root: " | grep -oE "0x[[:alnum:]]{64}" | uniq)
 
-  ## Get final balance
-  final_balance=$(aligned get-user-balance --rpc_url $RPC_URL --user_addr  $SENDER_ADDRESS --network $NETWORK 2>&1 | awk '{print $7}')
-  echo "Final balance: $final_balance"
+  batch_explorer_urls=()
+
+  # Fetch the logs of both submission and response
+  from_block_number=$((current_block_number - LOGS_BLOCK_RANGE))
```
JuArce: current_block_number is never calculated

JuArce (changes requested):

JuArce (approved):

PR #1637, opened 2024-12-18
hotfix: batcher queue ord

uri-99:
> [!important]
> This PR was rebased here: #1664
# Fix batcher queue ord

## Description

There was a bug in the batcher queue ordering of elements, which led to a wrong placement of proofs when a same sender sent proofs with same max_fee.
The consecuence of this bug was seen when the batcher queue is filled, the proofs sent to chain will be not from lowest to highest nonce, but the other way around.

## To Test

You can view the unit tests.

Also you can add the following print statements (and can do before this PR so you can verify the bug indeed existed).

in `batcher/aligned-batcher/src/lib.rs`, line 1175:
```
        info!("resulting:");
        for (entry, _priority) in resulting_batch_queue.iter() {
            info!(
                "nonce: {:?}, max fee: {:?}",
                entry.nonced_verification_data.nonce, entry.nonced_verification_data.max_fee
            );
        }

        info!("finalized:");
        for entry in finalized_batch.iter() {
            info!(
                "nonce: {:?}, max fee: {:?}",
                entry.nonced_verification_data.nonce, entry.nonced_verification_data.max_fee
            );
        }
```

in `messaging.rs`, line 161:
```
        info!("Last proof nonce: {:?}", last_proof_nonce);
        info!("Current proof nonce: {:?}", batch_inclusion_data_message.user_nonce);
```

This will help to view the resulting state of the batcher queue. 

To execute the bug you should send a burst bigger than the batch_qty limit. For this it is recommended to lower this value, `config-batcher.yaml`:
```
  max_batch_proof_qty: 5 # 5 proofs in a batch, for testing
```

Then send a burst of size 8, so that the first batch is of size 5 and the second of size 3. (this ensures you don't get the `batch already submitted` contract revert).

For this you can set `BURST_SIZE ?= 8` in the Makefile.

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Refactor

## Checklist

- [ ] Linked to Github Issue
- [x] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/batcher/aligned-batcher/src/connection.rs
+++ b/batcher/aligned-batcher/src/connection.rs
@@ -22,7 +22,7 @@ pub(crate) async fn send_batch_inclusion_data_responses(
     batch_merkle_tree: &MerkleTree<VerificationCommitmentBatch>,
 ) -> Result<(), BatcherError> {
     // iter in reverse because each sender wants to receive responses in ascending nonce order
-    // and finalized_batch is ordered as the PriorityQueue , low max_nonce first && high nonce first.
+    // and finalized_batch is ordered as the PriorityQueue , low max_fee first && high nonce first.
     for (vd_batch_idx, entry) in finalized_batch.iter().enumerate().rev() {
         let batch_inclusion_data = BatchInclusionData::new(
             vd_batch_idx,
--- a/batcher/aligned-batcher/src/connection.rs
+++ b/batcher/aligned-batcher/src/connection.rs
@@ -21,8 +21,8 @@ pub(crate) async fn send_batch_inclusion_data_responses(
     finalized_batch: Vec<BatchQueueEntry>,
     batch_merkle_tree: &MerkleTree<VerificationCommitmentBatch>,
 ) -> Result<(), BatcherError> {
-    // iter in reverse because each sender wants to receive responses in ascending nonce order
-    // and finalized_batch is ordered as the PriorityQueue , low max_fee first && high nonce first.
+    // Finalized_batch is ordered as the PriorityQueue, ordered by: ascending max_fee && if max_fee is equal, by descending nonce.
+    // We iter it in reverse because each sender wants to receive responses in ascending nonce order
     for (vd_batch_idx, entry) in finalized_batch.iter().enumerate().rev() {
         let batch_inclusion_data = BatchInclusionData::new(
             vd_batch_idx,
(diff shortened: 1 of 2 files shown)


PatStiles (approved):

batcher/aligned-batcher/src/connection.rs
```diff
@@ -21,7 +21,9 @@ pub(crate) async fn send_batch_inclusion_data_responses(
     finalized_batch: Vec<BatchQueueEntry>,
     batch_merkle_tree: &MerkleTree<VerificationCommitmentBatch>,
 ) -> Result<(), BatcherError> {
-    for (vd_batch_idx, entry) in finalized_batch.iter().enumerate() {
+    // iter in reverse because each sender wants to receive responses in ascending nonce order
+    // and finalized_batch is ordered as the PriorityQueue , low max_nonce first && high nonce first.
```
MarcosNicolau:
```suggestion
    // and finalized_batch is ordered as the PriorityQueue , low max_nonce first && high nonce last.
```

MarcosNicolau (approved):
Pretty edgy situation... nice catch. Working on my machine.

batcher/aligned-batcher/src/connection.rs
JulianVentura: nit: 
When we revert this vec, we are also reverting the orden of the proofs to low `max_fee` first, and higher `max_fee` last, too.
This shouldn't be a problem, since the sender is waiting until arrival of the last nonce, it doesn't care about the fees.
I think we should add a comment clarifying that.

JulianVentura (approved):
Code looks good and works on my machine

batcher/aligned-batcher/src/connection.rs
uri-99:
```suggestion
    // and finalized_batch is ordered as the PriorityQueue , low max_fee first && high nonce first.
```

uri-99: PriorityQueue is ordered by which elements do we discard first.
So we first discard proofs of low max_fee and high nonce. Because we want the high max_fee and LOW nonce to go in first

uri-99: you are right, this is not good. because the low max_fee tends to be the high nonce, and we want the high nonce to be last.

uri-99: Priority queue orders first by max_fee.
This means the priorityQueue has ascending max_fee (we first eliminate those with low max_fee). And when max_fee is equal, it has descending nonce (we first eliminate those with high nonce)

This means priorityQueue.rev() has descending max_fee. And when the max_fee is equal, it will have ascending nonce. 

This is the desired behaviour, since descending max_fee => ascending nonce, and equal fee => ascending nonce.

PR #1638, opened 2024-12-18
feat(explorer): highlight current page

avilagaston9:
# Highlight Current Page

## Description

Highlights the current page in the nav bar in the explorer.

image
image

## Type of change

- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/explorer/lib/explorer_web/components/nav.ex
+++ b/explorer/lib/explorer_web/components/nav.ex
@@ -126,19 +126,12 @@ defmodule NavComponent do
   defp active_view_class(ExplorerWeb.Home.Index, ExplorerWeb.Home.Index), do: "text-4xl"
   defp active_view_class(_other, ExplorerWeb.Home.Index), do: "text-3xl"
 
-  defp active_view_class(current_view, ExplorerWeb.Batches.Index = target_view) do
-    if current_view == target_view || current_view == ExplorerWeb.Batch.Index do
-      "text-green-500 font-bold"
-    else
-      "text-foreground/80 hover:text-foreground font-semibold"
-    end
-  end
+  defp active_view_class(ExplorerWeb.Batches.Index, ExplorerWeb.Batches.Index ), do: "text-green-500 font-bold"
+  defp active_view_class(ExplorerWeb.Batch.Index, ExplorerWeb.Batches.Index ), do: "text-green-500 font-bold"
+  defp active_view_class(_other, ExplorerWeb.Batches.Index ), do: "text-foreground/80 hover:text-foreground font-semibold"
 
-  defp active_view_class(current_view, ExplorerWeb.Operators.Index = target_view) do
-    if current_view == target_view || current_view == ExplorerWeb.Operator.Index do
-      "text-green-500 font-bold"
-    else
-      "text-foreground/80 hover:text-foreground font-semibold"
-    end
-  end
+  defp active_view_class(ExplorerWeb.Operators.Index, ExplorerWeb.Operators.Index ), do: "text-green-500 font-bold"
+  defp active_view_class(ExplorerWeb.Operator.Index, ExplorerWeb.Operators.Index ), do: "text-green-500 font-bold"
+  defp active_view_class(_other, ExplorerWeb.Operators.Index ), do: "text-foreground/80 hover:text-foreground font-semibold"
+
 end
--- a/explorer/lib/explorer_web/components/nav.ex
+++ b/explorer/lib/explorer_web/components/nav.ex
@@ -18,20 +18,28 @@ defmodule NavComponent do
     }>
       <div class="gap-x-6 inline-flex">
         <.link
-          class="text-3xl hover:scale-105 transform duration-150 active:scale-95"
+          class={
+            classes([
+              "hover:scale-105",
+              "transform",
+              "duration-150",
+              "active:scale-95",
+              active_class(assigns.socket.view, ExplorerWeb.Home.Index)
+            ])
+          }
           navigate={~p"/"}
         >
           🟩 <span class="sr-only">Aligned Explorer Home</span>
         </.link>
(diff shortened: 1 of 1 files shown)


explorer/lib/explorer_web/components/nav.ex
```diff
@@ -18,20 +18,28 @@ defmodule NavComponent do
...
       <div class="gap-x-6 inline-flex">
         <.link
-          class="text-3xl hover:scale-105 transform duration-150 active:scale-95"
+          class={
+            classes([
+              "hover:scale-105",
+              "transform",
+              "duration-150",
+              "active:scale-95",
+              active_view_class(assigns.socket.view, ExplorerWeb.Home.Index)
```
MarcosNicolau: I would not change the logo size, it looks weird.

explorer/lib/explorer_web/components/nav.ex
```diff
@@ -109,4 +122,16 @@ defmodule NavComponent do
...
+  defp active_view_class(ExplorerWeb.Home.Index, ExplorerWeb.Home.Index), do: "text-4xl"
+  defp active_view_class(_other, ExplorerWeb.Home.Index), do: "text-3xl"
+
+  defp active_view_class(ExplorerWeb.Batches.Index, ExplorerWeb.Batches.Index ), do: "text-green-500 font-bold"
+  defp active_view_class(ExplorerWeb.Batch.Index, ExplorerWeb.Batches.Index ), do: "text-green-500 font-bold"
+  defp active_view_class(_other, ExplorerWeb.Batches.Index ), do: "text-foreground/80 hover:text-foreground font-semibold"
+
+  defp active_view_class(ExplorerWeb.Operators.Index, ExplorerWeb.Operators.Index ), do: "text-green-500 font-bold"
+  defp active_view_class(ExplorerWeb.Operator.Index, ExplorerWeb.Operators.Index ), do: "text-green-500 font-bold"
+  defp active_view_class(_other, ExplorerWeb.Operators.Index ), do: "text-foreground/80 hover:text-foreground font-semibold"
```
MarcosNicolau: I would avoid pattern matching here. You end up repeating the styles all over and it make testing and updating each case very difficult.

explorer/lib/explorer_web/components/nav.ex
uri-99: +1 , i would experiment other effects. i.e. making it brighter or underline it

explorer/lib/explorer_web/components/nav.ex
uri-99: What alternative do you suggest?
I think the only problem with the current solution is that uncatched cases are not identifiable (i.e. the restakings page)

uri-99 (commented):
left some comments

explorer/lib/explorer_web/components/nav.ex
Oppen: If brighter, I think we'd need better contrast, like a much darker default and much brighter highlight. I think the underline would be easier.

explorer/lib/explorer_web/components/nav.ex
Oppen: For reducing styling duplicates, maybe just use a separate CSS class and return only that class might work, right? I mean, with the exception of the logo everything follows the pattern "if active make green and bold, otherwise default text-foreground and semibold".

Then, we can trim the cases a bit by considering that the only same-class pair that does not follow this pattern is the logo, so everything else could be covered by a single pattern:
```elixir
defp active_view_class(class_x, class_x), do: "active-view"
```
Similarly, our fallthrough to non-highlighter would be:
```elixir
defp active_view_class(class_x, class_y), do: "inactive-view"
```

Oppen: At the end I think we would have something like this, if applying this idea:
```elixir
defp active_view_class(ExplorerWeb.Home.Index, ExplorerWeb.Home.Index), do: "active-view aligned-icon"
defp active_view_class(_other, ExplorerWeb.Home.Index), do: "inactive-view aligned-icon"
defp active_view_class(class_x, class_x), do: "active-view"
defp active_view_class(ExplorerWeb.Operator.Index, ExplorerWeb.Operators.Index), do: "active-view"
defp active_view_class(ExplorerWeb.Batch.Index, ExplorerWeb.Batches.Index), do: "active-view"
defp active_view_class(class_x, class_y), do: "inactive-view"
```

avilagaston9: How much do we actually gain with this? I don't find the last solution significantly clearer than the current one. Also, we are only reducing two lines, but now we need to add four new class definitions. That said, I'm open to changing it if you think it's worthwhile.

explorer/lib/explorer_web/components/nav.ex
avilagaston9: I will try using the underline. I have tried other effects, but they didn't seem to have any impact, possibly because it is an icon, not text.

avilagaston9: Removed the effects on the icon since they have no impact: [#39a6ec2](https://github.com/yetanotherco/aligned_layer/pull/1638/commits/39a6ec2831c77ec2365f2dc9a7b6bfc5b599402b)

explorer/lib/explorer_web/components/nav.ex
avilagaston9: Since I removed the effects on the icon, I was able to simplify the code a bit: [#39a6ec2](https://github.com/yetanotherco/aligned_layer/pull/1638/commits/39a6ec2831c77ec2365f2dc9a7b6bfc5b599402b)

uri-99 (approved):

PR #1645, opened 2024-12-18
chore(claim): move state update before call

xqft:
# move state update before call

## Description

If the `nonReentrant` modifier is removed, the  `claim` function is prone to the [reentrancy bug](https://github.com/crytic/not-so-smart-contracts/tree/master/reentrancy), by moving the state update above the transfer call we add redundancy to protect against the bug.

## Type of change

- [ ] Bug fix


ilitteri (approved):

jrchatruc (approved):

ManuelBilbao (approved):

ilitteri:
Fixes Least Authority Suggestion 1

PR #1621, opened 2024-12-16
feat: operator liveness metric

avilagaston9:
# Operator Liveness Metric

## Motivation

We need a way to rapidly determine if an operator is down.

## Description

Adds a Bar Gauge with the count of missed tasks for each operator over a specified time range.

image

## How To Test

1. Start anvil:
```bash
make anvil_start_with_block_time
```
2. Start the telemetry server in interactive mode:
```bash
make telemetry_run_db
make telemetry_ecto_migrate
cd telemetry_api
ALIGNED_CONFIG_FILE="../contracts/script/output/devnet/alignedlayer_deployment_output.json" OPERATOR_FETCHER_WAIT_TIME_MS=5000 ENVIRONMENT=devnet RPC_URL=http://localhost:8545 iex -S mix phx.server
```
3. Run metrics:
```bash
make run_metrics
```
4. Initialize different operator names from the telemetry terminal:
```Elixir
TelemetryApi.PrometheusMetrics.initialize_operator_metrics("gaston")
```
4. Call the `missing_operator` method from the telemetry terminal with the initialized operator names:
```Elixir
TelemetryApi.PrometheusMetrics.missing_operator("gaston")
```
6. Go to the aggregator-batcher dashboard in `localhost:3000` and you should see the dashboard with the values.

### Test also the full flow:
1. Start anvil:
```bash
make anvil_start_with_block_time
```
2. Start batcher:
```bash
make batcher_start_local
```
3. Go to `config-files/config-aggregator.yaml` and reduce the `bls_service_task_timeout`.
4. Start the aggregator:
```bash
make aggregator_start
```
7. Register and start the operator:
```bash
make operator_register_and_start
```
8. Send tasks:
```bash
make batcher_send_burst_groth16
```

9. Kill the operator and watch the panel.

## Type of change

- [x] New feature

## Checklist

- [ ] “Hotfix” to `testnet`, everything else to `staging`
- [ ] Linked to Github Issue
- [ ] This change depends on code or research by an external entity
  - [ ] Acknowledgements were updated to give credit
- [ ] Unit tests added
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.
- [ ] This change is an Optimization
  - [ ] Benchmarks added/run
- [ ] Has a known issue
  - [Link to the open issue addressing it]() 
- [ ] If your PR changes the Operator compatibility (Ex: Upgrade prover versions)
  - [ ] This PR adds compatibility for operator for both versions and do not change batcher/docs/examples
[...]

PR #1648, opened 2024-12-18
feat: add operator responses dashboard

avilagaston9:
# Add Operator Responses Dashboard

## Description

This PR adds a new Grafana dashboard with the count of the operators responses received in the selected time range:

image

## How to test

You can follow similar steps as #1621 to test the full flow of this PR.

## Type of change

- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


PatStiles (approved):
Tested locally and everything works!

MarcosNicolau (approved):
Works nicely! Question: doesn't this metric make the one from #1621 redundant? Since this metric shows the number of responded tasks and the total tasks, the difference in responded tasks can be derived (this could be done via Grafana). Also, this metric updates in real time, whereas the one in #1621 updates only when quorum is reached.

JuArce (changes requested):
- Sort the values by lower responding operators first
- Show a single color without gradient

avilagaston9:
> - Sort the values by lower responding operators first
> - Show a single color without gradient
> @JuArce 

Addressed both using a table + override with a gauge display. I had to use the "RETRO LCD" display because with the basic one, the progress bar appears empty when all values are the same:

image

image

JuArce (approved):

PR #1663, opened 2024-12-19
feat: add batcher accumulated cost metric

avilagaston9:
# Add Batcher Cost Metric

## Description

This PR adds four panels showing the accumulated spending costs by the Batcher for creating and canceling tasks.

- One for the total accumulated  cost (CreateTask + cancel transaction) over a specified a time range.
- One for the accumulated cost of cancel transactions only, over a specified a time range.

image

- Two more, with the historical versions of the previous ones:

image

## How to test

To easily send a cancel transaction, you can go to `batcher/aligned-batcher/src/retry/batcher_retryables.rs:165` and edit the code to return an error:
```Rust
// timeout to prevent a deadlock while waiting for the transaction to be included in a block.
    let _result = timeout(Duration::from_millis(transaction_wait_timeout), pending_tx)
        .await
        .map_err(|e| {
            warn!("Error while waiting for batch inclusion: {e}");
            RetryError::Permanent(BatcherError::ReceiptNotFoundError)
        })?
        .map_err(|e| {
            warn!("Error while waiting for batch inclusion: {e}");
            RetryError::Permanent(BatcherError::ReceiptNotFoundError)
        })?
        .ok_or(RetryError::Permanent(BatcherError::ReceiptNotFoundError));
    Err(RetryError::Permanent(BatcherError::ReceiptNotFoundError))
```

## Type of change

- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


PatStiles (approved):

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1646,14 +1651,30 @@ impl Batcher {
...
         };
         self.metrics
             .cancel_create_new_task_duration
             .set(start.elapsed().as_millis() as i64);
     }
 
+    fn calculate_gas_cost(gas_price: Option<U256>, gas_used: Option<U256>) -> U256 {
+        if let (Some(gas_price), Some(gas_used)) = (gas_price, gas_used) {
+            let wei_gas_cost = gas_price * gas_used;
+            let gwei_gas_cost = wei_gas_cost / U256::from(10e9 as i64);
```
JulianVentura: I think this conversion is not right, should be `1e9` instead

batcher/aligned-batcher/src/lib.rs
```diff
@@ -1646,14 +1651,30 @@ impl Batcher {
...
+                    .inc_by(gas_cost.as_u64());
+            }
             Err(e) => error!("Could not cancel createNewTask transaction: {e}"),
         };
         self.metrics
             .cancel_create_new_task_duration
             .set(start.elapsed().as_millis() as i64);
     }
 
+    fn calculate_gas_cost(gas_price: Option<U256>, gas_used: Option<U256>) -> U256 {
```
JulianVentura: I would name it `calculate_gas_cost_in_gwei` or `gas_cost_in_gwei`

JulianVentura (commented):
I think we are not converting to gwei properly, I left you a comment pointing were I think the problem is.
Other than that, everything seems to be working fine.

batcher/aligned-batcher/src/lib.rs
MarcosNicolau: indeed, this is doing $10 * 10^9 = 10^{10}$

avilagaston9: lol, that's right! Addressed in [#6c40b9a](https://github.com/yetanotherco/aligned_layer/pull/1663/commits/6c40b9af638dbc80f05e7ec73f19d4d325b256dd)

batcher/aligned-batcher/src/lib.rs
avilagaston9: Addressed in [#6c40b9a](https://github.com/yetanotherco/aligned_layer/pull/1663/commits/6c40b9af638dbc80f05e7ec73f19d4d325b256dd)!

JulianVentura (approved):
Works fine, but the cost measurement has some error close to `0.7%`, probably due to the usage of floating point.
Anyways, I think this is low priority so we can merge it.

MarcosNicolau (approved):

uri-99 (approved):

JuArce (approved):

PR #1666, opened 2024-12-20
fix: periodic updates for batcher page

### LINKED ISSUE #1646: bug: irregular updates in explorer
While working on #1624 it was noticed that the LiveViews sometime stop updating,
despite the broadcast messages being added and other measures.
It's probably an RTFM issue but still needs to be better understood and fixed.
It's not local to the changes in #1624, as the same behavior is expected for
other unchanged properties and seems to not be applied.

Oppen:
## Motivation

Details like age and badge weren't updating. Fixes #1646.

* Extract batch age to a field in the dict so LiveView knows when it changed
* Broadcast an event on the block update task so views know to update assigns

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Refactor

## Testing

1. Deploy a devnet and the explorer.
2. Send a few proofs.
3. Stop sending proofs.
4. Go to the ex batches page. You should see the age column updating about once a minute.
5. Kill the operator.
6. Send a few more proofs.
7. Stop sending proofs.
8. After a while (5 minutes by default) the pending proofs should start turning red, with the badge signaling "unverified".
9. Verify also for a given batch and for the frontpage.

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


MarcosNicolau (approved):
Working nicely!

uri-99 (approved):

PR #1692, opened 2024-12-30
fix(operator): Operator panic's when reconnect fails after max retries

### LINKED ISSUE #960: fix: operator panic when reconnect fails after max retries
Operator panics when unable to resubscribe to rpc after max retries are reached.

stack trace:

```
goroutine 61 [running]:
github.com/yetanotherco/aligned_layer/core/chainio.(*AvsSubscriber).SubscribeToNewTasks.func2()
        /root/aligned_layer/core/chainio/avs_subscriber.go:115 +0x5e
created by github.com/yetanotherco/aligned_layer/core/chainio.(*AvsSubscriber).SubscribeToNewTasks in goroutine 1
        /root/aligned_layer/core/chainio/avs_subscriber.go:112 +0x36a
```

PatStiles:
# Fix: Operator Panic when reconnect fails after max Retries

## Description

Closes #960. 
`SubscribeToNewTasksV2Retryable` and `SubscribeToNewTasksV3Retryable` are called when the [operator starts](https://github.com/yetanotherco/aligned_layer/blob/staging/operator/pkg/operator.go#L208) . `SubscribeToNewTasksV2Retryable` and `SubscribeToNewTasksV3Retryable` both spawn a go-routine to handle errors from the subscription to two rpc's a primary and a fallback. Upon erroring the go-routine attempts to subscribe again up to a retry limit.

When `SubscribeToNewTasksV2Retryable` and `SubscribeToNewTasksV3Retryable` exhaust the maximum number of retries the respective subscribe functions panics due to dereferencing a `nil` reference within the auto-generated contract bindings. 

In the case that one of the subscriptions errors and then fails to re-connect the operator will panic while one subscription (the fallback subscription) is still available and the operator can continue to function as normal. 

In the case that both connections fail error and fail to reconnect then the operator should shut down.

If the operator fails to subscribe initially the operator should shut down.

This pr addressed this issue by:
1.) Adding a `defer()` to the go routine the converts the panic into an error before the operator process exits.
2.) Changing the retry parameters for `SubscribeToNewTasksV2Retryable` and `SubscribeToNewTasksV3Retryable` to retry infinitely (up to retry intervals of 60 sec).

### To Test:
- install and use nginx  as a proxy to test the connection `brew install nginx`
- create a directory nginx with the file nginx.conf inside of it, with this content:
```
events { }

http {
    server {
        listen 8082;

        location /main/ {
            proxy_pass http://host.docker.internal:8545/;  # Note the trailing slashes
            proxy_set_header Host $host;
            proxy_set_header X-Real-IP $remote_addr;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header X-Forwarded-Proto $scheme;
            proxy_http_version 1.1;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection "upgrade";
            proxy_connect_timeout 1d;
            proxy_send_timeout 1d;
            proxy_read_timeout 1d;
        }
    }
}
```
- create a docker-compose.yml with the following contents:
```
version: '3'
services:
  nginx:
    image: nginx:alpine
    container_name: nginx-anvil-proxy
    volumes:
      - ./nginx:/etc/nginx
    ports:
      - "8082:8082"
 ```
- Make sure the nginx.conf file is located within ./nignx folder and you docker-compose-file has the proper path to the ngix.conf file.
 - change the respective Operator RPC url's in `config-files` within the aligned-layer repo to use `:8082/main/` port and path.
 - Start a local anvil node `make anvil_start_with_block_time`
 - Start nginx in a docker container via `docker compose -f nginx-docker-compose.yaml up -d`
 - Observe proof verification is successful.
 - kill the connection to anvil via `docker compose -f nginx-docker-compose.yaml stop`
 - observe operator blocks with the following message:
 
Screenshot 2024-12-30 at 17 08 42

- Start the nginx proxy again `docker compose -f nginx-docker-compose.yaml up -d` && restart the aggregator `make aggregator_start`
- Observe the operator reconnects and start verifying batches
Screenshot 2024-12-30 at 17 10 41
Panic Operation:
- change line 49 in `core/retry.go` to 	
```
SubscribeToNewTasksNumRetries = 1
```
- Start the local network with the nginx proxy as described above and kill the proxy
- Observe the operator errors and shuts down with the panic wrapped within an error.

Screenshot 2024-12-30 at 16 42 17

## Type of change

- [x] Bug fix

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/core/chainio/avs_subscriber.go
+++ b/core/chainio/avs_subscriber.go
@@ -128,7 +128,7 @@ func (s *AvsSubscriber) SubscribeToNewTasksV2(newTaskCreatedChan chan *servicema
 			case err := <-subFallback.Err():
 				s.logger.Warn("Error in fallback new task subscription", "err", err)
 				subFallback.Unsubscribe()
-				subFallback, err = SubscribeToNewTasksV2Retryable(&bind.WatchOpts{}, s.AvsContractBindings.ServiceManagerFallback, internalChannel, nil, retry.NetworkRetryParams())
+				subFallback, err = SubscribeToNewTasksV2Retryable(&bind.WatchOpts{}, s.AvsContractBindings.ServiceManagerFallback, internalChannel, nil, retry.SubscribeToNewTasksParams())
 				if err != nil {
 					errorChannel <- err
 				}
--- a/core/chainio/avs_subscriber.go
+++ b/core/chainio/avs_subscriber.go
@@ -180,6 +180,13 @@ func (s *AvsSubscriber) SubscribeToNewTasksV3(newTaskCreatedChan chan *servicema
 
 	// Handle errors and resubscribe
 	go func() {
+		defer func() {
+			err := recover() //stops panics
+			if err != nil {
+				s.logger.Error("SubscribeToNewTasksV3 error channel recovered from panic", "err", err)
+			}
+		}()
+
 		for {
 			select {
 			case err := <-sub.Err():
--- a/core/chainio/avs_subscriber.go
+++ b/core/chainio/avs_subscriber.go
@@ -192,14 +192,14 @@ func (s *AvsSubscriber) SubscribeToNewTasksV3(newTaskCreatedChan chan *servicema
 			case err := <-sub.Err():
 				s.logger.Warn("Error in new task subscription", "err", err)
 				sub.Unsubscribe()
-				sub, err = SubscribeToNewTasksV3Retryable(&bind.WatchOpts{}, s.AvsContractBindings.ServiceManager, internalChannel, nil, retry.NetworkRetryParams())
+				sub, err = SubscribeToNewTasksV3Retryable(&bind.WatchOpts{}, s.AvsContractBindings.ServiceManager, internalChannel, nil, retry.SubscribeToNewTasksV3Params())
 				if err != nil {
 					errorChannel <- err
 				}
 			case err := <-subFallback.Err():
 				s.logger.Warn("Error in fallback new task subscription", "err", err)
 				subFallback.Unsubscribe()
-				subFallback, err = SubscribeToNewTasksV3Retryable(&bind.WatchOpts{}, s.AvsContractBindings.ServiceManagerFallback, internalChannel, nil, retry.NetworkRetryParams())
+				subFallback, err = SubscribeToNewTasksV3Retryable(&bind.WatchOpts{}, s.AvsContractBindings.ServiceManagerFallback, internalChannel, nil, retry.SubscribeToNewTasksV3Params())
 				if err != nil {
 					errorChannel <- err
 				}
--- a/core/chainio/avs_subscriber.go
+++ b/core/chainio/avs_subscriber.go
@@ -109,12 +109,19 @@ func (s *AvsSubscriber) SubscribeToNewTasksV2(newTaskCreatedChan chan *servicema
 
 	// Handle errors and resubscribe
 	go func() {
+		defer func() {
+			err := recover() //stops panics
+			if err != nil {
+				s.logger.Error("SubscribeToNewTasksV2 error channel recovered from panic", "err", err)
+			}
+		}()
+
 		for {
 			select {
 			case err := <-sub.Err():
 				s.logger.Warn("Error in new task subscription", "err", err)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: core/chainio/avs_subscriber.go [hunk 1]
```diff
@@ -109,12 +109,19 @@ func (s *AvsSubscriber) SubscribeToNewTasksV2(newTaskCreatedChan chan *servicema
...
+			}
+		}()
+
 		for {
 			select {
 			case err := <-sub.Err():
 				s.logger.Warn("Error in new task subscription", "err", err)
 				sub.Unsubscribe()
-				sub, err = SubscribeToNewTasksV2Retryable(&bind.WatchOpts{}, s.AvsContractBindings.ServiceManager, internalChannel, nil, retry.NetworkRetryParams())
+				sub, err = SubscribeToNewTasksV2Retryable(&bind.WatchOpts{}, s.AvsContractBindings.ServiceManager, internalChannel, nil, retry.SubscribeToNewTasksParams())
```
avilagaston9, line 124: Aren't we missing the same change with the fallback?

avilagaston9, line 124: We are missing the same change with the fallback right?

core/chainio/retryable.go
```diff
@@ -215,7 +215,7 @@ func SubscribeToNewTasksV2Retryable(
 SubscribeToNewTasksV3Retryable
 Subscribe to NewBatchV3 logs from the AVS contract.
 - All errors are considered Transient Errors
-- Retry times (3 retries): 1 sec, 2 sec, 4 sec.
+- Retry times (Infinite Retries): 1 sec, 2 sec, 4 sec, 8 sec, ..., 60 sec, ...,
```
avilagaston9, line 218: We should update `SubscribeToNewTasksV2Retryable` comments as well

core/retry.go
```diff
@@ -32,18 +32,21 @@ const (
...
 
 	// Retry Parameters for RespondToTaskV2 in the Fee Bump
-	RespondToTaskV2MaxInterval           = time.Millisecond * 500 // Maximum interval for an individual retry.
-	RespondToTaskV2MaxElapsedTime        = 0                      //	Maximum time all retries may take. `0` corresponds to no limit on the time of the retries.
-	RespondToTaskV2NumRetries     uint64 = 0                      // Total number of retries attempted. If 0, retries indefinitely until maxElapsedTime is reached.
+	RespondToTaskV2MaxInterval    = time.Millisecond * 500 // Maximum interval for an individual retry.
+	RespondToTaskV2MaxElapsedTime = 0                      //	Maximum time all retries may take. `0` corresponds to no limit on the time of the retries.
+	RespondToTaskV2NumRetries     = 0                      // Total number of retries attempted. If 0, retries indefinitely until maxElapsedTime is reached.
+
+	// Retry Parameters for SubscribeToNewTasksV3
```
avilagaston9:
```suggestion
	// Retry Parameters for SubscribeToNewTasks
```

PatStiles: addressed!

core/chainio/retryable.go
PatStiles: addressed!

core/chainio/avs_subscriber.go
PatStiles: Great Catch! Addressed!

avilagaston9 (approved):
Works in my machine! Maybe we should add logs for each retry to warn that we are trying to reconnect. Otherwise, it might seem like nothing is happening.

PatStiles:
> Works in my machine! Maybe we should add logs for each retry to warn that we are trying to reconnect. Otherwise, it might seem like nothing is happening.

Agreed! @JuArce should we do this in a separate pr.

MauroToscano (changes requested):
I'm not convinced not shutting down is the best solution. I'm putting this on hold until the description tells in which scenario the subscription is retrying for a long time but the node is working

PatStiles:
Closing in favor of #1729

PR #1694, opened 2024-12-30
fix(telemetry): Fix operator responses appearing as verified and missed

JulianVentura:
# Fix operators responses appearing both as verified and missed at the same time

## Description

This PR fixes a bug on which the grafana dashboard is showing some operator responses both as missed and received at the same time.

On telemetry, we are storing active traces information on the `TraceStore`. This module implements an Agent which allow us to make crud operations to the traces. The problem with this is that those operations are not atomic, so if more than one tracing message is received by telemetry at the same time and for the same merkle root, then a write on write conflict may occur while trying to modify the trace information or when setting the current trace context for OpenTelemetry.

This PR fixes this by refactoring the `traces.ex` module, implementing a GenServer, which proceses each tracing message sequentially, eliminating the race condition.

#### Happy path

Here we test the normal behavior of the system

1. Start all services
2. Send some proofs
3. Check that traces are correctly generated on Jaeger.
4. Check that metrics are correctly updated on Grafana

#### Not responding operator

Here we test the system's behavior when one operator is not responding

1. Start all services, including 4 operators.
2. Send infinite proofs.
3. Check that traces are correctly generated on Jaeger.
4. Check that metrics are correctly updated on Grafana. You should see one of the operators (and just one) not responding to any task.

image

#### BLS service timeout

This following test scenario should not happen in mainnet, due to the really large configured bls task timeout. Anyways, we test it so we are sure the system is stable if it occurs.
For this, you should modify the `bls_service_task_timeout` parameter under `config-files/config-aggregator.yaml` with a small value, for instance:

``` yaml
aggregator:
  bls_service_task_timeout: 20s
```

Then, start the system as always:

1. Start all services.
2. Send infinite proofs.
4. Check that traces are correctly generated on Jaeger.
5. Check that metrics are correctly updated on Grafana.

You should see that some errors are thrown on the aggregator due to a BLS task expiration. Those batches should have an extra event in the aggregator jaeger logs, indicating the batch verification failure due to expiration. Anyways, the remaining logs should be there, showing that the batch was successfully verified.

image

On telemetry execution logs, you should see a "Context not found for 0x...". This is because when the task is finished, the context is erased from the internal storage structure of telemetry. Notice how in this capture, we first finish the trace sucessfully, and then a new finish request is received, logging that error message.

image

## Type of change

- [ ] New feature
- [X] Bug fix
- [ ] Optimization
- [X] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


telemetry_api/lib/telemetry_api/traces.ex
```diff
@@ -344,7 +186,285 @@ defmodule TelemetryApi.Traces do
...
+    else
+      {:error, information} -> 
+        {:reply, {:error, information}, traces}
+    end
+  end
+
+  @impl true
+  def handle_call({:create_batcher_task_trace, merkle_root}, _from, traces) do
+    with {:ok, total_stake} <- StakeRegistry.get_current_total_stake() do
+      root_span_ctx =
```
avilagaston9, line 306:
```suggestion
      Ctx.clear()
      root_span_ctx =
```

JulianVentura: That seems to fix the bug. Great!

PatStiles (approved):

telemetry_api/lib/telemetry_api/traces.ex
```diff
@@ -344,7 +186,327 @@ defmodule TelemetryApi.Traces do
...
+        })
+
+      {:reply, :ok, traces}
+    else
+      {:error, information} ->
+        {:reply, {:error, information}, traces}
+      {:error, error_type, message} ->
+        {:reply, {:error, error_type, message}, traces}
+      unknown ->
+        {:reply, {:error, unknown}, traces}
```
avilagaston9: We can have something like:
```Elixir
error -> handle_error(error)
```
To avoid repeating this logic everywhere.

avilagaston9 (approved):
Works as expected! Left a nit comment

telemetry_api/lib/telemetry_api/traces.ex
JulianVentura: Sure! I can add that if it makes it clearer

JulianVentura: Done!

JuArce (commented):
Working, I'm reviewing the code

JuArce (approved):

JuArce (changes requested):
Check the solution with @Oppen

Oppen (approved):

JuArce (approved):

PR #1698, opened 2025-01-02
chore: output `BatcherPaymentService` addresses as JSON

MegaRedHand:
# Output BatcherPaymentService addresses as JSON

## Description

This PR changes the `BatcherPaymentService` deployment script to follow the convention of outputting the addresses of deployed contracts in a JSON file.

## Type of change

Refactor

## Checklist

- [X] “Hotfix” to `testnet`, everything else to `staging`
- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/contracts/scripts/anvil/deploy_aligned_contracts.sh
+++ b/contracts/scripts/anvil/deploy_aligned_contracts.sh
@@ -38,13 +38,13 @@ forge script ../examples/verify/script/VerifyBatchInclusionCallerDeployer.s.sol 
 output_path=./script/output/devnet/batcher_deployment_output.json
 
 # Deploy Batcher Payments Contract
-forge_output=$(forge script script/deploy/BatcherPaymentServiceDeployer.s.sol \
+forge script script/deploy/BatcherPaymentServiceDeployer.s.sol \
     ./script/deploy/config/devnet/batcher-payment-service.devnet.config.json \
     $output_path \
     --rpc-url "http://localhost:8545" \
     --private-key "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80" \
     --broadcast \
-    --sig "run(string batcherConfigPath, string outputPath)")
+    --sig "run(string batcherConfigPath, string outputPath)"
 
 # Extract the batcher payment service values from the output
 # new_aligned_layer_service_manager_implementation=$(echo "$forge_output" | awk '/1: address/ {print $3}')
--- a/contracts/scripts/anvil/deploy_aligned_contracts.sh
+++ b/contracts/scripts/anvil/deploy_aligned_contracts.sh
@@ -47,7 +47,6 @@ forge script script/deploy/BatcherPaymentServiceDeployer.s.sol \
     --sig "run(string batcherConfigPath, string outputPath)"
 
 # Extract the batcher payment service values from the output
-# new_aligned_layer_service_manager_implementation=$(echo "$forge_output" | awk '/1: address/ {print $3}')
 batcher_payment_service_proxy=$(jq -r '.addresses.batcherPaymentService' $output_path)
 batcher_payment_service_implementation=$(jq -r '.addresses.batcherPaymentServiceImplementation' $output_path)
 
--- a/contracts/scripts/anvil/deploy_aligned_contracts.sh
+++ b/contracts/scripts/anvil/deploy_aligned_contracts.sh
@@ -38,10 +38,11 @@ forge script ../examples/verify/script/VerifyBatchInclusionCallerDeployer.s.sol 
 # Deploy Batcher Payments Contract
 forge_output=$(forge script script/deploy/BatcherPaymentServiceDeployer.s.sol \
     ./script/deploy/config/devnet/batcher-payment-service.devnet.config.json \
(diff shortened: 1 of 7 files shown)


contracts/scripts/anvil/deploy_aligned_contracts.sh
```diff
@@ -38,10 +38,11 @@ forge script ../examples/verify/script/VerifyBatchInclusionCallerDeployer.s.sol
...
     ./script/deploy/config/devnet/batcher-payment-service.devnet.config.json \
+    ./script/deploy/output/devnet/batcher_deployment_output.json \
     --rpc-url "http://localhost:8545" \
     --private-key "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80" \
     --broadcast \
-    --sig "run(string batcherConfigPath)")
+    --sig "run(string batcherConfigPath, string outputPath)")
 
 # Extract the batcher payment service values from the output
 # new_aligned_layer_service_manager_implementation=$(echo "$forge_output" | awk '/1: address/ {print $3}')
```
JuArce: Shouldn't we do the same change as in `[contracts/scripts/deploy_batcher_payment_service.sh](https://github.com/yetanotherco/aligned_layer/pull/1698/files#diff-4a582041a459061d1464141cee38003ede0385187a21ad63ffdb01dc84cdf8c5)`

Both solutions work, but just for keep consistency

```
batcher_payment_service_proxy=$(jq -r '.addresses.batcherPaymentService' $BATCHER_PAYMENT_SERVICE_OUTPUT_PATH)
batcher_payment_service_implementation=$(jq -r '.addresses.batcherPaymentServiceImplementation' $BATCHER_PAYMENT_SERVICE_OUTPUT_PATH)
```

contracts/scripts/deploy_batcher_payment_service.sh
```diff
@@ -10,23 +10,26 @@ cd ../
 
 source scripts/.env
 
+BATCHER_PAYMENT_SERVICE_OUTPUT_PATH=./script/deploy/output/devnet/batcher_deployment_output.json
```
JuArce: Why is it hardcoded?

contracts/scripts/anvil/deploy_aligned_contracts.sh
MegaRedHand: Yes. I didn't do it just to minimize the amount of changes. I'll add it in a bit

contracts/scripts/deploy_batcher_payment_service.sh
MegaRedHand: My mistake. I thought the config path was a local constant too.

uri-99:
```suggestion
BATCHER_PAYMENT_SERVICE_OUTPUT_PATH=./script/output/devnet/batcher_deployment_output.json
```

uri-99:
@MegaRedHand Are you aware the batcher payment addresses are currently being outputted in `alignedlayer_deployment_output.json` ?

contracts/scripts/deploy_batcher_payment_service.sh
uri-99: they are defined in the `contracts/scripts/.env`

MegaRedHand:
I saw the bash script uses `jq` to add those addresses to `alignedlayer_deployment_output.json`, but that doesn't work for me (and would fail if the format changes).

An alternative could be to do that from the Solidity script. I can change to that approach if that seems better.

contracts/scripts/deploy_batcher_payment_service.sh
MegaRedHand: Done

contracts/scripts/anvil/deploy_aligned_contracts.sh
MegaRedHand: Done. I also removed some stale comments (https://github.com/yetanotherco/aligned_layer/pull/1698/commits/0c0bbc8fb8ad07f21b8dc7a9c9b1ce60d4d9f6ef), lmk if that's OK

uri-99:
> I saw the bash script uses `jq` to add those addresses to `alignedlayer_deployment_output.json`, but that doesn't work for me (and would fail if the format changes).
> 
> An alternative could be to do that from the Solidity script. I can change to that approach if that seems better.

If the issue arises from the  usage of `jq` , and it is possible to have the BatcherPayment addresses written in another way, I would try to stick to the existing files structure.

Alternatevly, if we need to have these addresses in another file, we would have to make a big migration since many components read the BatcherPayment addresses from the existing output file.

MegaRedHand:
> If the issue arises from the usage of jq , and it is possible to have the BatcherPayment addresses written in another way, I would try to stick to the existing files structure.

Sure! Just to clarify, the changes in this PR don't touch the part that updates the `alignedlayer_deployment_output.json` file with the BPS addresses. The only difference is that those addresses are _also_ in `batcher_deployment_output.json`.

JuArce (approved):
lgtm

uri-99 (approved):

avilagaston9 (approved):
LGTM!

PR #1753, opened 2025-01-17
hotfix(aggregator): nil dereference for invalid signature

### LINKED ISSUE #1747: bug: malformed operator response can crash the aggregator
From cantina#32. Transcript:

```
Due to a nil dereference, a malicious operator can bring down an aggregator by nil dereference when verifying the bls key.

Inside SendSignedTaskResponseToAggregator(), a call is made to the aggregator.

The payload that is provided is from the following type:

type SignedTaskResponse struct {
	BatchMerkleRoot [32]byte
	SenderAddress [20]byte
	BatchIdentifierHash [32]byte
	BlsSignature    bls.Signature
	OperatorId      eigentypes.OperatorId
}
If we take a look further inside bls.Signature:

type Signature struct {
	*G1Point `json:"g1_point"`
}
We see that it is a pointer to G1Point.

This opens up the ability of any operator to force the receiving aggregator to panic.

Please follow along with the Proof of Concept below, which will demonstrate the panic on a running localnet.
```

Given we don't control the signature type and forking would be overkill, the fix consists in adding a `nil`-check.

Oppen:
An operator could trigger a `nil`-dereference in the aggregator and bring it down by sending a message with a `nil`-signature.
Catch this error, log a warning and return an error instead.

The log is level warning to avoid polluting with unnecessary stack traces.

Fixes #1747

## Type of change

- [x] Bug fix

## Checklist

- [x] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

## Testing

Apply the following patch:
```diff
diff --git a/operator/pkg/rpc_client.go b/operator/pkg/rpc_client.go
index 5726f14d..414f9c86 100644
--- a/operator/pkg/rpc_client.go
+++ b/operator/pkg/rpc_client.go
@@ -38,6 +38,10 @@ func NewAggregatorRpcClient(aggregatorIpPortAddr string, logger logging.Logger)
 // their signed task response.
 func (c *AggregatorRpcClient) SendSignedTaskResponseToAggregator(signedTaskResponse *types.SignedTaskResponse) {
 	var reply uint8
+
+	// @audit nil derefernce
+	signedTaskResponse.BlsSignature.G1Point = nil
+
 	for retries := 0; retries < MaxRetries; retries++ {
 		err := c.rpcClient.Call("Aggregator.ProcessOperatorSignedTaskResponseV2", signedTaskResponse, &reply)
 		if err != nil {
```

Then start sending proofs, the aggregator should log an error an keep going. In testnet this should crash the aggregator.

### DIFF
--- a/aggregator/pkg/server.go
+++ b/aggregator/pkg/server.go
@@ -51,7 +51,7 @@ func (agg *Aggregator) ProcessOperatorSignedTaskResponseV2(signedTaskResponse *t
 		"operatorId", hex.EncodeToString(signedTaskResponse.OperatorId[:]))
 
 	if signedTaskResponse.BlsSignature.G1Point == nil {
-		agg.logger.Error("invalid operator response with nil signature",
+		agg.logger.Warn("invalid operator response with nil signature",
 			"BatchMerkleRoot", "0x"+hex.EncodeToString(signedTaskResponse.BatchMerkleRoot[:]),
 			"SenderAddress", "0x"+hex.EncodeToString(signedTaskResponse.SenderAddress[:]),
 			"BatchIdentifierHash", "0x"+hex.EncodeToString(signedTaskResponse.BatchIdentifierHash[:]),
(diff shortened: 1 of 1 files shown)


JuArce:
It seems ok. But now I have the following question, shouldn't we add nil checks to all parameters from `ProcessOperatorSignedTaskResponseV2` function?

MauroToscano (approved):

JuArce (approved):

uri-99 (approved):

MauroToscano (approved):

PR #1756, opened 2025-01-20
hotfix: workaround gnark 0.10.0 unsoundness bug

### LINKED ISSUE #1749: bug: unsoundness risk in gnark 0.10
Reported in cantina#5. Transcript:

```
The Batcher accepts invalid proof in gnark groth16 proving system whenever at least two commitments are used in the circuit.

This breaks the soundness of the protocol whenever the circuit uses at least two commitments. Although using more than 1 commitment is relatively rare, malicious applications can intentionally add redundant commitments and inputs to their circuits without affecting the main logic of the circuit (as shown in PoC circuit below) and exploit this bug at any time.
```

The problem itself was known but slipped. PR #1471 attempted to fix it by upgrading to the recommended version, 0.11.0, but while testing it we found an OOM that lead to DoS. The fix was accepted upstream but hasn't yet been released. We closed the PR expecting a release soon after, but that didn't happen.

This is a critical bug: it's allegedly easy to exploit and compromises soundness of the network.

Possible solutions:
1. Disable (at least temporarily) the groth16 verifier;
2. Use a `replace` directive to point to (at least) the fix commit;
3. Push for a release upstream;
4. Possibly restrict the number of commitments to 1 (at least temporarily) so the error can't be triggered.

The solutions aren't necessarily mutually exclusive.

Oppen:
A lot of time passed since gnark v0.10.0 unsoundness bug[0] was reported
and fixed. We posponed the upgrade because the fixed release, v0.11.0,
contains another vulnerability, an OOM[1], for which a fix has been in
main since last November but no release appeared until now.

Our options here are limited, and none quite happy:
- We can `redirect` to a commit in `main`;
- We can disable groth16 verifiers from the network, which we currently
  use; or
- We can enforce that proofs have only one commitment, as the
  unsoundness can only be triggered with multiple commitments per proof.

This implements the latter option, being the least invasive one.

[0]: https://www.zellic.io/blog/gnark-bug-groth16-commitments
[1]: https://github.com/Consensys/gnark/security/advisories/GHSA-cph5-3pgr-c82g

Fixes #1749

## Type of change

- [x] Bug fix

## Checklist

- [x] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

## Testing

You'll need to compare the behavior of batcher and operators in both `testnet` and `hotfix/gnark_workaround` branches.

Download the attached file to your repo checkout and extract it with `tar xf fake_proof.tgz`. The contents are a witness, proof and verification key that should not be accepted by gnark but are, and they were generated with the following [PoC](https://gist.github.com/Oppen/3166ce50e3cd1922846d1cf4799be3e2). The PoC relies on a modified checkout of gnark, as explained in the document explaining the vulnerability.

Then apply this patch:
```diff
diff --git a/Makefile b/Makefile
index ad1ee3e7..4fc322dc 100644
--- a/Makefile
+++ b/Makefile
@@ -444,9 +444,9 @@ batcher_send_groth16_bn254_task: batcher/target/release/aligned
 	@echo "Sending Groth16Bn254 1!=0 task to Batcher..."
 	@cd batcher/aligned/ && cargo run --release -- submit \
 		--proving_system Groth16Bn254 \
-		--proof ../../scripts/test_files/gnark_groth16_bn254_infinite_script/infinite_proofs/ineq_1_groth16.proof \
-		--public_input ../../scripts/test_files/gnark_groth16_bn254_infinite_script/infinite_proofs/ineq_1_groth16.pub \
-		--vk ../../scripts/test_files/gnark_groth16_bn254_infinite_script/infinite_proofs/ineq_1_groth16.vk \
+		--proof ../../scripts/test_files/gnark_groth16_bn254_script/fake_groth16.proof \
+		--public_input ../../scripts/test_files/gnark_groth16_bn254_script/fake_groth16.pub \
+		--vk ../../scripts/test_files/gnark_groth16_bn254_script/fake_groth16.vk \
 		--proof_generator_addr 0x66f9664f97F2b50F62D13eA064982f936dE76657 \
 		--rpc_url $(RPC_URL) \
 		--network $(NETWORK)
```

Start a devnet and run:
```sh
make batcher_send_groth16_bn254_task
```

On `testnet` you should see a successful batch verification, as it's vulnerable to the unsoundness bug. On `hotfix/gnark_workaround` you should see both the batcher and operators (I suggest trying them separately with the rest of the stack running from `testnet`, to be able to test each of them) reject the proof, logging that it has too many commitments.

[fake_proof.tgz](https://github.com/user-attachments/files/18481737/fake_proof.tgz)

### DIFF
--- a/batcher/aligned-batcher/gnark/verifier.go
+++ b/batcher/aligned-batcher/gnark/verifier.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/consensys/gnark-crypto/ecc"
 	"github.com/consensys/gnark/backend/groth16"
+	bn254 "github.com/consensys/gnark/backend/groth16/bn254"
 	"github.com/consensys/gnark/backend/plonk"
 	"github.com/consensys/gnark/backend/witness"
 )
@@ -95,6 +96,17 @@ func verifyGroth16Proof(proofBytesRef C.ListRef, pubInputBytesRef C.ListRef, ver
 	proof := groth16.NewProof(curve)
 	if _, err := proof.ReadFrom(proofReader); err != nil {
 		log.Printf("Could not deserialize proof: %v", err)
+		return false
+	}
+
+	bn254Proof, ok := proof.(*bn254.Proof)
+	if !ok {
+		log.Printf("groth16 proof is not bn254")
+		return false
+	}
+	numCommitments := len(bn254Proof.Commitments)
+	if numCommitments > 1 {
+		log.Printf("too many commitments (%v) for groth16 proof (unsound for v0.10.0)", numCommitments)
 		return false
 	}
 
--- a/operator/pkg/operator.go
+++ b/operator/pkg/operator.go
@@ -31,6 +31,7 @@ import (
 	eigentypes "github.com/Layr-Labs/eigensdk-go/types"
 	"github.com/consensys/gnark-crypto/ecc"
 	"github.com/consensys/gnark/backend/groth16"
+	bn254 "github.com/consensys/gnark/backend/groth16/bn254"
 	"github.com/consensys/gnark/backend/plonk"
 	"github.com/consensys/gnark/backend/witness"
 	ethcommon "github.com/ethereum/go-ethereum/common"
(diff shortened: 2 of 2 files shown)


uri-99 (approved):

JuArce (commented):
There is missing to specify that limitation in the docs

uri-99:
> There is missing to specify that limitation in the docs

done

MauroToscano:
Gnark 0.12 fixes this, we should replace this PR for a bump of Gnark

https://github.com/Consensys/gnark/releases/tag/v0.12.0

PR #1764, opened 2025-01-20
refactor: simplify signature timeout logic

Oppen:
`ProcessNewSignature` already takes a context to handle cancellation and
timeout. We need just to create the context and pass it off, and it will
return an appropriate error if the timeout expires.

This saves us from creating a goroutine and handling messages just for
that.

Also improve a little bit of the error reporting.

## Type of change

- [x] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

## Testing

Basic testing: start a devnet, show it's able to verify proofs.
Alightly more involved testing: register three operators and leave only one up,
so it doesn't reach quorum, the timeout should kick.


uri-99 (approved):

JuArce (approved):

JuArce (changes requested):
This PR is wrong. Eigen's BLSAgg service only honors the context if it can start servicing the request

PR #1860, opened 2025-04-09
Checks functions refactor

MauroToscano:
# Checks functions

- Refactors all checks when receiving tasks from users, before starting to use locks into it's own functions
- Change the logic of checking max size of the verification data to get the whole size, and not only the proof 

## Type of change

- [x] Refactor
- [x] BugFix


MarcosNicolau (approved):
Nice, it is noticeably more readable now 👍

JuArce (approved):

PR #1872, opened 2025-04-14
feat(aggregation_mode): save last processed block in a file

### LINKED ISSUE #1870: feat(aggregation_mode): save last processed block in a file
# Description

**Current Behavior**

Now, when the aggregation_mode starts, it fetches blocks using 

```
# How far in the past should the service go to fetch batch logs
fetch_logs_from_secs_ago: 86400 # 24hs
# Anvil start with block time is 7 seconds
block_time_secs: 7
```

**New Behavior**

Instead of using time parameters, it should save to a file the latest processed block.

In the config `config-proof-aggregator.yaml` add a path to the file

The file should have the following yaml format

```yaml
last_processed_block: <>
```

The aggregation_mode should read this file and fetches events from the `last_processed_block` to the `current_block`. Once the process finishes, the file should be updated

Note: The `current_block` is the block when the process starts, do not confuse with the block when the process finishes

Mechanix97:
# Feat (aggregation mode) save last processed block in a file

## Description

From Issue #1870 

Save the last processed block in a json file which path is indicated in the argument config file.

Check if the last_processed_block is lower than the current block of the chain.

Add a method to the config to save the current last_processed_block (discarding the previous file)

## Type of change

- [ ] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ batcher/aligned/batch_inclusion_responses/*
 **/broadcast
 volume
 config-files/*.last_processed_batch.json
+config-files/*.last_processed_block.json
 
 nonce_*.bin
 
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ batcher/aligned/batch_inclusion_responses/*
 **/broadcast
 volume
 config-files/*.last_processed_batch.json
-config-files/*.last_processed_block.json
+config-files/*.last_aggregated_block.json
 
 nonce_*.bin
 
--- a/aggregation_mode/src/backend/config.rs
+++ b/aggregation_mode/src/backend/config.rs
@@ -52,7 +52,7 @@ impl Config {
         last_processed_block: u64,
     ) -> Result<(), Box<dyn std::error::Error>> {
         let last_processed_block_struct = LastProcessedBlock {
-            last_processed_block: last_processed_block,
+            last_processed_block,
         };
 
         let mut file = OpenOptions::new()
--- a/aggregation_mode/src/backend/config.rs
+++ b/aggregation_mode/src/backend/config.rs
@@ -8,8 +8,8 @@ pub struct ECDSAConfig {
 }
 
 #[derive(Debug, Deserialize, Serialize)]
-pub struct LastProcessedBlock {
-    pub last_processed_block: u64,
+pub struct LastAggregatedBlock {
+    pub last_aggregated_block: u64,
 }
 
 #[derive(Debug, Deserialize, Serialize)]
@@ -19,7 +19,7 @@ pub struct Config {
     pub max_proofs_in_queue: u16,
     pub proof_aggregation_service_address: String,
     pub aligned_service_manager_address: String,
-    pub last_processed_block_filepath: String,
+    pub last_aggregated_block_filepath: String,
     pub ecdsa: ECDSAConfig,
 }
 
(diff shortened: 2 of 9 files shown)


JuArce (approved):
lgtm

- It creates the file if it does not exist ✅
- It updates the file only if the verification was correctly sent to ethereum ✅
- It use the current block when the events were fetched and not when the process finishes, which is correct ✅

.gitignore
```diff
@@ -12,6 +12,7 @@ batcher/aligned/batch_inclusion_responses/*
 **/broadcast
 volume
 config-files/*.last_processed_batch.json
+config-files/*.last_processed_block.json
```
MauroToscano: Change this to last_aggregated_block.json, or the folder/name to reflect this is for aggregation

config-files/config-proof-aggregator.yaml
```diff
@@ -3,10 +3,7 @@ proof_aggregation_service_address: "0xcbEAF3BDe82155F56486Fb5a1072cb8baAf547cc"
 eth_rpc_url: "http://localhost:8545"
 eth_ws_url: "ws://localhost:8545"
 max_proofs_in_queue: 1000
-# How far in the past should the service go to fetch batch logs
-fetch_logs_from_secs_ago: 86400 # 24hs
-# Anvil start with block time is 7 seconds
-block_time_secs: 7
+last_processed_block_filepath: config-files/proof-aggregator.last_processed_block.json
```
MauroToscano: Idem the other suggestion

Mechanix97: fixed

.gitignore
Mechanix97: fixed

aggregation_mode/src/backend/config.rs
```diff
@@ -27,4 +31,32 @@ impl Config {
         let config: Config = serde_yaml::from_str(&contents)?;
         Ok(config)
     }
+
+    pub fn get_last_aggregated_block(&self) -> Result<u64, Box<dyn std::error::Error>> {
+        let mut file = File::open(&self.last_aggregated_block_filepath)?;
+        let mut contents = String::new();
+        file.read_to_string(&mut contents)?;
+        let lpb: LastAggregatedBlock = serde_json::from_str(&contents)?;
```
JuArce: It is using `LastAggregatedBlock` and `LastProcessedBlock`. Would be better to use only one concept (`LastAggregatedBlock`)

aggregation_mode/src/backend/config.rs
```diff
@@ -27,4 +31,32 @@ impl Config {
...
+        Ok(lpb.last_aggregated_block)
+    }
+
+    pub fn update_last_aggregated_block(
+        &self,
+        last_aggregated_block: u64,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let last_processed_block_struct = LastAggregatedBlock {
```
JuArce: It is using LastAggregatedBlock and LastProcessedBlock. Would be better to use only one concept (LastAggregatedBlock)

JuArce (commented):
Lgtm but this structure could be removed to avoid an extra level of indirection

```
#[derive(Debug, Deserialize, Serialize)]
pub struct LastAggregatedBlock {
    pub last_aggregated_block: u64,
}
```

MauroToscano (approved):

PR #1896, opened 2025-04-25
feat(aggregation_mode): aggregate proofs in chunks

MarcosNicolau:
## Description

Split the aggregation of proofs in chunks. For this, the aggregation program had to be split in two:
1. User proofs: this runs for each one of the chunks (each chunk is configured to have a max size of 512)
    - Aggregates proofs.
    - Commits the Merkle root of each chunk as a public input.
2. Chunk aggregator: 
    - Takes all the chunks proof along with their corresponding leaves (proofs commitments).
    - Verifies that each proof originated from the chunk aggregator program.
    - Verifies that the merkle root from the proof public input matches the merkle root from the provided proof leaves.
    - Computes the Merkle root of all leaves and commits it as a public input.
    - Produces the final aggregated proof that is verified on-chain.

Note: Because the root aggregator needs to verify that all input proofs were produced by the chunk aggregator, it must know the chunk aggregator’s program ID. This introduces a dependency between the two programs, requiring the root aggregator to be built after the chunk aggregator.

A script was created to automates this process. Running make `agg_mode_write_program_ids` now will:
1. Compile programs.
2. Copy the chunk aggregator's program ID and replace it in the root aggregator program as a constant.
3. Recompile programs (this step is necessary since changing the root aggregator's constant affects its own program ID).

**Considerations**
- `Hash32` type and `MerkleTreeBackend` is defined three times:
    - In the sdk
    - In sp1 root aggregator program
    - In Risc0 root aggregator program
  We should consider having a `types` crates to share this kind of types. We can tackle this once we re-define the structure for the rust projects to have a single workspace with a `crates` folder containing them all. 

**How to test**

1. Start a local devnet:

```shell
make ethereum_package_start
```

2. Run batcher
```
```

3. Send many proofs to the batcher:

```shell
# Run it as much as you want to process more chunks
make batcher_send_sp1_burst 
# Leave this in the background to generate batches with different merkle roots 
batcher_send_burst_groth16
```

4. Wait some time until enough proofs have been sent to run the proof aggregator with many chunks (consider lowering the `MAX_PROOFS_PER_AGGREGATION` variable to wait less):

```shell
make start_proof_aggregator_gpu_ethereum_package AGGREGATOR=sp1
make start_proof_aggregator_gpu_ethereum_package AGGREGATOR=risc0
```

5. You should see via the logs the number of chunks to be processed. You can also inspect the gpu usage via `nvtop`.

6. Finally, once the aggregator finishes, you can verify that one of the proofs you sent has been aggregated via the aligned-cli and sdk:
```
cd batcher/aligned
# For SP1
cargo run verify-agg-proof \
		--network devnet \
		--from-block 0 \
		--proving_system SP1 \
		--public_input ../../scripts/test_files/sp1/sp1_fibonacci_4_1_3.pub \
		--program-id-file ../../scripts/test_files/sp1/sp1_fibonacci_4_1_3.vk \
		--beacon_url http://localhost:58801 \
		--rpc_url http://localhost:8545

# For Risc0
cargo run verify-agg-proof \
		--network devnet \
		--from-block 0 \
		--proving_system Risc0 \
		--program-id-file ../../scripts/test_files/risc_zero/fibonacci_proof_generator/fibonacci_id_2_0.bin \
		--public_input ../../scripts/test_files/risc_zero/fibonacci_proof_generator/risc_zero_fibonacci_2_0.pub \
		--beacon_url http://localhost:58801 \
		--rpc_url http://localhost:8545
```
## Type of change

- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


MauroToscano (changes requested):
We need to double check the soundness of this

MauroToscano:
This is unsound, you need to check if the aggregated proof came from the right ELF, you can test hacking it with another program that has the right public inputs but it's not fro man Agg proof. Also, don't do something that commits and with an if you don't commit in one case, make it more explicit

MauroToscano (approved):

aggregation_mode/src/aggregators/mod.rs
```diff
@@ -11,6 +11,9 @@ use sha3::{Digest, Keccak256};
 use sp1_aggregator::{
     AlignedSP1VerificationError, SP1AggregationError, SP1ProofWithPubValuesAndElf,
 };
+use tracing::info;
+
+const MAX_PROOFS_PER_AGGREGATION: usize = 512;
```
JuArce: It could be a config parameter, but we can modify this in another PR

aggregation_mode/aggregation_programs/sp1/src/chunk_aggregator_main.rs
```diff
@@ -0,0 +1,50 @@
...
+    let input = sp1_zkvm::io::read::<ChunkAggregatorInput>();
+
+    let mut leaves = vec![];
+
+    // Verify the proofs.
+    for (proof, leaves_commitment) in input.proofs_and_leaves_commitment {
+        let vkey = proof.vk;
+        let public_values_digest = Sha256::digest(&proof.public_inputs);
+
+        // Ensure the aggregated chunk originates from the L1 aggregation program.
```
JuArce:
```suggestion
        // Ensure the aggregated chunk originates from the user proofs aggregation program.
```

contracts/script/deploy/config/devnet/proof-aggregator-service.devnet.config.json
```diff
@@ -1,13 +1,13 @@
 {
   "address": {
     "sp1VerifierAddress": "0x06530DD13aFB697B81E2d602a279EBAEC8Ca2E31",
-    "risc0VerifierAddress": "0x14C66A76b06A414e62048c50e4d21B295ad2Cb0C",
+    "risc0VerifierAddress": "0x663F3ad617193148711d28f5334eE4Ed07016602",
```
JuArce, line 4: Why does this address change?

MarcosNicolau: Honestly, I don't really know. But `Risc0` verification was reverting until I realize it was the address of the verifier.

aggregation_mode/src/aggregators/mod.rs
MarcosNicolau: Solved in [46ebd22](https://github.com/yetanotherco/aligned_layer/pull/1896/commits/46ebd22f42c5a49fc3e332082b1ddc4c2ebd1196).

JuArce (approved):

PR #1910, opened 2025-04-30
fix(aggregation_mode): hardcode image_id and vk_hash from solidity contract

MarcosNicolau:
## Description

1. The verifySp1 and verifyRisc0 functions no longer require `image_id` and `vk_hash` as input parameters. Instead, these values are now stored in the contract’s storage and can only be modified by the contract owner.
2. This change ensures that only trusted and verified programs can be used for verification.

Also, a new CLI binary was added to export the image_id and vk_hash to a JSON file to simplify the process of managing and registering valid programs.

## How to test

Run the proof aggregator with proving and verify that it works:
1. `make start_proof_aggregator AGGREGATOR=sp1`
2. `make start_proof_aggregator AGGREGATOR=risc0`

## Type of change

- [x] Bug fix

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


aggregation_mode/bin/write_program_image_id_vk_hash.rs
```diff
@@ -0,0 +1,42 @@
...
+        "sp1_vk_hash": format!("0x{}", sp1_vk_hash_hex),
+        "risc0_image_id": format!("0x{}", risc0_image_id_hex),
+    });
+
+    // Write to the file
+    fs::write(
+        &dest_path,
+        serde_json::to_string_pretty(&json_data).unwrap(),
+    )
+    .unwrap();
```
MarcosNicolau: I also considered printing it to the console instead of writing it to a file, as this is only available in the contracts deployment configs.

Makefile
```diff
@@ -202,6 +202,9 @@ verify_aggregated_proof_risc0_holesky_stage:
 install_aggregation_mode: ## Install the aggregation mode with proving enabled
 	cargo install --path aggregation_mode --features prove
```
JuArce:
```suggestion
	cargo install --path aggregation_mode --features prove --bin proof_aggregator
```

JuArce (commented):
Lgtm
I tested it changing the expected ELF and it works as expected rejecting the proof

If we merge first #1916 and #1918, we should regenerate `program_ids.json` to use the deterministic ones

MauroToscano (changes requested):
We should re review the hashes after deterministic compiling gets added on this PR for Risc0 (SP1 is already merged)

https://github.com/yetanotherco/aligned_layer/pull/1918

MarcosNicolau:
> We should re review the hashes after deterministic compiling gets added on this PR for Risc0 (SP1 is already merged)

The program ids with the new builds have been updated: [cdf6342](https://github.com/yetanotherco/aligned_layer/pull/1910/commits/cdf634201893efaf79bf63a5a138ceef66ce4bcb).

contracts/src/core/AlignedProofAggregationService.sol
```diff
@@ -115,4 +125,16 @@ contract AlignedProofAggregationService is
...
+    /// @param _risc0AggregatorProgramImageId The new imageid for the Risc0 aggregator program
+    function setRisc0AggregatorProgramImageId(bytes32 _risc0AggregatorProgramImageId) external onlyOwner {
+        risc0AggregatorProgramImageId = _risc0AggregatorProgramImageId;
+    }
+
+    /// @notice Sets the vk hash of the sp1 program
+    /// @param _sp1AggregatorProgramVKHash The new vk hash for the sp1 aggregator program
+    function setSP1AggregatorProgramVKHash(bytes32 _sp1AggregatorProgramVKHash) external onlyOwner {
+        sp1AggregatorProgramVKHash = _sp1AggregatorProgramVKHash;
+    }
```
MauroToscano, line 139: Add an event to log the hash changed

JuArce (approved):
Lgtm
I added to the `anvil_deploy_aligned_contracts` the step to update the program_ids and fixed the `install_aggregation_mode` target

MauroToscano (approved):

PR #1930, opened 2025-05-12
feat(aggregation_mode): add missing functions and events

MauroToscano:
# Add missing functions and events on agg contract

## Description

Add missing function to set the Risc0 verifier address and events on agg contract

## Type of change

- [x] New feature


MarcosNicolau (approved):

JuArce (commented):
Just a picky comment, I would declare first errors, events and then functions. It is not related, but as we are changing it we can change the order here

JuArce (approved):

PR #1955, opened 2025-05-30
feat/perf(aggregation-mode): verification & sp1 aggregation

MarcosNicolau:
## Description

While aggregating many proofs in sp1 we noticed that setting up the prover has quite the overhead (around `400ms`). This pr addresses that issue by:
- Parallelising the verification in the fetching process with rayon.
- Obtaining the vk once instead of the setting up the client with the program `elf` every time: for this the proof pre verification was moved to the creation of the struct so that the vk is computed only once and in parallel regarded if the pre-verification is enabled or not.

Other minor things have been added:
- A pre-verification flag: to skip the verification process in the fetching process
- Logs for sp1 cuda prover have been filtered to only show up to warn levels.

**Results**

With a 3090 the following results were observed:

The verification in the fetching process for a batch of 40 proofs takes:
- Staging: about 1 minute
- Head: about 20 seconds

Aggregation of 150 proofs in chunks of 128 took:
- Staging: about 20min.
- Head: about 6min.

## Type of change

- [x] New feature
- [x] Optimization

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/aggregation_mode/src/aggregators/sp1_aggregator.rs
+++ b/aggregation_mode/src/aggregators/sp1_aggregator.rs
@@ -37,17 +37,8 @@ pub enum AlignedSP1VerificationError {
 }
 
 impl SP1ProofWithPubValuesAndElf {
-    pub fn new(proof_with_pub_values: SP1ProofWithPublicValues, elf: Vec<u8>) -> Self {
-        let vk = vk_from_elf(&elf);
-
-        Self {
-            proof_with_pub_values,
-            elf,
-            vk,
-        }
-    }
-
-    pub fn new_with_verification(
+    /// Constructs a new instance of the struct by verifying a given SP1 proof with its public values.
+    pub fn new(
         proof_with_pub_values: SP1ProofWithPublicValues,
         elf: Vec<u8>,
     ) -> Result<Self, AlignedSP1VerificationError> {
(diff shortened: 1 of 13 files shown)


aggregation_mode/src/aggregators/sp1_aggregator.rs
```diff
@@ -13,25 +13,68 @@ const CHUNK_PROGRAM_ELF: &[u8] =
 const USER_PROOFS_PROGRAM_ELF: &[u8] =
     include_bytes!("../../aggregation_programs/sp1/elf/sp1_user_proofs_aggregator_program");
 
+#[allow(dead_code)]
```
MauroToscano: This seems like a missing feature flag, ignoring the dead code looks wrong

```suggestion
#[cfg(feature = "prove")]
```

aggregation_mode/src/main.rs
```diff
@@ -17,7 +17,9 @@ fn read_config_filepath_from_args() -> String {
 
 #[tokio::main]
 async fn main() {
-    let subscriber = FmtSubscriber::builder().finish();
+    // ignore spi_cude info logs
```
JuArce:
```suggestion
    // ignore sp1_cuda info logs
```

JuArce (approved):

MauroToscano (changes requested):
Remove the flag, and always pre verify, it has no cost and may avoid issues

MauroToscano (approved):

PR #1956, opened 2025-05-30
feat(aggregation-mode): limit number of proofs to fetch

MarcosNicolau:
## Description

Adds a limit for the number of proofs to fetch. When the fetcher reaches that limit or when the proofs for the batch surprass that limit the fetching stops and the last aggregated block us updated to the block number of that log so that the next aggregation starts at that block. 

**Considerations:**

Currently, we are limited by the blob size:
- Each blob has a capacity of (4096 * 32) = 131.072 bytes
- But, since we need to pad it with a 0x0 byte to not surpass the field modulus we are left with (4096 * 31) = 126.976
- Each proof commitments takes 32 bytes hash
 
So we can aggregate as much proofs as 126.976 / 32 = 3968 per blob.

We could increase the capacity by:
- Running the aggregator more frequently
- Adding the logic to send a vector of blobs instead of only one 

## Type of change

- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/aggregation_mode/src/backend/config.rs
+++ b/aggregation_mode/src/backend/config.rs
@@ -22,6 +22,7 @@ pub struct Config {
     pub last_aggregated_block_filepath: String,
     pub ecdsa: ECDSAConfig,
     pub proofs_per_chunk: u16,
+    pub total_proofs_limit: u16,
 }
 
 impl Config {
--- a/aggregation_mode/src/backend/fetcher.rs
+++ b/aggregation_mode/src/backend/fetcher.rs
@@ -55,6 +55,7 @@ impl ProofsFetcher {
     pub async fn fetch(
         &mut self,
         engine: ZKVMEngine,
+        limit: u16,
     ) -> Result<Vec<AlignedProof>, ProofsFetcherError> {
         // Get current block
         let current_block = self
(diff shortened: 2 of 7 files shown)


MauroToscano (approved):

JuArce (commented):
There is a case where you can DoS the aggregation mode
Let's suppose `total_proofs_limit = X`
If you have a batch with `n` proofs and `n > X`, the aggregation mode won't process the batch and every time it starts, it will try to process that batch, being stuck on that batch

MarcosNicolau:
> There is a case where you can DoS the aggregation mode
Let's suppose total_proofs_limit = X
If you have a batch with n proofs and n > X, the aggregation mode won't process the batch and every time it starts, it will try to process that batch, being stuck on that batch

You are right. I can think of two ways of solving this issue:
1. Save the last processed block + the index of the last processed proof: this way we don't have to process the whole batch at once, we can do it in pieces.
2. Make sure the batcher limit coincides with this one. Currently the batcher limit per batch is set to be `3000` so we should be covered.

JuArce (approved):

PR #1959, opened 2025-05-30
chore(aggregation-mode): prevent rebuilding programs if not changed

MarcosNicolau:
## Description

Since we introduced deterministic buildings with docker for the zkvm programs the proof aggregator started to take too much time to compile. This got even worse after the multilayer feature. This pr, recompiles the programs only if they have changed. For that, we've been orthodox and simply solved it by hashing the files + features.

This means that now the programs have to be compiled only once for each feature (prove, gpu)

To test it I suggest removing the `use_docker` options for faster builds in `build.rs`

## Type of change

- [x] Optimization

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/aggregation_mode/build.rs
+++ b/aggregation_mode/build.rs
@@ -6,7 +6,6 @@ use std::io::Read;
 use std::path::Path;
 use std::{env, fs};
 
-// Hash all inputs
 fn hash_files_and_features<P: AsRef<Path>>(paths: &[P], features: Vec<String>) -> String {
     let mut hasher = Keccak256::new();
     for path in paths {
@@ -61,7 +60,7 @@ fn main() {
                 output_directory: Some("./aggregation_programs/sp1/elf".to_string()),
                 // We use Docker to generate a reproducible ELF that will be identical across all platforms
                 // (https://docs.succinct.xyz/docs/sp1/writing-programs/compiling#production-builds)
-                // docker: true,
+                docker: true,
                 ..Default::default()
             }
         });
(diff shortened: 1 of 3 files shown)


aggregation_mode/build.rs
```diff
@@ -1,33 +1,91 @@
...
+        "aggregation_programs/sp1/src/user_proofs_aggregator_main.rs",
+        "aggregation_programs/sp1/src/chunk_aggregator_main.rs",
+        "aggregation_programs/risc0/Cargo.toml",
+        "aggregation_programs/risc0/src/user_proofs_aggregator_main.rs",
+        "aggregation_programs/risc0/src/chunk_aggregator_main.rs",
+        "aggregation_programs/risc0/src/lib.rs",
+    ];
+
+    for file in &programs {
+        println!("cargo:rerun-if-changed={}", file);
```
JuArce, line 42: These prints are not displayed

MarcosNicolau: > *Build scripts communicate with Cargo by printing to stdout. Cargo will interpret each line that starts with cargo:: as an instruction that will influence compilation of the package. All other lines are ignored.*

These flags indicate that the `build.rs` should run if the file outputted has been modified. They weren't really working for me that is why the hash solution was applied. Right now, they are there just in case. You can read more in the source.

Source: https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

JuArce (approved):

MauroToscano:
This should be done with a makefile, I'm making another pr to solve this issue:

https://github.com/yetanotherco/aligned_layer/pull/1962

PR #1961, opened 2025-06-03
fix(batcher): stop listening to blocks when one of the rpcs disconnects

MarcosNicolau:
## Description

This PR improves the batcher's ws connection. The batcher maintains two ws connections: a primary and a fallback. Previously, if either connection failed during `listen_to_new_blocks`, the entire process would fail. 

With this pr now:
- The batcher only returns an error if both connections fail. If at least one succeeds, the process continues.
- Previously, a select call would return immediately on the first event, when one disconnected then it would be the first one to return making it fail. The new logic listens to both connections and only exits if both fail.

The process now is the following:
1. Attempts to connect to both nodes.
2. Listens for new blocks from both connections.
3. If one connection fails, it continues listening on the other.
4. If both fail, it retries the connection process from step 1.

Note: when one fails then the connection we don't try to reconnect until both have failed. Adding this logic is not trivial at all as we would need to create s new process that handles it in the background and deal with mutex, etc.

**How to test**
1. Start `ethereum-pacakge`:
```shell
make ethereum_package_start
```
2. Start batcher:
```shell
```
3. Locate the rpcs container ids in docker handled by ethereum-package:
```shell
docker ps

# You should be looking for:
5ae8707b5dbf   ghcr.io/paradigmxyz/reth:latest                       "/usr/local/bin/reth…"   24 minutes ago   Up 15 minutes   0.0.0.0:8549->8549/tcp, 0.0.0.0:8549->8549/udp, 30303/tcp, 30303/udp, 0.0.0.0:8552->8545/tcp, 0.0.0.0:8553->8546/tcp, 0.0.0.0:8550->8551/tcp, 0.0.0.0:8551->9001/tcp   el-2-reth-lighthouse--3e21610d756d4ef588441314a9733ca1
180a0bcae8c2   ghcr.io/paradigmxyz/reth:latest                       "/usr/local/bin/reth…"   24 minutes ago   Up 19 minutes   0.0.0.0:8542->8542/tcp, 0.0.0.0:8545-8546->8545-8546/tcp, 0.0.0.0:8542->8542/udp, 30303/tcp, 30303/udp, 0.0.0.0:8543->8551/tcp, 0.0.0.0:8544->9001/tcp                 el-1-reth-lighthouse--9b08863b6524476092e0770690968857
```
4. Play around with starting and stopping the containers you should see the behavior explained above. 

## Type of change

- [x] Bug fix

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/batcher/aligned-batcher/src/lib.rs
+++ b/batcher/aligned-batcher/src/lib.rs
@@ -322,38 +322,70 @@ impl Batcher {
     pub async fn listen_new_blocks_retryable(
         self: Arc<Self>,
     ) -> Result<(), RetryError<BatcherError>> {
-        let eth_ws_provider = Provider::connect(&self.eth_ws_url).await.map_err(|e| {
-            warn!("Failed to instantiate Ethereum websocket provider");
-            RetryError::Transient(BatcherError::EthereumSubscriptionError(e.to_string()))
-        })?;
+        // Try to connect at least to one of the nodes (main or fallback)
+        let eth_ws_provider = Provider::connect(&self.eth_ws_url).await.ok();
+        let eth_ws_provider_fallback = Provider::connect(&self.eth_ws_url_fallback).await.ok();
+        if eth_ws_provider.is_none() {
+            warn!("Failed to instantiate Ethereum main websocket provider");
+        }
+        if eth_ws_provider_fallback.is_none() {
+            warn!("Failed to instantiate fallback Ethereum websocket provider");
+        }
+        if eth_ws_provider.is_none() && eth_ws_provider_fallback.is_none() {
+            return Err(RetryError::Transient(
+                BatcherError::EthereumSubscriptionError(
+                    "Both Ethereum websocket providers failed to connect".to_string(),
+                ),
+            ));
+        }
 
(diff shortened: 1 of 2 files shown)


batcher/aligned-batcher/src/lib.rs
```diff
@@ -322,39 +322,80 @@ impl Batcher {
...
-
-        let mut stream_fallback =
-            eth_ws_provider_fallback
-                .subscribe_blocks()
-                .await
-                .map_err(|e| {
-                    warn!("Error subscribing to blocks.");
-                    RetryError::Transient(BatcherError::EthereumSubscriptionError(e.to_string()))
-                })?;
+        // Tru to connect to one stream (main or fallback)
```
MarcosNicolau:
```suggestion
        // Try to connect to one stream (main or fallback)
```

MauroToscano:
Code seems fine

MauroToscano (approved):

JuArce (approved):

PR #1989, opened 2025-06-18
fix: update explorer URL printed after proof submission

maximopalopoli:
## Description

This PR updates the explorer URL printed after a proof submission. Also left a comment explaining that the default URL used is the one for mainnet, and in case the explorer's URL changes for another network, it should change here too.

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -622,9 +622,14 @@ async fn main() -> Result<(), AlignedError> {
             }
 
             for batch_merkle_root in unique_batch_merkle_roots {
+                let base_url = match submit_args.network.clone().into() {
+                    Network::Holesky => "https://holesky.explorer.alignedlayer.com/batches/0x",
+                    _ => "https://explorer.alignedlayer.com/batches/0x",
+                };
+
                 info!(
-                    "https://explorer.alignedlayer.com/batches/0x{}",
-                    hex::encode(batch_merkle_root)
+                    "{}",
+                    base_url.to_string() + hex::encode(batch_merkle_root).as_str()
                 );
             }
         }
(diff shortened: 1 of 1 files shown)


MauroToscano (approved):

MauroToscano (approved):
A follow up making the Network a reference would be nice, to avoid all this cloning. Maybe wallet can be a reference too

crates/cli/src/main.rs
```diff
@@ -622,9 +622,16 @@ async fn main() -> Result<(), AlignedError> {
             }
 
             for batch_merkle_root in unique_batch_merkle_roots {
+                let base_url = match submit_args.network.clone().into() {
+                    // Note: in case the explorer address changes for other networks, we should add an arm to this
+                    // match with that network since the default URL used here is the mainnet one
+                    Network::Holesky => "https://holesky.explorer.alignedlayer.com/batches/0x",
```
JuArce, line 626: There is missing to add the other networks
- Stage
- Mainnet
- MainnetStage
- Devnet

Generally we use devnet as default

JuArce (changes requested):

crates/cli/src/main.rs
maximopalopoli: Done in [c5b2f65](https://github.com/yetanotherco/aligned_layer/pull/1989/commits/c5b2f65c5c95a5edc3145bf8c333ff7bcc81298d)

JuArce (approved):

PR #2123, opened 2025-09-16
feat(batcher): poll for unlock events and remove users with unlocked funds

### LINKED ISSUE #2122: feat(batcher): poll for unlock events and remove users with unlocked funds
Lock logic order:

- User lock
- Batch lock

JuArce:
# Feat(batcher): Poll for unlock events and remove users with unlocked funds

## Description

This PR fetches `BalanceUnlocked` events from `BatcherPaymentService` to check if an user has proofs in the queue and has unlocked their funds. 
In that case, the batcher will remove their proofs until the user locks funds again to avoid a withdrawal and trying to discount balace from a user without funds deposited.

Closes #2122

## How to Test

1. Start anvil

2. Set `balance_unlock_polling_interval_seconds` to `60` in `config-files/config-batcher.yaml`. So the polling is executed every 1 minute

3. Start batcher

4. Deposit funds

```
aligned deposit-to-batcher \
    --amount 1ether \
    --private_key 0xdbda1821b80551c9d65939329250298aa3472ba22feea921c0cf5d620ea67b97 \
    --network devnet \
    --rpc_url http://localhost:8545
```

5. Send a proof with low fee

```
aligned submit \
    --proving_system Risc0 \
    --proof ./scripts/test_files/risc_zero/no_public_inputs/risc_zero_no_pub_input_3_0_3.proof \
    --vm_program ./scripts/test_files/risc_zero/no_public_inputs/risc_zero_no_pub_input_id_3_0_3.bin \
    --network devnet \
    --custom_fee_estimate 16 \
    --private_key 0xdbda1821b80551c9d65939329250298aa3472ba22feea921c0cf5d620ea67b97 \
    --random_address
```

You can send multiple proofs (less than 16 :p)

6. Unlock your funds

```
cast send 0x7bc06c482DEAd17c0e297aFbC32f6e63d3846650  "unlock()" \
    --rpc-url http://localhost:8545 \
    --private-key 0xdbda1821b80551c9d65939329250298aa3472ba22feea921c0cf5d620ea67b97
```

7. You should see logs like:

```
[2025-09-16T21:27:14Z INFO  aligned_batcher] Found 1 BalanceUnlocked events in blocks 770 to 870
[2025-09-16T21:27:14Z INFO  aligned_batcher] Processing BalanceUnlocked event for user: 0x23618e81e3f5cdf7f54c3d65f7fbc0abf5b21e8f
```

And the user should receive the error `User funds have been unlocked and proofs removed from queue.`

Other tests cases:

Unlock and lock funds. The batcher should not eject the user proofs

To lock funds again, use

```
cast send 0x7bc06c482DEAd17c0e297aFbC32f6e63d3846650  "lock()" \
    --rpc-url http://localhost:8545 \
    --private-key 0xdbda1821b80551c9d65939329250298aa3472ba22feea921c0cf5d620ea67b97
```

## Type of change

- [x] New feature
- [x] Bug fix

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/config-files/config-batcher.yaml
+++ b/config-files/config-batcher.yaml
@@ -37,3 +37,5 @@ batcher:
   # When replacing the message, how much higher should the max fee in comparison to the original one
   # The calculation is replacement_max_fee >= original_max_fee + original_max_fee * min_bump_percentage / 100
   min_bump_percentage: 10
+  # How often to poll for BalanceUnlocked events in seconds (default: 600 seconds = 10 minutes)
+  balance_unlock_polling_interval_seconds: 600
--- a/crates/batcher/src/lib.rs
+++ b/crates/batcher/src/lib.rs
@@ -678,7 +678,7 @@ impl Batcher {
             .batch_queue
             .extract_if(|entry, _| entry.sender == user_address);
 
-        // Notify user via websocket before removing the proofs
+        // Notify user via websocket
         for (entry, _) in removed_entries {
             if let Some(ws_sink) = entry.messaging_sink {
                 let ws_sink_clone = ws_sink.clone();
--- a/crates/batcher/src/lib.rs
+++ b/crates/batcher/src/lib.rs
@@ -7,10 +7,10 @@ use eth::utils::{calculate_bumped_gas_price, get_batcher_signer, get_gas_price};
 use ethers::contract::ContractError;
 use ethers::signers::Signer;
 use retry::batcher_retryables::{
-    cancel_create_new_task_retryable, create_new_task_retryable, get_user_balance_retryable,
-    get_user_nonce_from_ethereum_retryable, simulate_create_new_task_retryable,
-    user_balance_is_unlocked_retryable, get_current_block_number_retryable,
-    query_balance_unlocked_events_retryable,
+    cancel_create_new_task_retryable, create_new_task_retryable,
+    get_current_block_number_retryable, get_user_balance_retryable,
+    get_user_nonce_from_ethereum_retryable, query_balance_unlocked_events_retryable,
+    simulate_create_new_task_retryable, user_balance_is_unlocked_retryable,
 };
 use retry::{retry_function, RetryError};
 use tokio::time::{timeout, Instant};
@@ -40,7 +40,7 @@ use aligned_sdk::common::types::{
 
 use aws_sdk_s3::client::Client as S3Client;
 use eth::payment_service::{BatcherPaymentService, CreateNewTaskFeeParams, SignerMiddlewareT};
-use ethers::prelude::{Middleware, Provider, Http};
+use ethers::prelude::{Http, Middleware, Provider};
 use ethers::types::{Address, Signature, TransactionReceipt, U256, U64};
 use futures_util::{future, join, SinkExt, StreamExt, TryStreamExt};
 use lambdaworks_crypto::merkle_tree::merkle::MerkleTree;
@@ -333,7 +333,9 @@ impl Batcher {
             max_batch_proof_qty: config.batcher.max_batch_proof_qty,
             amount_of_proofs_for_min_max_fee: config.batcher.amount_of_proofs_for_min_max_fee,
             min_bump_percentage: U256::from(config.batcher.min_bump_percentage),
-            balance_unlock_polling_interval_seconds: config.batcher.balance_unlock_polling_interval_seconds,
+            balance_unlock_polling_interval_seconds: config
+                .batcher
+                .balance_unlock_polling_interval_seconds,
             last_uploaded_batch_block: Mutex::new(last_uploaded_batch_block),
             pre_verification_is_enabled: config.batcher.pre_verification_is_enabled,
             non_paying_config,
@@ -502,11 +504,13 @@ impl Batcher {
     /// Runs at configurable intervals and checks recent blocks for events (2x the polling interval).
     /// When an event is detected, removes user's proofs from queue and resets UserState.
     pub async fn poll_balance_unlocked_events(self: Arc<Self>) -> Result<(), BatcherError> {
-        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(self.balance_unlock_polling_interval_seconds));
-        
+        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(
+            self.balance_unlock_polling_interval_seconds,
+        ));
+
         loop {
             interval.tick().await;
-            
+
             if let Err(e) = self.process_balance_unlocked_events().await {
                 error!("Error processing BalanceUnlocked events: {:?}", e);
                 // Continue polling even if there's an error
@@ -528,37 +532,59 @@ impl Batcher {
         // Formula: interval / 12 * 2 (assuming 12-second block times, look back 2x the interval)
         let block_range = (self.balance_unlock_polling_interval_seconds / 12) * 2;
         let from_block = current_block.saturating_sub(U64::from(block_range));
-        
+
         // Query events with retry logic
-        let events = match self.query_balance_unlocked_events(from_block, current_block).await {
+        let events = match self
+            .query_balance_unlocked_events(from_block, current_block)
+            .await
+        {
             Ok(events) => events,
             Err(e) => {
-                warn!("Failed to query BalanceUnlocked events after retries: {:?}", e);
+                warn!(
+                    "Failed to query BalanceUnlocked events after retries: {:?}",
+                    e
+                );
                 return Ok(());
             }
         };
 
-        info!("Found {} BalanceUnlocked events in blocks {} to {}", 
-              events.len(), from_block, current_block);
+        info!(
+            "Found {} BalanceUnlocked events in blocks {} to {}",
+            events.len(),
+            from_block,
+            current_block
+        );
 
         // Process each event
         for event in events {
             let user_address = event.user;
-            info!("Processing BalanceUnlocked event for user: {:?}", user_address);
-            
+            info!(
+                "Processing BalanceUnlocked event for user: {:?}",
+                user_address
+            );
+
             // Check if user has proofs in queue
             if self.user_has_proofs_in_queue(user_address).await {
-                info!("User {:?} has proofs in queue, verifying funds are still unlocked", user_address);
-                
+                info!(
+                    "User {:?} has proofs in queue, verifying funds are still unlocked",
+                    user_address
+                );
+
                 // Double-check that funds are still unlocked by calling the contract
                 if self.user_balance_is_unlocked(&user_address).await {
                     info!("User {:?} funds confirmed unlocked, removing proofs and resetting UserState", user_address);
                     self.remove_user_proofs_and_reset_state(user_address).await;
                 } else {
-                    info!("User {:?} funds are now locked, ignoring stale unlock event", user_address);
+                    info!(
+                        "User {:?} funds are now locked, ignoring stale unlock event",
+                        user_address
+                    );
                 }
             } else {
-                info!("User {:?} has no proofs in queue, ignoring event", user_address);
+                info!(
+                    "User {:?} has no proofs in queue, ignoring event",
+                    user_address
+                );
             }
         }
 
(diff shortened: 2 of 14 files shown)


crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+            "Found {} BalanceUnlocked events in blocks {} to {}",
+            events.len(),
+            from_block,
+            current_block
+        );
+
+        // Process each event
+        for event in events {
+            let user_address = event.user;
+            info!(
```
MarcosNicolau: I would make this log a debug

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+                    info!("User {:?} funds confirmed unlocked, removing proofs and resetting UserState", user_address);
+                    self.remove_user_proofs_and_reset_state(user_address).await;
+                } else {
+                    info!(
+                        "User {:?} funds are now locked, ignoring stale unlock event",
+                        user_address
+                    );
+                }
+            } else {
+                info!(
```
MarcosNicolau, line 585: Ditto

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+                    "User {:?} has proofs in queue, verifying funds are still unlocked",
+                    user_address
+                );
+
+                // Double-check that funds are still unlocked by calling the contract
+                if self.user_balance_is_unlocked(&user_address).await {
+                    info!("User {:?} funds confirmed unlocked, removing proofs and resetting UserState", user_address);
+                    self.remove_user_proofs_and_reset_state(user_address).await;
+                } else {
+                    info!(
```
MarcosNicolau: Ditto

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+        for event in events {
+            let user_address = event.user;
+            info!(
+                "Processing BalanceUnlocked event for user: {:?}",
+                user_address
+            );
+
+            // Check if user has proofs in queue
+            if self.user_has_proofs_in_queue(user_address).await {
+                info!(
```
MarcosNicolau: Ditto

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+                    "User {:?} has proofs in queue, verifying funds are still unlocked",
+                    user_address
+                );
+
+                // Double-check that funds are still unlocked by calling the contract
+                if self.user_balance_is_unlocked(&user_address).await {
+                    info!("User {:?} funds confirmed unlocked, removing proofs and resetting UserState", user_address);
```
MarcosNicolau: This one is good in info, we don't expect to happen very often so it shouldn't be spammy at all

config-files/config-batcher.yaml
```diff
@@ -37,3 +37,5 @@ batcher:
   # When replacing the message, how much higher should the max fee in comparison to the original one
   # The calculation is replacement_max_fee >= original_max_fee + original_max_fee * min_bump_percentage / 100
   min_bump_percentage: 10
+  # How often to poll for BalanceUnlocked events in seconds (default: 600 seconds = 10 minutes)
+  balance_unlock_polling_interval_seconds: 600
```
MarcosNicolau, line 41: Should we add the formula here? And say that it should obviously be higher than the block time (12 seconds)

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+            );
+
+            // Check if user has proofs in queue
+            if self.user_has_proofs_in_queue(user_address).await {
+                info!(
+                    "User {:?} has proofs in queue, verifying funds are still unlocked",
+                    user_address
+                );
+
+                // Double-check that funds are still unlocked by calling the contract
```
MarcosNicolau: Not necessary and it might be obvious, but a better explanation of why we are doing this:

```suggestion
                // Double-check that funds are still unlocked by calling the contract
               // This is necessary because we query events over a block range, and the
              // user’s state may have changed (e.g., funds could be locked again) after
             // the event was emitted. Verifying on-chain ensures we don’t act on stale data.
```

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+
+    async fn process_balance_unlocked_events(&self) -> Result<(), BatcherError> {
+        // Get current block number using HTTP providers
+        let current_block = match self.get_current_block_number().await {
+            Ok(block) => block,
+            Err(e) => {
+                warn!("Failed to get current block number: {:?}", e);
+                return Ok(());
+            }
+        };
```
MarcosNicolau: We are already catching the error in the caller, so there is no need for the match and this can be simplified with `?`:

```suggestion
        let current_block = self.get_current_block_number().await?;
```

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+        {
+            Ok(events) => events,
+            Err(e) => {
+                warn!(
+                    "Failed to query BalanceUnlocked events after retries: {:?}",
+                    e
+                );
+                return Ok(());
+            }
+        };
```
MarcosNicolau: Ditto

```suggestion
        let events = self
            .query_balance_unlocked_events(from_block, current_block)
            .await?;
```

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+                )
+            },
+            ETHEREUM_CALL_MIN_RETRY_DELAY,
+            ETHEREUM_CALL_BACKOFF_FACTOR,
+            ETHEREUM_CALL_MAX_RETRIES,
+            ETHEREUM_CALL_MAX_RETRY_DELAY,
+        )
+        .await
+        .map_err(|e| {
+            error!("Failed to get current block number: {:?}", e);
```
MarcosNicolau: Remove this should be logged by the caller in my opinion

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+                )
+            },
+            ETHEREUM_CALL_MIN_RETRY_DELAY,
+            ETHEREUM_CALL_BACKOFF_FACTOR,
+            ETHEREUM_CALL_MAX_RETRIES,
+            ETHEREUM_CALL_MAX_RETRY_DELAY,
+        )
+        .await
+        .map_err(|e| {
+            error!("Failed to query BalanceUnlocked events: {:?}", e);
```
MarcosNicolau: Ditto

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+            .find(|(entry, _)| entry.sender == user_address)
+            .map(|(entry, _)| entry.clone())
+        {
+            // Notify user via websocket before removing the proof
+            if let Some(ws_sink) = entry.messaging_sink.as_ref() {
+                send_message(
+                    ws_sink.clone(),
+                    SubmitProofResponseMessage::UserFundsUnlocked,
+                )
+                .await;
```
MarcosNicolau: Spawn a task here as it might block a bit

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+        };
+
+        let mut batch_state_guard = batch_state_guard;
+
+        // Process all entries for this user
+        while let Some(entry) = batch_state_guard
+            .batch_queue
+            .iter()
+            .find(|(entry, _)| entry.sender == user_address)
+            .map(|(entry, _)| entry.clone())
```
MarcosNicolau: I think a single filter here is better, the find is looping through the whole queue after every found entry. Even better, you can use a `drain_filter` to drop the batch lock earlier and avoid cloning by consuming the entries.

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,243 @@ impl Batcher {
...
+                .try_user_lock_with_timeout(user_address, user_state.lock())
+                .await
+            {
+                let proofs_count = user_state_guard.proofs_in_batch;
+                user_state_guard.nonce -= U256::from(proofs_count);
+                user_state_guard.proofs_in_batch = 0;
+                user_state_guard.total_fees_in_queue = U256::zero();
+                user_state_guard.last_max_fee_limit = U256::max_value();
+                info!("Reset UserState for user {:?}", user_address);
+            } else {
```
MarcosNicolau: Can you just do a `user_state.remove(user_address)` or something similar?

MarcosNicolau (commented):
General flow worked alright for me

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,199 @@ impl Batcher {
...
+            }
+        } else {
+            false
+        }
+    }
+
+    async fn remove_user_proofs_and_reset_state(&self, user_address: Address) {
+        // Use timeout for batch lock
+        let mut batch_state_guard = match self
+            .try_batch_lock_with_timeout(self.batch_state.lock())
```
MauroToscano: If this lock fails we could end up in big trouble right ? We will miss the removal. I'd put a highish timeout (5 mins)

MauroToscano: Also a grafana, and maybe an alert ?

crates/batcher/src/lib.rs
```diff
@@ -491,6 +500,199 @@ impl Batcher {
...
+            .await
+        {
+            Some(guard) => guard,
+            None => {
+                warn!(
+                    "Failed to acquire batch lock for user {:?}, skipping removal",
```
MauroToscano:
```suggestion
                    "Failed to acquire batch lock when trying to remove proofs from user {:?}, skipping removal",
```

crates/batcher/src/lib.rs
MauroToscano: Agreed

crates/batcher/src/lib.rs
JuArce: I increased the timeout and added the metric. The alert can be done in a small PR (Issue: #2124)

crates/batcher/src/lib.rs
```diff
@@ -491,6 +505,209 @@ impl Batcher {
...
+            }
+        };
+
+        let removed_entries: Vec<_> = batch_state_guard
+            .batch_queue
+            .extract_if(|entry, _| entry.sender == user_address)
+            .collect();
+
+        // Notify user via websocket before removing the proofs
+        for (entry, _) in removed_entries {
```
MarcosNicolau: We could avoid an allocation by not collecting and using the iterator that `extract_if` returns

MarcosNicolau (approved):

MarcosNicolau:
Worked correctly for me following #2119 instructions, but the CI is failing for some reason

crates/batcher/src/lib.rs
JuArce: Addressed

JuArce:
Ci is working now

crates/batcher/src/lib.rs
```diff
@@ -491,6 +505,209 @@ impl Batcher {
...
+                self.metrics.inc_unlocked_event_polling_batch_lock_timeout();
+                return;
+            }
+        };
+
+        let removed_entries = batch_state_guard
+            .batch_queue
+            .extract_if(|entry, _| entry.sender == user_address);
+
+        // Notify user via websocket before removing the proofs
```
MauroToscano:
```suggestion
        // Notify user via websocket
```

crates/batcher/src/lib.rs
```diff
@@ -491,6 +505,214 @@ impl Batcher {
...
+                    // Close websocket connection
+                    let mut sink_guard = ws_sink_clone.write().await;
+                    if let Err(e) = sink_guard.close().await {
+                        warn!(
+                            "Error closing websocket for user {:?}: {:?}",
+                            user_address, e
+                        );
+                    } else {
+                        info!("Closed websocket connection for user {:?}", user_address);
+                    }
```
MauroToscano: I don't think there is a strong reason to close the connection in this error and not in others

JuArce: You are right. The client closes the connection correctly. I removed it

MauroToscano (approved):

PR #2151, opened 2025-10-09
fix(proof_aggregator): construct transactions following the new eip_7594 for blobs

MarcosNicolau:
## Description

After the Holesky Fusaka update, transactions containing blob sidecars are now being rejected with the following error:

```shell
unexpected eip-4844 sidecar after osaka
```

This occurs because blobs must now be sent according to the new [EIP-7594](https://eips.ethereum.org/EIPS/eip-7594) specification. However, Alloy does not yet support sending blobs in this new format. To work around this limitation, we have switched to using [Ethrex](https://github.com/lambdaclass/ethrex/pulls) for transaction submission.

**Notes**
- The Rust version in CI has been updated to 1.87.0, as Ethrex depends on features introduced up to that version.
- The beacon node must be run as a supernode (at least in Lighthouse; [see here](https://lighthouse-book.sigmaprime.io/help_bn.html?highlight=--supernode#beacon-node)) to properly retrieve blob data.

**How to Test**

Run the proof aggregator on a chain that includes the Fusaka update and verify that it continues to operate correctly.

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


JuArce:
Replaced by #2172

PR #2156, opened 2025-10-16
feat: improve periodic proof sending script

maximopalopoli:
## Description

This PR improves the previous script that sent proofs every N seconds, considering the time elapsed and the current network gas price:
- Every iteration occurs after 30 minutes
- Between 10 and 14 iterations, if the gas price is higher than 2 gwei, send a proof
- Between 14 and 16 iterations, if the gas price is higher than 5 gwei, send a proof
- Between 16 and 24 iterations, if the gas price is higher than 15 gwei, send a proof
- After 50 iterations, send a proof

## How to test

We are using devnet to test the feature. You can use your own wallet or the one used in the alerts/.env.devnet file.

If you are using a local wallet, fund it using:

```
cast send <address> --value 10ether --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 --rpc-url http://localhost:8545/
```

Deposit into Aligned using the following CLI command:

```
aligned deposit-to-batcher \
--network devnet \
--amount 5ether \
--keystore_path <KEYSTORE_PATH>
```

Or use the `--private_key` flag with the private key used in the .env.devnet file.

Now, you can run the script, but since the sleep time between intervals is 30 minutes, you'll have to reduce it, for example, to 10 seconds.

### Add gas cost variations

Since the gas cost won't vary much during the testing interval, we have to modify the gas cost of the network manually.

Do it with the following command:

```
curl -s http://localhost:8545/ -H "Content-Type: application/json" -d '{
"jsonrpc":"2.0","id":1,
"method":"anvil_setNextBlockBaseFeePerGas",
"params":["0x3b9aca00"]   // 1 Gwei in hex
}'
```

Change the Gwei amount in hex for the one you want to test. Some sample values:
- "0x3b9aca00" - 1 Gwei
- "0x9502F900" - 2.5 Gwei
- "0x165A0BC00" - 6 Gwei
- "0x41314CF00" - 17.5 Gwei
- "0xBA43B7400" - 50 Gwei

I noticed that once it's set to a value, it decreases with the passing of time, so I recommend increasing it to a high value and waiting for the time to pass.

Note: You'll see a log like the following

```
curl: (2) no URL specified
curl: try 'curl --help' or 'curl --manual' for more information
```

This is logged because it tries to connect to Slack via curl, but we have not set the Slack URL variable, as it's not relevant for this test.

## Type of change

- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


MarcosNicolau (commented):
Worked for me locally, not approving to discuss the params

MarcosNicolau (approved):

JuArce (approved):

alerts/periodic_sender.sh
```diff
@@ -0,0 +1,78 @@
...
+        ./alerts/sender_with_alert.sh "$ENV_FILE"
+        elapsed_intervals=0
+    elif { [ $elapsed_intervals -ge 16 ] && [ $elapsed_intervals -lt 24 ] && [ $current_gas_price -lt 15000000000 ]; }; then
+        # Between 16 and 24 elapsed intervals (8 to 12 hours), if gas price is below 15 gwei, send a proof
+        message="Sending proof at $elapsed_intervals with gas price $current_gas_price wei"
+        echo "$message"
+        ./alerts/sender_with_alert.sh "$ENV_FILE"
+        elapsed_intervals=0
+    elif { [ $elapsed_intervals -ge 50 ]; }; then
+        # After 50 elapsed intervals (25 hours), if gas price is below 50 gwei, send a proof
```
JuArce:
```suggestion
        # After 50 elapsed intervals (25 hours) send a proof
```

JuArce (changes requested):

JuArce (approved):

PR #2179, opened 2025-12-01
feat(aggregation): have constant costs proof sending

maximopalopoli:
## Description

Description of the pull request changes and motivation.

### Unit test

You can run the unit test cases with:

```sh
make proof_aggregator_test_without_compiling_agg_programs
```

This target avoids recompiling for the tests when the aggregation programs are already compiled (to be used when the programs are not being tested).

### Ethereum package

Another way to test this is to test the proof aggregation service, which can be done in the following way:

1. Start Ethereum package:
```shell
make ethereum_package_start
```
2. Start batcher:
```shell
```
3. Send SP1 and Risc0 proofs:
```shell
make batcher_send_sp1_burst BURST_SIZE=1
make batcher_send_risc0_burst BURST_SIZE=1
```
4. Run proof aggregator for SP1 and RISC0:
```shell
make proof_aggregator_start AGGREGATOR=sp1
make proof_aggregator_start AGGREGATOR=risc0
```
5. Verify the proofs have been aggregated:
```shell
make verify_aggregated_proof_sp1 FROM_BLOCK=0
make verify_aggregated_proof_risc0 FROM_BLOCK=0
```

### Submission skipping with anvil
To test that the proof is skipped when the gas cost is too high, you should use Anvil to test, following these steps:

1. Start the anvil node:
```shell
make anvil_start
```
2. Start the batcher:
```shell
make batcher_start_local ENVIRONMENT=devnet
```
3. Send a proof to the batcher:
```shell
make batcher_send_sp1_burst BURST_SIZE=1
```
4. Start the proof aggregator:
```shell
make proof_aggregator_start AGGREGATOR=sp1
```
5. Before the proof is sent, increase the base gas cost by making a post to the anvil node:
```shell
curl -s http://localhost:8545/ -H "Content-Type: application/json" -d '{ "jsonrpc":"2.0","id":1, "method":"anvil_setNextBlockBaseFeePerGas", "params":["0xffffffff000"] }'
```
You should see the log `Skipping sending proof to ProofAggregationService contract due to budget/time constraints.`.
The proof submission to the contract will fail because the anvil node does not work for the Fusaka fork, but with these steps, we were testing that the skipping works.

## Type of change

- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Refactor

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


MauroToscano (approved):

aggregation_mode/src/backend/mod.rs
```diff
@@ -136,23 +141,88 @@ impl ProofAggregator {
...
+        expected_cost_in_wei <= max_to_spend_in_wei
+    }
+
     async fn send_proof_to_verify_on_chain(
         &self,
-        blob: BlobTransactionSidecar,
+        blob: &BlobTransactionSidecar,
         blob_versioned_hash: [u8; 32],
-        aggregated_proof: AlignedProof,
+        aggregated_proof: &AlignedProof,
```
MarcosNicolau: Why are you passing it as reference? You end up cloning everything and it isn't so cheap in this case. Better if we can avoid it.

aggregation_mode/src/backend/fetcher.rs
```diff
@@ -188,4 +189,15 @@ impl ProofsFetcher {
...
+    /// Try to obtain a sensible gas price from two providers.
+    /// Tries `primary` first, falls back to `fallback` if the first fails.
+    pub async fn get_gas_price(&self) -> Result<u128, ProofsFetcherError> {
+        match self.rpc_provider.get_gas_price().await {
+            Ok(price) => Ok(price),
+            Err(e1) => Err(ProofsFetcherError::GasPriceError(format!(
+                "gas price error: {e1}"
+            ))),
+        }
+    }
```
MarcosNicolau: This function is unnecessary, it should be an in place call instead of creating a wrapper around it (notice that you are doing `map_err` anyways. Also, the `ProofFetcher` is responsible for fetching the proofs from the fast mode why did you add it here?

aggregation_mode/src/backend/mod.rs
```diff
@@ -136,23 +141,88 @@ impl ProofAggregator {
...
-        );
+        // Iterate until we can send the proof on-chain
+        let start_time = Instant::now();
+
+        let mut sent_proof = false;
+        while !sent_proof {
+            // We add 24 hours because the proof aggregator runs once a day, so the time elapsed
+            // should be considered over a 24h period.
+            let time_elapsed: Duration =
+                Instant::now().duration_since(start_time) + Duration::from_secs(24 * 3600);
```
MarcosNicolau: The aggregator is meant to be run as a command, that is, it doesn't have a concept of time to wait before sending proof for now. One should be able to run it at any point to aggregate and post the proofs as soon as it finishes if the gas price is ok.

The aggregator already has a systemd service that runs every 24hs.

aggregation_mode/src/backend/mod.rs
```diff
@@ -136,23 +141,88 @@ impl ProofAggregator {
...
+                    receipt.transaction_hash
+                );
+
+                sent_proof = true;
+            } else {
+                info!("Skipping sending proof to ProofAggregationService contract due to budget/time constraints.");
+            }
+
+            // Sleep for 5 minutes before re-evaluating
+            sleep(Duration::from_secs(300));
```
MarcosNicolau: Please add the number of blocks (25). 

Maybe we should re-consider this number knowing that every block the gas price can increase/decrease a 12,5% depending on the traffic.

aggregation_mode/src/backend/mod.rs
```diff
@@ -136,23 +141,88 @@ impl ProofAggregator {
...
+        let start_time = Instant::now();
+
+        let mut sent_proof = false;
+        while !sent_proof {
+            // We add 24 hours because the proof aggregator runs once a day, so the time elapsed
+            // should be considered over a 24h period.
+            let time_elapsed: Duration =
+                Instant::now().duration_since(start_time) + Duration::from_secs(24 * 3600);
+
+            let gas_price = self.fetcher.get_gas_price().await.map_err(FetchingProofs)?;
```
MarcosNicolau: Ditto

aggregation_mode/src/backend/mod.rs
```diff
@@ -136,23 +141,88 @@ impl ProofAggregator {
...
-            .await?;
-        info!(
-            "Proof sent and verified, tx hash {:?}",
-            receipt.transaction_hash
-        );
+        // Iterate until we can send the proof on-chain
+        let start_time = Instant::now();
+
+        let mut sent_proof = false;
+        while !sent_proof {
```
MarcosNicolau: Use a `loop` and then break after the proof is sent

aggregation_mode/src/backend/mod.rs
```diff
@@ -136,23 +141,88 @@ impl ProofAggregator {
...
+
+    /// Decides whether to send the aggregated proof to be verified on-chain based on
+    /// time elapsed since last submission and monthly ETH budget.
+    /// We make a linear function with the eth to spend this month and the time elapsed since last submission.
+    /// If eth to spend / elapsed time is over the linear function, we skip the submission.
+    fn should_send_proof_to_verify_on_chain(
+        &self,
+        time_elapsed: Duration,
+        monthly_eth_budget: f64,
+        gas_price_in_wei: U256,
```
MarcosNicolau: The gas price is assumed to be in wei

aggregation_mode/src/backend/mod.rs
```diff
@@ -136,23 +141,88 @@ impl ProofAggregator {
...
+        const SECONDS_PER_MONTH: u64 = 30 * 24 * 60 * 60;
+
+        let monthly_budget_in_wei = Self::floating_eth_to_wei(monthly_eth_budget);
+
+        let elapsed_seconds = U256::from(time_elapsed.as_secs());
+
+        let budget_available_per_second_in_wei = monthly_budget_in_wei / SECONDS_PER_MONTH;
+
+        budget_available_per_second_in_wei * elapsed_seconds
+    }
```
MarcosNicolau: I would move this from here, it can be another entity or a helper file but don't bloat the `ProofAggregator` struct.

aggregation_mode/src/backend/mod.rs
MarcosNicolau: You can leave it though, It's ok

aggregation_mode/src/backend/mod.rs
```diff
@@ -136,23 +141,88 @@ impl ProofAggregator {
...
+        }
 
         Ok(())
     }
 
+    fn floating_eth_to_wei(eth: f64) -> U256 {
+        let wei_in_eth = 1_000_000_000_000_000_000f64;
+        let wei = eth * wei_in_eth;
+        U256::from(wei as u64)
+    }
```
MarcosNicolau: You can use `parse_units` from `alloy`.

MarcosNicolau: or `parse_ether`

aggregation_mode/src/backend/mod.rs
```diff
@@ -148,6 +183,40 @@ impl ProofAggregator {
         Ok(())
     }
 
+    fn max_to_spend_in_wei(time_elapsed: Duration, monthly_eth_budget: f64) -> U256 {
+        const SECONDS_PER_MONTH: u64 = 30 * 24 * 60 * 60;
+
+        // Note: this unwrap is safe because parse_ether only fails for negative numbers or invalid strings
+        let monthly_budget_in_wei = parse_ether(monthly_eth_budget).unwrap_or(U256::zero());
```
MarcosNicolau: I would unwrap or a proper `expect` messge here. And instead, check on program initialization that `monthly_budget_in_wei` has been provided and log the value as well.

aggregation_mode/src/backend/mod.rs
```diff
@@ -136,6 +146,31 @@ impl ProofAggregator {
             hex::encode(blob_versioned_hash)
         );
 
+        // Iterate until we can send the proof on-chain
+        let mut time_elapsed: Duration = Duration::from_secs(24 * 3600);
```
MarcosNicolau:
```suggestion
        let mut time_elapsed = Duration::from_secs(24 * 3600);
```

config-files/config-proof-aggregator-mock-ethereum-package.yaml
```diff
@@ -12,7 +12,8 @@ proofs_per_chunk: 512 # Amount of proofs to process per chunk
 # Since each proof commitments takes 32 bytes hash
 # We can aggregate as much proofs as 126.976 / 32 = 3968 per blob
 total_proofs_limit: 3968
-
+# Monthly ETH budget for on-chain proof verification
+monthly_budget_eth: 15.0
```
MarcosNicolau: Please explain how to set this value and some examples

MarcosNicolau (commented):
It worked well for me. I'll approve it after solving the comments.

config-files/config-proof-aggregator-mock-ethereum-package.yaml
maximopalopoli: Done in [c1cd90a](https://github.com/yetanotherco/aligned_layer/pull/2179/commits/c1cd90ae44a5a6afc588a1957195b9331da51df7)

aggregation_mode/src/backend/mod.rs
maximopalopoli: Done in [7a391f8](https://github.com/yetanotherco/aligned_layer/pull/2179/commits/7a391f8f7870575711a534d978750e1175a48ae6)

JuArce (changes requested):
We should not use ethers

aggregation_mode/src/backend/mod.rs
```diff
@@ -136,6 +152,36 @@ impl ProofAggregator {
...
+        loop {
+            // We add 24 hours because the proof aggregator runs once a day, so the time elapsed
+            // should be considered over a 24h period.
+
+            let gas_price = match self.rpc_provider.get_gas_price().await {
+                Ok(price) => Ok(price),
+                Err(e1) => Err(AggregatedProofSubmissionError::GasPriceError(
+                    e1.to_string(),
+                )),
+            }?;
```
MarcosNicolau:
```suggestion
            let gas_price = match self.rpc_provider.get_gas_price().await.map_err(|e| AggregatedProofSubmissionError::GasPriceError(
                    e.to_string(),
                ))?;
```

aggregation_mode/src/backend/mod.rs
```diff
@@ -61,6 +61,14 @@ impl ProofAggregator {
         .expect("Keystore signer should be `cast wallet` compliant");
         let wallet = EthereumWallet::from(signer);
 
+        // Check if the monthly budget is non-negative to avoid runtime errors later
+        let monthly_budget_in_wei = parse_ether(config.monthly_budget_eth).expect("Monthly budget must be a non-negative value");
+        
+        info!(
+            "Monthly budget set to {} wei",
+            monthly_budget_in_wei
+        );
```
MarcosNicolau: Show it in ether please

maximopalopoli: Done in [329c65d](https://github.com/yetanotherco/aligned_layer/pull/2179/commits/329c65dc2e553a750d818abdd73e8cec4e98adbd)

aggregation_mode/src/backend/mod.rs
```diff
@@ -284,3 +364,113 @@ impl ProofAggregator {
...
+        ProofAggregator::new(config)
+    }
+
+    #[test]
+    fn test_should_send_proof_to_verify_on_chain_updated_cases() {
+        let aggregator = make_aggregator();
+
+        let gas_price_20gwei: U256 = U256::from(20_000_000_000u64);
+
+        // With 0 seconds elapsed and 1 ETH budget, we cannot send the proof
```
JuArce: It is not clear what are we testing.
It should be better to specify the parameter with numbers instead of "very low budget"

maximopalopoli: Improved in [893209d](https://github.com/yetanotherco/aligned_layer/pull/2179/commits/893209d94b0c0afff26ab26df611be0361657051)

aggregation_mode/proof_aggregator/src/backend/mod.rs
```diff
@@ -302,3 +423,146 @@ impl ProofAggregator {
...
+
+    #[test]
+    fn test_should_send_proof_to_verify_on_chain_updated_cases() {
+        let aggregator = make_aggregator();
+
+        // The should_send_proof_to_verify_on_chain function returns true when:
+        // gas_price * 600_000 <= (seconds_elapsed) * (monthly_eth_budget / (30 * 24 * 60 * 60))
+
+        // Case 1: Base case -> should return true
+        // Monthly Budget: 0.15 ETH -> 0.5 ETH per day -> 0.000058 ETH per hour
```
JuArce: This is wrong

0.15 ETH monthly -> 0.15/30 = 0.005 eth per day -> 0.000208 eth per hour

JuArce (changes requested):

JuArce (approved):

aggregation_mode/proof_aggregator/src/backend/fetcher.rs
```diff
@@ -33,6 +33,25 @@ pub struct ProofsFetcher {
...
+        );
+
+        let last_aggregated_block = 0;
+
+        Self {
+            rpc_provider,
+            aligned_service_manager,
+            last_aggregated_block,
+        }
+    }
```
MarcosNicolau, line 53:
```suggestion
    pub fn new_for_testing(config: &Config) -> Self {
      let mut proof_fetcher = self.new();
      proof_fetcher.last_aggregated_block = 0;
      proof_fetcher
    }
```

MarcosNicolau (approved):

aggregation_mode/proof_aggregator/build.rs
```diff
@@ -1,8 +1,29 @@
...
+            println!("cargo:warning=SKIP_AGG_PROGRAMS_BUILD=1: methods.rs already exists, skipping aggregation programs build");
+            return;
+        } else {
+            println!(
+                "cargo:warning=SKIP_AGG_PROGRAMS_BUILD=1 set, but {path} does not exist, running full build",
+                path = methods_path.display()
+            );
+        }
+    }
+
```
MarcosNicolau, line 29: I would move this to a function: `should_skip_build() -> bool`

aggregation_mode/proof_aggregator/src/backend/fetcher.rs
maximopalopoli: This is done this way to avoid having to read from the files, which was causing conflicts in the CI. Instead of reading from the file we just set it to zero (since it isn’t important for what we’re testing). If we call the other constructor, then we’ll get the error that led us to create this new_for_testing in the first place.

aggregation_mode/proof_aggregator/build.rs
maximopalopoli: Done in [675540b](https://github.com/yetanotherco/aligned_layer/pull/2179/commits/675540b03aeb6644bb353246885c7bd8cd96b8fb)

PR #2210, opened 2025-12-17
feat(aggregation-mode): handle processing that have failed in the aggregation process

MarcosNicolau:
## Description

When aggregating proofs, we mark all the proofs to aggregate to `processing` so no other proof aggregator aggregates them. The problem is that in the case the proof aggregator fails at any step, the tasks status would stay in `processing` and no other process would take them to aggregate. 

The presented solution to this is:
1. After an err the proof aggregator will try to set their status back to `pending` again, so it become available for other aggregators
2. If `1.` fails or the machine was shut down in the middle of the process, then we check the new `status_updated_at` flag in the task. If 12 hours have passed since the update to `processing` then it is also valid for the aggregator to take.

**How to test**
1. Start ethereum package:
```shell
make ethereum_package_start
```
2. Start gateway
```shell
make agg_mode_gateway_start_ethereum_package
agg_mode_payments_poller_start_ethereum_package
```
3. Send proofs
```shell
make agg_mode_gateway_send_payment
make agg_mode_gateway_send_sp1_proof
```
4. Update the code to force an error in proof aggregator, for example in `aggregation_mode/proof_aggregator/src/backend/merkle_tree.rs`:
```rust
pub fn compute_proofs_merkle_root(
    proofs: &[AlignedProof],
) -> Option<(MerkleTree<AlignedProof>, Vec<[u8; 32]>)> {
   None
}
```
5. Start proof aggregator:
```shell
make proof_aggregator_start_ethereum_package
```
6. See that the proof aggregator fails and sets the proofs in pending status again. 
7. Another test you can do is:
   - Go to adminer in `http://localhost:8090`
   - Select the tasks and update one to `processing` status and set the timestamp to a number that is 12 hours in the past from now (you can do that [here](https://www.epochconverter.com/)).
   - Run the proof aggregator and it should fetch the task
   - You can run the same one but setting it to a timestamp that is less than 12 hours in the past, in this case the proof aggregator should not select the task.

## Type of change

- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue


JuArce (approved):

aggregation_mode/proof_aggregator/src/backend/db.rs
```diff
@@ -32,12 +32,19 @@ impl Db {
             "WITH selected AS (
                     SELECT task_id
                     FROM tasks
-                    WHERE proving_system_id = $1 AND status = 'pending'
+                    WHERE proving_system_id = $1
+                      AND (
+                        status = 'pending'
+                        OR (
+                            status = 'processing'
+                            AND status_updated_at <= now() - interval '12 hours'
```
JuArce, line 51: I would add this as documentation of the function

JuArce (changes requested):

aggregation_mode/proof_aggregator/src/backend/db.rs
MauroToscano: Rename the function to ```get_tasks_to_process_and_update_their_status()```

And document it

MarcosNicolau: Solved in [9783115](https://github.com/yetanotherco/aligned_layer/pull/2210/commits/97831157011209061559af524a2b05bcad93e676).

maximopalopoli (approved):

aggregation_mode/db/migrations/001_init.sql
```diff
@@ -9,7 +9,8 @@ CREATE TABLE tasks (
     merkle_path BYTEA,
     status task_status DEFAULT 'pending',
     nonce BIGINT NOT NULL,
-    inserted_at TIMESTAMPTZ NOT NULL DEFAULT now()
+    inserted_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+    status_updated_at TIMESTAMPTZ DEFAULT now()
```
JuArce: We should move it to a migration

MarcosNicolau: Solved in [b388f1c](https://github.com/yetanotherco/aligned_layer/pull/2210/commits/b388f1c7a6dca4628bb364014679dd02ac775853).

aggregation_mode/db/src/types.rs
```diff
@@ -21,6 +24,7 @@ pub struct Task {
     pub program_commitment: Vec<u8>,
     pub merkle_path: Option<Vec<u8>>,
     pub status: TaskStatus,
+    pub status_updated_at: Option<DateTime<Utc>>,
```
JuArce:
```suggestion
    pub status_updated_at: DateTime<Utc>,
```

MarcosNicolau: Solved in [0b01101](https://github.com/yetanotherco/aligned_layer/pull/2210/commits/0b011019a17ca2ed726428b13bc29249d760302e).

JuArce (approved):

PR #2239, opened 2026-01-09
feat(aggregation mode): bind gateway server to Non-TLS port when using TLS

JuArce:
# feat(aggregation mode): bind gateway server to non tls port when using tls

## Description

This PR just binds the server not only to 443 port for TLS but also to port 8080 to connect through VPN without using the public domain

## Type of change

- [x] New feature

## Checklist

- [ ] Linked to Github Issue
- [ ] Unit tests added
  - [ ] Benchmarks added/run
- [ ] Has a known issue

### DIFF
--- a/aggregation_mode/gateway/src/http.rs
+++ b/aggregation_mode/gateway/src/http.rs
@@ -86,7 +86,7 @@ impl GatewayServer {
 
     pub async fn start(&self) {
         // Note: GatewayServer is thread safe so we can just clone it (no need to add mutexes)
-        let port = self.config.port;
+        let http_port = self.config.port;
         let state = self.clone();
 
         // Note: This creates a new Prometheus server different from the one created in GatewayServer::new. The created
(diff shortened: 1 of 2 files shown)


MarcosNicolau (approved):

aggregation_mode/gateway/src/http.rs
```diff
@@ -121,21 +109,48 @@ impl GatewayServer {
...
+            );
 
-        server.run().await.expect("Server to never end");
+            server
+                .bind((self.config.ip.as_str(), http_port))
+                .expect("To bind HTTP socket correctly")
+                .run()
+                .await
+                .expect("Server to never end");
+        }
```
maximopalopoli: I think that the HTTP socket part is common to both arms, so the only thing inside the if should be the HTTPS socket binding. I mean, the logic should be the following:
1. Create the server as it's being done
2. If the TLS feature flag is set, bind the HTTPS port
3. Bind the HTTP socket and run the server.

JuArce: Done in #[1bba485](https://github.com/yetanotherco/aligned_layer/pull/2239/commits/1bba4853eeb336e72e64a68c90bb8cedb9f765e9)

maximopalopoli (approved):

PR #2281, opened 2026-04-13
refactor: contracts

MarcosNicolau:
## Summary
- Encode `validFrom` in the Merkle leaf so each claim leaf represents a vesting stage gated by timestamp.
- Add `claimBatch` entry point to claim multiple stages in one tx with a single `transferFrom`.
- Replace `mapping(address => bool) hasClaimed` with per-stage tracking keyed by `keccak256(claimer, leaf)`.
- Extract `_verifyAndMark` internal helper shared by `claim` and `claimBatch`.

## Notes
- Breaking change to the claim ABI (`claim` now takes `validFrom`) and to the Merkle leaf format. Any off-chain tree builder must be updated before this is deployed.
- Storage layout change: `hasClaimed` key type changes from `address` to `bytes32`. Safe for a fresh deploy; not an in-place upgrade.

### DIFF
--- a/claim_contracts/src/ClaimableAirdrop.sol
+++ b/claim_contracts/src/ClaimableAirdrop.sol
@@ -30,9 +30,9 @@ contract ClaimableAirdrop is
     /// @notice Merkle root of the claimants.
     bytes32 public claimMerkleRoot;
 
-    /// @notice Mapping of the claimants that have claimed the tokens.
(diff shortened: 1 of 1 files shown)


JuArce (commented):
Testing WIP

MauroToscano (approved):

PR #2294, opened 2026-08-15
fix(aggregator): guard nil subscription in SubscribeToNewTasksV3 error goroutine

### LINKED ISSUE #2293: [Bug] Nil pointer panic in SubscribeToNewTasksV3 error handling goroutine when primary RPC subscription fails
## Summary
When the primary RPC WebSocket subscription fails on startup but the fallback succeeds, the aggregator launches an error-handling goroutine that calls `sub.Err()` on a nil `event.Subscription` interface, causing a runtime panic and crashing the aggregator.

## Affected File
`core/chainio/avs_subscriber.go`, lines 76 and 125

## Root Cause
`SubscribeToNewTasksV3Retryable` returns `(nil, err)` on failure. If `errMain != nil` but `errFallback == nil`, the function proceeds to launch the error handling goroutine (since one subscription is still active). Inside that goroutine, the select statement reads from `sub.Err()` but `sub` is nil, causing a nil pointer dereference panic.

```go
// Line 76: if this fails, sub == nil
sub, errMain := SubscribeToNewTasksV3Retryable(...)

// Line 119-144: goroutine launched when errFallback == nil
go func() {
    for errMain == nil || errFallback == nil {
        select {
        case err := <-sub.Err(): // PANIC: sub is nil
```

## Reproduction Scenario
1. Start aggregator with a primary RPC endpoint that is unreachable
2. Fallback RPC endpoint is reachable
3. `SubscribeToNewTasksV3` returns `nil, errMain` + valid `subFallback`
4. Error-handling goroutine starts → `sub.Err()` panics immediately

## Fix
Guard `sub.Err()` with a nil check, or use a closed channel as a sentinel:

```go
var subErrCh <-chan error
if sub != nil {
    subErrCh = sub.Err()
}
// same for subFallback
```

## Severity
**Critical** crashes the aggregator process. Any network blip on the primary RPC
at startup is enough to trigger it.

Sertug17:
Closes #2293

## Problem
When the primary RPC WebSocket subscription fails on startup but the fallback succeeds, the error-handling goroutine called `sub.Err()` on a nil `event.Subscription` interface causing an immediate nil pointer dereference panic that crashes the aggregator.

## Fix
Introduce `nilSafeErrCh()` which returns a never-sending channel when the subscription is nil. The select statement in the goroutine now degrades gracefully instead of panicking. The channel variable is also refreshed after each successful resubscription so the fix stays effective for the lifetime of the goroutine.

## Reproduction
1. Start aggregator with primary RPC unreachable, fallback reachable
2. `SubscribeToNewTasksV3Retryable` returns `(nil, err)` for main
3. Error goroutine starts → `sub.Err()` panics immediately

### DIFF
--- a/core/chainio/avs_subscriber.go
+++ b/core/chainio/avs_subscriber.go
@@ -120,22 +120,31 @@ func (s *AvsSubscriber) SubscribeToNewTasksV3(newTaskCreatedChan chan *servicema
 		s.logger.Info("Starting error handling goroutine")
 		var errMain, errFallback error
 		var auxSub, auxSubFallback event.Subscription
+
+		// Guard against nil subscriptions: if a subscription failed on startup,
+		// replace its Err() channel with a never-sending channel so the select
+		// below never blocks on a nil interface (which would panic).
+		subErrCh := nilSafeErrCh(sub)
+		subFallbackErrCh := nilSafeErrCh(subFallback)
+
 		for errMain == nil || errFallback == nil { //while one is active
 			select {
-			case err := <-sub.Err():
+			case err := <-subErrCh:
 				s.logger.Warn("Error in new task subscription of main connection", "err", err)
 
 				auxSub, errMain = SubscribeToNewTasksV3Retryable(&bind.WatchOpts{}, s.AvsContractBindings.ServiceManager, internalChannel, nil, retry.NetworkRetryParams())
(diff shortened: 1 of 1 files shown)


JuArce:
The Verification Layer is deprecated, so we're no longer accepting external PRs

